PR #29, opened 2025-06-04
feat: use linker script from Airbender

shamatar:
## What ❔

Reduced RAM to 1Gb range, and provides 768Mb for allocator to work with (we can adjust stack/heap more if needed). Also stack will underflow into ROM, so we will catch any write attempt to it automatically (but those cases are unprovable).

## Why ❔

Reduce memory consumption of simulator

## Is this a breaking change?
- [x] Yes
- [ ] No

## Checklist


shamatar:
Interesting performance change, most likely related to movement of BSS section or more logic required for allocator (less likely, as ECRECOVER doesn't allocate)

PR #30, opened 2025-06-04
feat: basic zksync-error integration

sayon:
## What ❔

Testing the effect of integrating zksync-error in the binary

## Why

The goal is improved devex and architecture

## Side effects
1. Introduces a new dependency from `zksync-basic-types`

## Is this a breaking change?
- [ ] Yes
- [x] No


zksync_error/errors/evm.json
```diff
@@ -0,0 +1,31 @@
+{
+  "types": [],
+  "domains": [
+    {
+      "domain_name": "OS",
```
antoniolocascio, line 5: Not sure what "domain" should be, is it expected to be the same as in zksync-os.json?

sayon:
The blocker is the dependence on zksync_basic_types, which seems to be incompatible with no_std. I'll think what is the good solution for that -- I actually only use it for U256/H256-like types, so maybe we need to define them separately in zksync_error and provide converters, or extract basic_basic_types from basic_types :)

sayon:
We have decided to instead integrate in the sequencer.

PR #40, opened 2025-06-06
fix: adapted asm, link and memory to match latest airbender

mm-zk:
## What ❔

* Necessary fixes to make app.bin work with the 0.0.6 airbender.

# Why ❔

* 0.0.6 did a bunch of changes how to stack is located and to its size.


0xVolosnikov:
But if everything is ok and this is a blocker (+ decide about https://github.com/matter-labs/zksync-os/pull/29) we can do it

mm-zk:
yeah, this is the same as 29.

(without this, zksyncos app.bin will not work with the newer airbender)

mm-zk:
closing in favor of 29.

PR #41, opened 2025-06-07
fix: Changes for fee model

StanislavBreadless:
## What ❔

Changes:

1. Introduced `UPGRADE_TX_NATIVE_PER_GAS`. Upgrade transactions have 0 fee per gas and so for them the native per gas would be 0. Setting a non-zero value there is IMO not desirable since to keep the invariant `amount_mint >= gas_limit * fee_per_gas`. At the same time the invariant the txs with smaller gas fee than base fee are not included is broken... In any case, I kept the same approach as the one used on Era.
2. Previously all the withheld resources would be burned (they would get into `native_used` [here](https://github.com/matter-labs/zksync-os/blob/53456cbe3f3d90315b506338510cefa019ff8999/basic_bootloader/src/bootloader/process_transaction.rs#L1002-L1006). But now I explicitly remember how much was withheld and return it. This feature is not mandatory (firstly it seemed like mandatory, but when I merged the `main` branch priority txs started working for me, but it is still really nice to not waste users' funds.
3. For L1 txs I always use `transaction.max_fee_per_gas`. Today they must be processed always, regardless of the base fee. Otherwise the priority queue would be stalled. 

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/basic_bootloader/src/bootloader/gas_helpers.rs
+++ b/basic_bootloader/src/bootloader/gas_helpers.rs
@@ -6,6 +6,10 @@ use zk_ee::system::{Computational, Resources};
 
 use super::*;
 
+/// Returns the resources for the transaction and the withheld resources.
+/// The withheld resources are the resources that are withheld from the transaction's
+/// execution to ensure that it does not use too many native computational resources.
+/// They are reclaimed at the end of the transaction and used to charge the pubdata.
 pub fn get_resources_for_tx<S: EthereumLikeTypes>(
     gas_limit: u64,
     native_per_pubdata: U256,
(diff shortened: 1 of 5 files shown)


basic_bootloader/src/bootloader/gas_helpers.rs
```diff
@@ -14,7 +14,7 @@ pub fn get_resources_for_tx<S: EthereumLikeTypes>(
     intrinsic_gas: usize,
     intrinsic_pubdata: usize,
     intrinsic_native: usize,
-) -> Result<S::Resources, TxError> {
```
antoniolocascio, line 17: Can you add a doc line for this function explaining what the return values represent?

basic_bootloader/src/bootloader/gas_helpers.rs
```diff
@@ -39,9 +39,25 @@ pub fn get_resources_for_tx<S: EthereumLikeTypes>(
...
+        (native_limit, S::Resources::from_ergs(Ergs(0)))
+    } else {
+        let withheld =
+            <<S as zk_ee::system::SystemTypes>::Resources as Resources>::Native::from_computational(
+                native_limit - MAX_NATIVE_COMPUTATIONAL,
+            );
+
+        (
+            MAX_NATIVE_COMPUTATIONAL,
+            S::Resources::from_native(withheld),
```
antoniolocascio, line 58: Can you also call set_as_limit for withheld? 
for Decreasing native it doesn’t matter, but it’s useful for Increasing (which are used for benchmarking). Then, in the reclaim_method, we should also set the limit to the sum of self and to_reclaim

antoniolocascio (approved):

PR #46, opened 2025-06-09
fix: Actually clean storage cache on selfdestruct

0xVolosnikov:
## What ❔

Reset storage values to 0.

## Why ❔

At least affects pubdata calculation for user's tx.

## Checklist


0xVolosnikov:
Open question: seems like we do not very actively use Appearance. May be this is redundant?

antoniolocascio (approved):

PR #54, opened 2025-06-09
feat: added reproducible build using docker, and CI to check

mm-zk:
## What ❔

* reproducible build using docker
* also had to add cargo.lock, and final app.bin

## Why ❔

* to allow reproducible builds.


antoniolocascio:
I'm not very familiar with reproducible builds, but is this enough? Does docker guarantee bit-for-bit reproducibility? Don't we need to pin OS version/deps?

mm-zk:
@antoniolocascio - yes, I was wrong, and if you change the underlying OS, you end up with different bytecodes. So we will have to use nix in the long term. (but as nix is running in docker too, you can consider these changes as part 1).

antoniolocascio (approved):

PR #55, opened 2025-06-09
feat: update airbender deps to 0.2.0

mm-zk:
## What ❔
* update airbender deps from 0.1.0 to 0.2.0

## Why ❔

* this uses new version of the circuits (that are faster).


mm-zk:
Seems that we still have blake test failing.. investigating.

PR #60, opened 2025-06-13
fix: replace OsImmutableSlice with ordinary slices and lifetimes

joonazan:
The old approach was the same as using `&'static [u8]` for heaps, returndata and calldata. These all have proper lifetimes now.

Also removes a weird hack from system functions. The heap can now be used as a scratch space perfectly safely.


joonazan:
### Things that I don't like
- the new error prone returning of resources. `must_use` and `+` on Resources would help a lot with that.
- naming: RunnerMemories and `'calldata`. The latter is actually closer to the lifetime of the EE.

zk_ee/src/system/mod.rs
```diff
@@ -46,33 +44,26 @@ pub trait SystemTypes {
...
 
     // These are just shorthands. They are completely defined by the above types.
     type IOTypes: SystemIOTypesConfig;
-    type Resources: Resources;
+    type Resources: Resources + Default;
     type Allocator: Allocator + Clone + Default;
 }
 pub trait EthereumLikeTypes: SystemTypes<IOTypes = EthereumIOTypesConfig> {}
 
 pub struct System<S: SystemTypes> {
```
AntonD3, line 59: I think at this point we need only one System trait which will contain IO + metadata method

AntonD3: We can do it in separate PR.

Actually, it's how it was at the very beginning, before IO and memory split. We actually split it to allow access to memory and io at the same time, but it's not relevant anymore.

shamatar:
"id" precompile has large decrease in number of cycles, most likely indicating absence of 1 more memcopy. Was the previous implementation inefficient, or it's some unintended byproduct?

zk_ee/src/memory/slice_vec.rs
```diff
@@ -78,6 +81,8 @@ impl<T: Clone> SliceVec<'_, T> {
             }
```
shamatar, line 81: I can not comment above this line, but it's related to drop loop above. Hopefully compiler can eliminate a loop completely if drop is noop, but it may be easier to wrap it into "if core::mem::needs_drop::<T>()" that is usually much better eliminated

shamatar:
> All tests pass and performance has improved even though I do not align heap and returndata to 4 bytes. Should I do it?

It's hard to actually estimate the benefits. Memcopy will always branch on alignments of source and destination, by my intuition is that most of EVM code ends up having everything aligned due to its type sizes. If execution is not too complex and it's just 1 contract deep (like ERC20), then we should also care that first calldata is aligned, but I believe it is

zk_ee/src/memory/slice_vec.rs
shamatar: Original Rust doc suggest that you can drop_in_place a slice directly, that would do the same check

basic_bootloader/src/bootloader/runner.rs
```diff
@@ -68,58 +73,78 @@ where
...
+            .handle_requested_deployment(initial_ee_version, true, deployment_parameters, heap)
             .map(TransactionEndPoint::CompletedDeployment),
     }
 }
 
-struct Run<'a, S: EthereumLikeTypes>
-where
-    S::Memory: MemorySubsystemExt,
-{
+pub struct RunnerMemoryBuffers<'a> {
```
antoniolocascio, line 81: Can you remove the commented out lines?

basic_bootloader/src/bootloader/runner.rs
```diff
@@ -754,6 +813,7 @@ where
             deployment_success
         ));
 
+        resources_returned.reclaim(resources_for_deployer);
```
antoniolocascio, line 816: I agree this isn't nice, but it goes beyond the name of the function (we can change it to + if you want). Ideally, this adding back of returned resources should only happen at 1 point instead of being scattered around. 
Maybe we can pass resources_to_give_back (from prep) to the EE from launch, and then the result will have a unified resource to rejoin with caller's?

zk_ee/src/memory/slice_vec.rs
joonazan: For the current use cases this doesn't matter but dropping a slice is a bit less code, so sounds good to me.

shamatar:
Hm, "id" cycle count didn't change compared to previous bot report if I remember correctly

zk_ee/src/memory/slice_vec.rs
0xVolosnikov: Replaced with `drop_in_place`, binary is the same

0xVolosnikov (dismissed):

zk_ee/src/memory/slice_vec.rs
0xVolosnikov: Thanks! Added cast, hope no more stupid mistakes

0xVolosnikov: I'm going to merge it. @joonazan please check this part later

PR #90, opened 2025-06-24
perf: even simpler storage tree

shamatar:
## What ❔

Add functionality to remove keys from the storage tree, and reuse leafs of previously deleted keys. Also, reads of non-existent values are now as efficient as for existing. The only downside is that deletion is 2 writes (same as insert).

## Why ❔

Performance, and try to use leafs more efficiently

## Is this a breaking change?
- [x] Yes
- [ ] No

## Checklist


shamatar:
@slowli Your review is needed here, if you see anything criminal in such model

basic_system/src/system_implementation/flat_storage_model/storage_tree/mod.rs
```diff
@@ -0,0 +1,2645 @@
...
+                    deletion_index: index,
+                },
+            );
+
+            let is_new = empty_slots_cache.insert(index);
+            assert!(is_new);
+        }
+
+        // and last one - insert. We will reuse slots
+        for (key, value) in inserts_iter {
```
0xVolosnikov, line 733: We could start with inserts and for every insert check if we have something in `deletes_iter`, and if any, consume and process one deletion. It may be a bit more complicated, but this way we don't need to relink slots and can basically override slot

0xVolosnikov:
@shamatar looks like the benchmark run fails at https://github.com/matter-labs/zksync-os/blob/ca689de9d88ef709b34841e239a7c0eaa0eb76b6/basic_system/src/system_implementation/flat_storage_model/storage_tree/mod.rs#L713


```
UART: line 
UART: 713
UART: , file 
UART: /zksync_os/basic_system/src/system_implementation/flat_storage_model/storage_tree/mod.rs
UART: , message:

UART: assertion `left 
UART: ==
UART:  right` failed
  left: 
UART: 0x
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 0
UART: 
 right: 
UART: 0x
UART: e7ed2649
UART: e17a9b13
UART: 1448eaa3
UART: 2267ef7a
UART: dc6a1e55
UART: ee8edf2c
UART: 92c12898
UART: ab6e049e
UART: 
```

We haven't used this block for benchmarks before, so the issue wasn't caught

PR #91, opened 2025-06-24
ci: add p256 to bench

yoaveshel:
## What ❔
Add p256 "precompile" to the CI bench

## Why ❔
so we can see if https://github.com/matter-labs/zksync-os/pull/70 is faster

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


yoaveshel:
@antoniolocascio I think the failure of the benchmark here is expected, right?

tests/rig/Cargo.toml
```diff
@@ -18,7 +18,7 @@ forward_system = { path = "../../forward_system" }
 oracle_provider = { path = "../../oracle_provider" }
 zk_ee = { path = "../../zk_ee" }
 zksync_os_runner = { path = "../../zksync_os_runner"}
-system_hooks = { path = "../../system_hooks" }
+system_hooks = { path = "../../system_hooks", features = ["p256_precompile"]}
```
antoniolocascio, line 21: Let's not include it here, but just for instances/precompiles. We might need to define a feature in rig to forward it. Otherwise, EVM state tests will fail

system_hooks/src/addresses_constants.rs
```diff
@@ -14,6 +14,8 @@ pub const ECPAIRING_HOOK_ADDRESS_LOW: u16 = 0x0008;
 pub const BLAKE_HOOK_ADDRESS_LOW: u16 = 0x0009;
 #[cfg(feature = "mock-unsupported-precompiles")]
 pub const POINT_EVAL_HOOK_ADDRESS_LOW: u16 = 0x000a;
+#[cfg(feature = "p256_precompile")]
+pub const P256_VERIFY_PREHASH_HOOK_ADDRESS_LOW: u16 = 0x000b;
```
antoniolocascio, line 18: Let's use some other address, maybe a bit further away. We don't support it, but 0xb -0x11 are taken here https://eips.ethereum.org/EIPS/eip-2537

yoaveshel:
@antoniolocascio I added it as a feature to `instances/precompile`. I can add it as a feature in the test invocation instead, I'm not sure what makes more sense

yoaveshel:
@antoniolocascio done

0xVolosnikov:
I think you should not delete Cargo.lock

PR #94, opened 2025-06-25
Add scripts for easy local rebuilds

StanislavBreadless:
## What ❔

When working on Era locally, there is often a need to build `app.bin` and `app_logging_enabled.bin` to be copied to the server

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/zksync_os/.gitignore
+++ b/zksync_os/.gitignore
@@ -1,4 +1,5 @@
-os.elf
-app.elf
-app.text
+*.elf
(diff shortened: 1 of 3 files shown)


antoniolocascio:
LGTM. Do you think it's worth rebuilding the the app for server in the CI (add it to reproduce.sh) and build it in Dockerfile?

PR #97, opened 2025-06-25
feat(EVM): Implement new stack separately from custom U256

0xVolosnikov:
## What ❔

Separates new EVM stack implementation from custom U256.

## Why ❔

A little bit more performant (reduces amount of U256 copying here and there) than current implementation.

## Is this a breaking change?
- [ ] Yes
- [X] No

## Checklist


shamatar:
I think the better way is some method on storage model implementation (or at least Bytes32), that would be line into_storage_format(U256) -> Bytes32, and vice versa

0xVolosnikov:
Yep, just wanted to see what happens. Will revert my change here, anyway it should be in a different PR

basic_system/src/system_implementation/system/io_subsystem.rs
```diff
@@ -298,7 +298,7 @@ where
                     Bytes32::ZERO
                 } else {
                     // EOA case:
-                    Bytes32::from_u256_be(U256::from_limbs([
+                    Bytes32::from_u256_be(&U256::from_limbs([
```
shamatar, line 301: This would be better to be outlined constant, but can do it separately (I also had const hex-init for Bytes32 stashed somewhere)

shamatar (commented):
I believe it's just a direct port, so fine by me. Comments are for information only

antoniolocascio (approved):

PR #99, opened 2025-06-26
feat: make env opcodes EVM equivalent (mix hash, gas limit)

antoniolocascio:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


evm_interpreter/src/instructions/environment.rs
```diff
@@ -33,10 +33,10 @@ impl<S: EthereumLikeTypes> Interpreter<'_, S> {
         Ok(())
     }
 
-    pub fn difficulty(&mut self) -> InstructionResult {
+    pub fn difficulty(&mut self, system: &mut System<S>) -> InstructionResult {
         self.gas
             .spend_gas_and_native(gas_constants::BASE, DIFFICULTY_NATIVE_COST)?;
-        self.stack.push(U256::ONE)?;
+        self.stack.push(system.get_mix_hash())?;
```
0xVolosnikov, line 39: Why do you name it `mix_hash`?

antoniolocascio: Receipts still use that name. Do you think we should use prevrandao instead? Or something more generic like `randomness`?

0xVolosnikov (approved):

PR #108, opened 2025-06-27
feat: Update airbender dependency to 0.3.1

mm-zk:
## What ❔

* updating airbender dependency from 0.2.0 to 0.3.1

## Why ❔

* It mostly brought features related to performance and multigpu.


0xVolosnikov:
You have to rebase / push something here, otherwise required CI check won't start

PR #113, opened 2025-06-30
Merge force deploy with ad

StanislavBreadless:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


system_hooks/src/l1_messenger.rs
```diff
@@ -148,6 +156,9 @@ where
                     ))
                 }
             };
+            if message_offset != 32 {
```
AntonD3, line 159: @StanislavBreadless in such case we are not fully compatible with solidity implementation. Do you think it's ok?

AntonD3 (approved):

PR #124, opened 2025-07-02
perf: try to walk over u16 using small table to analyze EVM bytecode for jumpdests

shamatar:
## What ❔

Naive trick to improve performance

## Why ❔

Instead of analyzing each byte we can analyze by byte-pairs

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


shamatar:
It's really minimal difference for one block that I tried, but let's see on another one in the performance bot

antoniolocascio:
Looks a bit too complex for the performance gain, no? On 22429200, this other approach had a much bigger effect #125. Wdyt?

PR #130, opened 2025-07-08
feat: Add custom u256

yoaveshel:
## What ❔
This PR introduces a custom U256 type with delegation feature. It does not yet integrate or use this type anywhere else in the codebase. Compared to [#73](https://github.com/matter-labs/zksync-os/pull/73), this version includes additional unit tests and small bug fixes.

## Why ❔
The original PR ([#73](https://github.com/matter-labs/zksync-os/pull/73)) was becoming too complex. This is the first step in splitting that work into smaller parts

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [x] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.
- [x] Code has been formatted.

PR #131, opened 2025-07-08
perf:  Use custom U256 in crypto crate

yoaveshel:
## What ❔
Step 2 in breaking #73 into multiple PRs. This one changes the crypto crate to use the custom U256
Note, this PR depends on #130 

## Why ❔
Reduce duplication, possibly a little speedup from using precompile for `Clone`

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


yoaveshel:
Well, these results don't make no sense. I'll change this PR into a draft

shamatar:
I think we still have inefficient  codegen for "copy if needed". Can you check an assembly?

shamatar:
Also we do not really control everything - it's anyway Arkworks on top. We can keep it as valid, but not merge for now. Can you now look at modexp?

PR #135, opened 2025-07-08
perf: move to custom U256

yoaveshel:
## What ❔
The third (and hopefully last) episode in the custom U256 saga. Here we take advantage of the U256 type from #130

## Why ❔
Use precompile for copies mostly, which should be a little faster

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


0xVolosnikov:
I'll check EVM tester

0xVolosnikov:
Ok, now this is interesting.

yoaveshel:
@0xVolosnikov I will change this into a draft and leave it as is for now

PR #141, opened 2025-07-11
test: add tests for native charging

antoniolocascio:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


tests/rig/src/chain.rs
```diff
@@ -154,6 +155,76 @@ impl<const RANDOMIZED_TREE: bool> Chain<RANDOMIZED_TREE> {
...
+            let aux_oracle: ForwardRunningOracleAux<
+                InMemoryTree<RANDOMIZED_TREE>,
+                InMemoryPreimageSource,
+                TxListSource,
+            > = oracle.clone().into();
+            let serialized_oracle = bincode::serialize(&aux_oracle).expect("should serialize");
+            let mut file = File::create(&path).expect("should create file");
+            file.write_all(&serialized_oracle)
+                .expect("should write to file");
+            info!("Successfully wrote oracle dumo to: {}", path);
```
0xVolosnikov, line 207:
```suggestion
            info!("Successfully wrote oracle dump to: {}", path);
```

0xVolosnikov:
So basically we just run some txs with low/med/high native per gas and check their status, right? Without any tight native gas checking at this step

antoniolocascio:
@0xVolosnikov the concern here is not the model itself, but rather the logic around charging/refunding. It previously caused internal errors/panics, that's what I want to mostly catch here.

0xVolosnikov (approved):
lgtm

PR #146, opened 2025-07-14
feat: Initial tracing support

0xVolosnikov:
## What ❔

Introduces initial limited tracing functionality.

Also I included some reference implementations:

- [CallTracer](https://github.com/matter-labs/zksync-os/blob/vv-tracer/forward_system/src/system/tracers/call_tracer.rs): similar to Geth's [callTracer](https://geth.ethereum.org/docs/developers/evm-tracing/built-in-tracers#call-tracer). Differences: it doesn't treat selfdestructs as calls for now and doesn't include proper error reasons (from system itself).
- [EvmOpcodesLogger](https://github.com/matter-labs/zksync-os/blob/vv-tracer/forward_system/src/system/tracers/evm_opcodes_logger.rs): mimics Geth's [opcode logger](https://geth.ethereum.org/docs/developers/evm-tracing/built-in-tracers#struct-opcode-logger). Differences: doesn't return gasCost (requires some redesign of our EVM EE), refund is always 0 (it is not implemented in our system yet), doesn't include proper error reasons (from system itself).

Stuff to do later (as separate PRs / issues):
- Remove debug logs (we can use some kind of debugging tracer in most cases)
- Include proper error reasons from the system. E.g. [Geth serializes inner error codes](https://github.com/ethereum/go-ethereum/blob/0dacfef8ac42e7be5db26c2956f2b238ba7c75e8/core/vm/interpreter.go#L309)
- Handle selfdestruct as external call. It's not easy for now to get value, gas, etc.
- Provide a proper way to get gasCost for opcodes

## Why ❔

Tracer support is required for proper sequencer functionality. Also can be used for debugging.

## Is this a breaking change?
- [X] Yes
- [ ] No

## Checklist


zk_ee/src/system/tracer.rs
```diff
@@ -0,0 +1,187 @@
...
+
+use crate::{memory::slice_vec::SliceVec, types_config::SystemIOTypesConfig};
+
+use super::{
+    CallOrDeployResult, ExecutionEnvironmentLaunchParams, ExecutionEnvironmentSpawnRequest,
+    SystemTypes,
+};
+
+pub trait Tracer<S: SystemTypes> {
+    fn should_call_before_evm_execution_step(&self) -> bool;
```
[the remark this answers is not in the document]
0xVolosnikov: Removed flags, compiler should optimize no-ops  out

basic_bootloader/src/bootloader/runner.rs
```diff
@@ -348,90 +318,96 @@ impl<'external, S: EthereumLikeTypes> Run<'_, 'external, S> {
...
+                            CallOrDeployResultRef::CallResult(call_result),
+                        ))
+                    })
+                    .unwrap_or_default(),
+            );
         }
+
+        let (resources_returned_from_callee, call_result) = callee_frame_execution_result?;
+        let mut resources_to_return = resources_returned_from_callee;
+        resources_to_return.reclaim(resources_returned_from_preparation); // TODO why do we reclaim it only here?
```
0xVolosnikov, line 382: @antoniolocascio pls take a look here

shamatar:
What are those magic things that need to modify “state” in a middle of execution? And how does forking work in e.g. foundry? If we try to make 100% equivalence with observable behavior of EVM then we mainly need to dump data to foundry and let it play with itself for all the mutations. What would be a data boundary in this case?

0xVolosnikov:
Thank you for your comments! I'll split this pr into runner changes and tracing itself and will address your suggestions. 
I think our top priority here is to support basic call tracing for API, so I'd focus on that. Regarding mutation of state and other advanced debugging - imo the best way to do it would be to create a separate "runner" implementation which will expose more functionality outside, like REVM does. So users of this debugging runner could do whatever they want

0xVolosnikov:
Runner simplification PR: https://github.com/matter-labs/zksync-os/pull/161

evm_interpreter/src/instructions/host.rs
```diff
@@ -236,7 +236,9 @@ impl<'ee, S: EthereumLikeTypes> Interpreter<'ee, S> {
         self.resize_heap(mem_offset, len)?;
         let data = &self.heap[mem_offset..mem_offset + len];
 
-        tracer.on_event(THIS_EE_TYPE, &self.address, &topics, data);
+        if tracer.is_on_event_enabled() {
```
joonazan, line 239: I don't think this is something that needs to be toggled on and off during tracer execution, so consider using `TracerType::IS_ON_EVENT_ENABLED` instead.

joonazan: Personally, I don't find the readability advantage of `impl` notation to be a reason not to use explicit type params.

Technically, you could avoid type params by adding optional methods to the trait that perform the if.

0xVolosnikov: Hmm, sounds interesting. Ok, I'll try to change it to generics after all

joonazan: Did you try and see that the separate `is_enabled` is necessary? Explicit generics or impl trait should make no difference for inlining.

0xVolosnikov: Seems like it works as expected now!

0xVolosnikov:
`Update app.bin `doesn't work now. I'll merge it, but we should not cut a release until it fixed

PR #154, opened 2025-07-17
feat(7702): Support parsing and hashing 7702 tx type

antoniolocascio:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


basic_bootloader/src/bootloader/process_transaction.rs
```diff
@@ -534,18 +539,11 @@ where
         let chain_id = system.get_chain_id();
 
         // Process access list
-        // Note: this operation should be performed before the hashing of the
```
0xVolosnikov, line 537: Can you pls clarify why it was removed? L1 tx can't be invalid?

tests/instances/transactions/src/lib.rs
```diff
@@ -501,6 +501,65 @@ fn test_tx_with_access_list() {
...
+    chain.set_evm_bytecode(B160::from_be_bytes(to.into_array()), &bytecode);
+
+    chain.set_balance(
+        B160::from_be_bytes(from.0),
+        U256::from(1_000_000_000_000_000_u64),
+    );
+
+    let output = chain.run_block(transactions, None, None);
+
+    // Assert all txs succeeded
```
0xVolosnikov, line 558: Does alloy have some negative cases in tests?

tests/rig/src/utils.rs
```diff
@@ -441,14 +498,32 @@ mod tests {
...
+                U256::from(53),
+            ),
+        ];
+
+        let encoded_list = encode_reserved_dynamic(access_list, authorization_list);
         let encoded = ethers::abi::encode(&[Token::Bytes(encoded_list)]);
-        // Prepend TX_OFFSET bytes, as those are then ignored by the parser.
-        let mut full_buffer = vec![0u8; TX_OFFSET];
-        full_buffer.extend(encoded);
+        // println!("Encoded: {}", hex::encode(encoded.clone()));
```
0xVolosnikov, line 522: Remove?

0xVolosnikov:
Generally lgtm, left some comments

basic_bootloader/src/bootloader/process_transaction.rs
antoniolocascio: This is L2 tx, right? Not sure what you mean by the second part.
Access/authorization list parsing is done lazily. This parsing can fail if the structure is incorrect. However, after the first time you traverse it, subsequent traversals cannot fail. This fact was being used before to make tx hashing infallible (other than out of resources), which means that if you hashed a tx before doing the first parsing, it would get to an internal error. But I thought it was dangerous to keep this invariant, so I reverted to making each traversal safe (returning validation error).

basic_bootloader/src/bootloader/transaction/mod.rs
```diff
@@ -856,6 +966,133 @@ impl<'a> ZkSyncTransaction<'a> {
...
+            + rlp::estimate_number_encoding_len(
+                self.max_priority_fee_per_gas
+                    .encoding(&self.underlying_buffer),
+            )
+            + rlp::estimate_number_encoding_len(
+                self.max_fee_per_gas.encoding(&self.underlying_buffer),
+            )
+            + rlp::estimate_number_encoding_len(self.gas_limit.encoding(&self.underlying_buffer));
+
+        // Handle `to` == null, indicates EVM deployment transaction
```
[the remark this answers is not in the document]
antoniolocascio: No, you're right, just forgot to remove this, it's never null

tests/instances/transactions/src/lib.rs
antoniolocascio: This was mostly for sanity, but the alloy tests are quite weird. Spec tests have a bunch of negative ones, let's rely on those

0xVolosnikov (approved):

PR #157, opened 2025-07-18
feat: Add helpers for API to encode/decode account

antoniolocascio:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


api/src/helpers.rs
```diff
@@ -0,0 +1,80 @@
...
+/// Retrieves nonce from an account.
+pub fn get_nonce(account: &AccountProperties) -> u64 {
+    account.nonce
+}
+
+/// Retrieves code for an account.
+/// This function returns unpadded code, without artifacts.
+pub fn get_code<P: PreimageSource>(
+    preimage_source: &mut P,
+    account: &AccountProperties,
```
joonazan, line 25: I think this function doesn't need to exist. However, there should be a function for splitting a preimage (`&[u8]`) into its parts.

joonazan (changes requested):
I would like to have a function that tells me what storage slot (B256) to look at to get AccountProperties for an address and then another function that turns an accountproperties preimage into the struct.

api/src/helpers.rs
antoniolocascio: Is there where you need that but can't use this function?

antoniolocascio: This one is quite useful, but it makes sense. I broke it down into two helpers.

joonazan: I didn't realize that AccountProperties are needed to be able to properly cut the preimage. Then the original signature does make more sense.

api/src/helpers.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/// Retrieves nonce from an account.
+pub fn get_nonce(account: &AccountProperties) -> u64 {
+    account.nonce
+}
+
+/// Get unpadded code from full bytecode with artifacts.
+pub fn get_unpadded_code(full_bytecode: &[u8], account: &AccountProperties) -> Vec<u8> {
+    full_bytecode[0..account.unpadded_code_len as usize].to_vec()
```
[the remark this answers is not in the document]
antoniolocascio: Sure, I kept the vec in the other one, but changed this to slice. Caller can use this one to avoid allocating

joonazan (dismissed):

PR #160, opened 2025-07-21
[NO MERGE] feat: impl keccak/sha3 precompile

mabbamOG:
## What ❔

new keccak impl using local precompile

DO NOT MERGE this because airbender dependencies have been moved to keccak branch for testing, but we should wait until they hit airbender main and then we can use normal deps

## Is this a breaking change?
- [ ] Yes
- [x] No


mabbamOG:
info: this pr can't pass tests and do benchmarks until zksync-airbender keccak pr is accepted. that one is on hold due to some final ongoing fixes + potential optimisations

crypto/src/sha3/mod.rs
```diff
@@ -1,35 +1,513 @@
-pub use sha3::*;
```
shamatar, line 1: Please follow the blake2s path - keep two implementations under features, one via "standard" implementation (sha3::Keccak256) and one via your code below

shamatar (changes requested):

crypto/src/sha3/mod.rs
mabbamOG: and the feature i should keep out of default ? just optional ?

mabbamOG: i put in default btw

shamatar:
Included in #271

PR #161, opened 2025-07-21
feat: Simplify runner flow

0xVolosnikov:
## What ❔

Current flow can be simplified. This change is also required for tracer implementation.
This will be followed with deeper changes later.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


basic_bootloader/src/bootloader/runner.rs
```diff
@@ -824,91 +812,57 @@ impl<'external, S: EthereumLikeTypes> Run<'_, 'external, S> {
...
-                    &call_request,
-                    is_entry_frame,
-                )
+                read_callee_account_properties(system, ee_version, inf_resources, &call_request)
             })
         })
     } else {
-        cycle_marker::wrap_with_resources!("prepare_for_call", resources_available, {
-            prepare_for_call(
+        cycle_marker::wrap_with_resources!("prepare_for_call", resources_in_caller_frame, {
```
0xVolosnikov, line 851: @antoniolocascio I'm not sure that it captures the scope that you intended anymore, pls take a look

0xVolosnikov:
TODO: deployment should properly handle EE-specific failures in callee frame before actual VM execution (e.g. address collision in EVM)

basic_bootloader/src/bootloader/runner.rs
```diff
@@ -479,74 +403,102 @@ impl<'external, S: EthereumLikeTypes> Run<'_, 'external, S> {
...
+                            CallResult::Successful { return_values }
+                        },
+                    ));
                 }
-                Ok(CallPreparationResult::Failure { resources_returned }) => {
-                    return Ok((resources_returned, CallResult::CallFailedToExecute))
+                ExecutionEnvironmentPreemptionPoint::End(
+                    TransactionEndPoint::CompletedDeployment(_),
+                ) => {
+                    //TODO should be misuse
```
antoniolocascio, line 475: Should we fix this now?

basic_bootloader/src/bootloader/runner.rs
antoniolocascio: It doesn't, but that has been true for a while (since we move the transfer out of there). Can you just remove the cycle marker call? I can re-add it properly if I need to re-run this part of the native model, we don't use this marker for anything else

antoniolocascio (approved):

antoniolocascio (approved):

PR #178, opened 2025-07-28
fix: change to overflowing_add

yoaveshel:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/crypto/src/secp256k1/field/field_10x26.rs
+++ b/crypto/src/secp256k1/field/field_10x26.rs
@@ -561,31 +561,31 @@ impl FieldElement10x26 {
         self.0[9] &= 0x03FFFFF;
 
         // The first pass ensures the magnitude is 1, ...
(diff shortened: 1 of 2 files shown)


shamatar:
As this code originally is a direct backport of C implementation that most likely didn't have carry propagation this way (as it defeats the purpose of underfilled limbs), can you check again if you need it? In C overflows are not checked, so maybe wrapping is indeed expected? Or we forget to normalize somewhere in the middle of our field ops, so magnitude overflows

yoaveshel:
@shamatar indeed in the C code they don't deal with overflows, but I am pretty sure this is a bug. It only every becomes a problem if one of the limbs is extremely unnormalised which I don't think ever arises in practice. But if it does k256 does not give the right answer: https://play.rust-lang.org/?version=stable&mode=release&edition=2024&gist=dccd151ea1c9f56987c325fed46a03bc

PR #194, opened 2025-07-30
feat: enable interop for zksync-os

koloz193:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


zk_ee/src/system/mod.rs
```diff
@@ -100,6 +101,10 @@ impl<S: SystemTypes> System<S> {
         }
     }
 
+    pub fn get_interop_roots(&self) -> [InteropRoot; 100] {
```
0xVolosnikov, line 104: Again, why 100? Should it be just `&[InteropRoot]`?

zk_ee/src/system/metadata.rs
```diff
@@ -71,6 +73,104 @@ impl UsizeDeserializable for BlockHashes {
     }
 }
 
+#[cfg_attr(feature = "testing", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Clone, Copy, Debug, PartialEq, Default)]
+pub struct InteropRoot {
+    pub root: [Bytes32; 1],
```
0xVolosnikov, line 79: Why not just Bytes32?

basic_bootloader/src/bootloader/mod.rs
```diff
@@ -476,4 +487,77 @@ impl<S: EthereumLikeTypes> BasicBootloader<S> {
...
+                data[56..64].copy_from_slice(&interop_root.block_number.to_be_bytes());
+                data[92..96].copy_from_slice(&32u32.to_be_bytes());
+                data[124..128].copy_from_slice(&1u32.to_be_bytes());
+                data[128..160].copy_from_slice(&interop_root.root[0].as_u8_ref());
+                interop_root_hasher.update(&data);
+
+                data[92..96].copy_from_slice(&96u32.to_be_bytes());
+
+                // fb6200c6: function addInteropRoot(uint256 chainId, uint256 blockOrBatchNumber, bytes32[] calldata sides) external;
+                let mut calldata = [0xfb, 0x62, 0x00, 0xc6].to_vec();
```
0xVolosnikov, line 531: I don't like this `.to_vec()` - we should not use global allocator

basic_bootloader/src/bootloader/mod.rs
```diff
@@ -476,4 +487,77 @@ impl<S: EthereumLikeTypes> BasicBootloader<S> {
...
+                calldata.extend(&data);
+
+                let _ = Self::run_single_interaction(
```
0xVolosnikov, line 534: What if this run fails for some reason? Imo we should handle result of this function

basic_bootloader/src/bootloader/mod.rs
```diff
@@ -476,4 +487,77 @@ impl<S: EthereumLikeTypes> BasicBootloader<S> {
             }
         }
     }
+
+    #[cfg(feature = "interop")]
```
0xVolosnikov, line 491: Maybe we should just check that interop roots list is empty if interop is disabled?

0xVolosnikov:
Left some comments. Look like this requires some more work, @koloz193 pls let us know if we should take it

zk_ee/src/system/metadata.rs
```diff
@@ -71,6 +73,104 @@ impl UsizeDeserializable for BlockHashes {
...
+    pub block_number: u64,
+    pub chain_id: u64,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct InteropRoots(pub [InteropRoot; 100]);
```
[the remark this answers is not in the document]
koloz193: yea

system_hooks/src/addresses_constants.rs
```diff
@@ -38,6 +38,11 @@ pub const NONCE_HOLDER_HOOK_ADDRESS_LOW: u16 = 0x8003;
 pub const NONCE_HOLDER_HOOK_ADDRESS: B160 =
     B160::from_limbs([NONCE_HOLDER_HOOK_ADDRESS_LOW as u64, 0, 0]);
 
+// l2 interop root storage system hook (contract) needed for all envs (add interop root)
```
AntonD3, line 41: nit: maybe makes sense to move this constant to bootloader to avoid confusion(since it's not a system hook)

zk_ee/src/system/metadata.rs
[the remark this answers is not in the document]
AntonD3: Looks like ArrayVec may be useful here

basic_bootloader/src/bootloader/mod.rs
AntonD3, line 531: maybe you can put selector just at the beginning of data - so you can avoid memcopy

AntonD3 (commented):
Left few nits, but also wanted to mention that we need to sync this PR with batch(aka "aggregation") - https://github.com/matter-labs/zksync-os/pull/152

zk_ee/src/system/mod.rs
AntonD3: nit, bur agree with Vlad that it's better to use slice here, otherwise we'll copy interoop roots with this functions calls.

As for borrowing issues, I think you can just rewrite loop in the bootloader, to borrow interop roots on each iteration only for calldata preparation. Alternatively, you can make this method return Option, and use `take` for return, but slice is better IMHO.

basic_bootloader/Cargo.toml
```diff
@@ -36,3 +36,4 @@ disable_system_contracts = []
 error_origins = ["zk_ee/error_origins"]
 evm_refunds = ["zk_ee/evm_refunds"]
 pectra = []
+interop = []
```
perekopskiy, line 39: please remove

forward_system/Cargo.toml
```diff
@@ -39,3 +39,4 @@ disable_system_contracts = ["basic_bootloader/disable_system_contracts"]
 prevrandao = ["zk_ee/prevrandao"]
 evm_refunds = ["zk_ee/evm_refunds", "basic_system/evm_refunds", "basic_bootloader/evm_refunds"]
 pectra = ["basic_bootloader/pectra"]
+interop = ["basic_bootloader/interop"]
```
perekopskiy, line 42: and here as well

zk_ee/src/system/metadata.rs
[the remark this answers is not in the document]
0xVolosnikov: I suppose that 100 roots limitation is somehow related to general limitation on amount of ZK chains, isn't it? Maybe we should keep ZKsync OS agnostic and check it only on L1?

Also maybe silly question: why can't we submit only root of roots?..

zk_ee/src/system/metadata.rs
```diff
@@ -71,12 +74,169 @@ impl UsizeDeserializable for BlockHashes {
...
+        }
+        let len = array_vec.len().try_into().expect("Deserialization failed");
+        Ok(Self {
+            roots: array_vec,
+            length: len,
+        })
+    }
+}
+
+impl UsizeSerializable for InteropRootsContainer {
```
0xVolosnikov, line 149: @AntonD3 pls take a look here. Maybe you'll have some better ideas how to encode/decode it. Iiuc, we can't have any dynamic structures here

0xVolosnikov:
Now it works - I manually compared the rolling hash with the L1 contracts, it matches

0xVolosnikov:
Need to check how it affects storage caches and block native resource limit etc

0xVolosnikov:
Closed in favor of https://github.com/matter-labs/zksync-os/pull/419

PR #223, opened 2025-08-14
feat(tracer): Calculate gas used by EVM opcodes

0xVolosnikov:
## What ❔

ditto

## Why ❔

This field is expected in opcodes trace

## Apologies 

This is really dirty and I'm ashamed a bit. But this is only a reference implementation (PoC) and our current design doesn't allow to handle it in a proper way. We **really** should move charging for reading callee account inside the EE.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


0xVolosnikov:
Requires https://github.com/matter-labs/zksync-os/pull/212

PR #244, opened 2025-08-25
fix: do not update block accumulators when tx is reverted

antoniolocascio:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/basic_bootloader/src/bootloader/mod.rs
+++ b/basic_bootloader/src/bootloader/mod.rs
@@ -293,25 +293,35 @@ impl<S: EthereumLikeTypes> BasicBootloader<S> {
                                 "Tx execution result = {:?}\n",
                                 &tx_processing_result,
                             ));
-                            block_gas_used += tx_processing_result.gas_used;
-                            block_computational_native_used +=
-                                tx_processing_result.computational_native_used;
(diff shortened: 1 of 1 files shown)


basic_bootloader/src/bootloader/mod.rs
```diff
@@ -293,25 +293,35 @@ impl<S: EthereumLikeTypes> BasicBootloader<S> {
...
+                                next_block_gas_used,
+                                next_block_computational_native_used,
+                                next_block_pubdata_used,
                                 block_logs_used,
                             ) {
                                 // Revert to state before transaction
                                 system.finish_global_frame(Some(&pre_tx_rollback_handle))?;
                                 result_keeper.tx_processed(Err(err));
                             } else {
+                                // Now update the accumulators
```
AntonD3, line 319: seems you can just move logic to update accumulators here

AntonD3 (changes requested):
It may also affect other state changes we are doing in the bootloader. I checked it, on line 356(366) we update `first_tx` flag - we should move it as well

antoniolocascio:
@AntonD3 fixed that

AntonD3 (approved):

PR #249, opened 2025-08-26
feat: Return deployed bytecode as returndata for top-level deployments (from EOA)

0xVolosnikov:
Generally it would be more clean to adjust our deployment flow in the runner - to properly **return** deployed bytecode. It would also made tracing easier. 

But in our current design we always copy returned data into the separate returndata buffer and probably it doesn't make sense to spend computation on these copies for every deployment. So this somewhat hacky solution may be a little bit better for performance.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


antoniolocascio (dismissed):

itegulov (dismissed):
Tested in zksync-os-server

antoniolocascio (approved):

PR #253, opened 2025-08-29
fix: make p256 EIP 7951 compliant

antoniolocascio:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


basic_system/src/system_functions/p256_verify.rs
```diff
@@ -32,6 +26,25 @@ impl<R: Resources> SystemFunction<R, P256VerifyErrors> for P256VerifyImpl {
     }
 }
 
+///
+/// Return value should be 1 if successful or empty in any other case:
+///  - signature is invalid
+///  - input is invalid
```
yoaveshel, line 32: It's a little confusing to use "input" here, since later input also includes the signature. Maybe curve point is clearer?

antoniolocascio: I meant input as a whole there. I think the problem is with "valid". There are generally 2 failing cases: bad input and signature verification failure.

yoaveshel: ok, that makes sense

tests/instances/precompiles/markers.bench
```diff
@@ -2,14 +2,2159 @@ Spent ergs for [ecrecover]: 768000
 
 Spent native for [ecrecover]: 522000
 
-Spent ergs for [prepare_for_call]: 0
+Spent ergs for [ecrecover]: 768000
```
0xVolosnikov, line 5: Should we remove this file?

0xVolosnikov:
I'd like to update our tester to Fusaka fixtures first. After that we can actually run state tests for the p256 precompile

0xVolosnikov:
Let's merge this PR: https://github.com/matter-labs/zksync-os/pull/305 and then try to run corresponding tests there

0xVolosnikov (approved):
I didn't review the changes in precompile really deeply, so I hope we have a good coverage by tests.

zksync_os/Cargo.lock
```diff
@@ -227,7 +227,7 @@ dependencies = [
 [[package]]
 name = "blake2s_u32"
 version = "0.1.0"
-source = "git+https://github.com/matter-labs/zksync-airbender?tag=v0.4.1#68953933e3a0846d2155ed0ee024437fc405558a"
+source = "git+https://github.com/matter-labs/zksync-airbender.git?tag=v0.4.3#c23fd0085ee87872e25ff83f780d2d0ee102a166"
```
0xVolosnikov, line 230: Why, btw? Did we forget to update Cargo.lock before?

PR #266, opened 2025-09-03
feat: change L1->L2/upgrade tx type id

itegulov:
## What ❔

L1->L2 priority tx type id is changed from 0xff to 0x7f
Upgrade tx type id is changed from 0xfe to 0x7e

## Why ❔

Makes these transaction types EIP-2718 compliant and RLP-encodable

## Is this a breaking change?
- [x] Yes
- [ ] No

## Checklist


0xVolosnikov:
@antoniolocascio pls take a loot at the fuzzer changes - I think we should check our fuzzy tests deeper, I'm not sure that they actually work as expected

0xVolosnikov (approved):

PR #274, opened 2025-09-09
feat: migrate to `zksync-os-interface`

itegulov:
## What ❔

Use types and implement traits from `zksync-os-interface`.

## Why ❔

Needed for multivm. 

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,7 +98,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           repository: matter-labs/era-evm-tester 
-          ref: bump-serde-and-alloy
+          ref: zk-ee
           path: era-evm-tester
       - name: Use local ZKsync OS version
         working-directory: ./era-evm-tester/evm_tester


.github/workflows/ci.yml
```diff
@@ -98,7 +98,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           repository: matter-labs/era-evm-tester 
-          ref: zk-ee
+          ref: bump-serde-and-alloy
```
perekopskiy, line 101: .

basic_system/Cargo.toml
```diff
@@ -34,7 +36,7 @@ evm-compatibility = []
 default = ["testing"]
 cycle_marker = ["cycle_marker/log_to_file"]
 wrap-in-batch = []
-proving = ["crypto/proving"]
+proving = ["crypto/proving", "zksync_os_types/unsafe-single-core-cas"]
```
perekopskiy, line 39: We want to use alloy primitive types for types in interface. Alloy import crates that use atomics and they can't be compiled for risc-v unless interface also uses `portable-atomic` crate with `unsafe-assume-single-core` feature.
Alternative approach is to not depend on the interface in basic_bootloader, zk_ee and duplicate some types and glue them with interface types in forward_system, it's implemented here https://github.com/matter-labs/zksync-os/commit/4b1f11a2743a60382ed35b510411bfc3b27888e0 (separate branch).

If you prefer an alternative approach please react and we will add commit to this PR.

0xVolosnikov:
I see a significant impact on benchmarks here for some reason

perekopskiy:
> I see a significant impact on benchmarks here for some reason

I [ran](https://github.com/matter-labs/zksync-os/pull/277) benchmarks with an alternative approach (see [this](https://github.com/matter-labs/zksync-os/pull/274#discussion_r2336660722)), impact is gone so it looks like it's the way to go

basic_system/Cargo.toml
```diff
@@ -12,18 +12,19 @@ categories.workspace = true
...
 crypto = { path = "../crypto", default-features = false, features = ["secp256k1-static-context"]}
-arrayvec = { version = "0.7.4", default-features = false }
-either = { version = "*", default-features = false }
 storage_models = { path = "../storage_models", default-features = false }
 modexp = { path = "../supporting_crates/modexp" }
 cycle_marker = { path = "../cycle_marker" }
 
+ruint = { version = "1.12.3", default-features = false, features = ["alloc"] }
+arrayvec = { version = "0.7.4", default-features = false }
+either = { version = "*", default-features = false }
```
mm-zk, line 22: Seems that this PR is only moving things around, but doesn't change anything.

If yes - then don't include this file.

On the side note - we should create a PR to get rid of '*' in our deps

basic_bootloader/Cargo.toml
```diff
@@ -17,6 +17,7 @@ system_hooks = { path = "../system_hooks", default-features = false }
 basic_system = { path = "../basic_system", default-features = false }
 ruint = { version = "1.12.3", default-features = false, features = ["alloc"] }
 crypto = { path = "../crypto", default-features = false }
+
```
mm-zk, line 20: revert?

mm-zk (approved):

basic_system/Cargo.toml
perekopskiy: > Seems that this PR is only moving things around, but doesn't change anything.

Mostly true, it also duplicates some types and implements interface.
I left this change because it split deps into groups: local and external and it make cargo.toml structure better imo

forward_system/src/run/mod.rs
```diff
@@ -6,6 +6,7 @@ mod tree;
 mod tx_result_callback;
```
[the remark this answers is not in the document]
perekopskiy: hmm, not sure I follow, we already have `interface_impl` with zksync_os_interface impl. I would keep `tx_result_callback`, `tree`, `tx_source`, `preimage_source` as is to keep the previous module structure.

forward_system/src/run/tx_source.rs
```diff
@@ -1,9 +1,2 @@
-#[derive(Debug, Clone)]
-pub enum NextTxResponse {
-    Tx(Vec<u8>),
-    SealBlock,
-}
-
-pub trait TxSource: 'static {
-    fn get_next_tx(&mut self) -> NextTxResponse;
-}
+// Re-export interface traits for backward compatibility.
```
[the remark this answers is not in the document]
perekopskiy: Looks like the only place that uses this re-export is
```
pub use tx_source::NextTxResponse;
pub use tx_source::TxSource;
```
in `forward_system/src/run/mod.rs`. IMO it looks better this way: for each trait used for the system we have a separate file; but I can drop the file and use zksync_os_interface::TxSource, zksync_os_interface::NextTxResponse directly if you think it's better

PR #302, opened 2025-09-25
fix: clean up compilation features

antoniolocascio:
## What ❔

This PR tries to clean up features and binaries. The key points are:
* No more default app.bin, bin type has to be selected explicitly.
* wrap-in-batch is removed as it's now the default behaviour 
* charge_priority_fee is removed as it's now the only behaviour 
* forward and proving systems now export a single feature for each use case (tests, evm_tester, eth_runner, etc). Each binary only needs to pick the corresponding feature in dump_bin.sh.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


0xVolosnikov:
There are some issues in CI + we need to revert that change with evm_refunds.
Should we add a separate feature for sequencing as well btw? An explicit option may be better than "default"

antoniolocascio:
@0xVolosnikov rebased and fixed for refunds. We can't really make it work for bench, as this change is not backwards compatible with main

Cargo.toml
```diff
@@ -26,7 +26,6 @@ members = [
     "tests/instances/precompiles",
     "tests/instances/bench",
     "tests/instances/unit",
-    "tests/instances/eth_runner",
```
0xVolosnikov: Why btw?

0xVolosnikov:
And I guess the same is valid for eth_runner as well

0xVolosnikov:
Reverted "charge_priority_fee is removed as it's now the only behaviour" - let's change it in a separate PR

Cargo.toml
0xVolosnikov: Can discuss later

0xVolosnikov:
Should be continued in other PRs

PR #309, opened 2025-09-29
feat: implement interface tracing API

itegulov:
## What ❔

This PR implements interface-level tracing API introduced in https://github.com/matter-labs/zksync-os-interface/pull/20

## Why ❔

Needed for historical availability of traces

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


forward_system/src/run/tracing_impl.rs
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    fn is_constructor(&self) -> bool {
+        self.inner.is_constructor()
+    }
+}
+
+fn convert_error(error: &EvmError) -> zksync_os_interface::error::EvmError {
+    match error {
+        EvmError::Revert => zksync_os_interface::error::EvmError::Revert,
```
0xVolosnikov, line 304: Imo it's better to use zksync_os_interface::error::EvmError inside of ZKsync OS instead

0xVolosnikov: Unfortunately this doesn't work due to std somewhere in deps etc.

0xVolosnikov: Probably we should move EVM errors to a separate crate, wdyt?

itegulov: I can make the whole crate `no_std` pretty easily. In fact I have already done it once during the first phase of integration. Using it across the code had a significant impact on benchmarks as you might remember. But if we limit the non-forward usage to just errors then it should be fine.

Sorry this is dragging on a bit, I got distracted by other work but I am hoping to bring this to a finish today/tomorrow.

itegulov: Actually it was too much hassle (forgot that it needed all that stuff with portable atomic and single-threaded CAS) so I crated another crate as you suggested. PTAL

forward_system/src/run/interface_impl.rs
```diff
@@ -20,22 +21,29 @@ impl RunBlock for RunBlockForward {
     type Config = ();
     type Error = ForwardSubsystemError;
 
-    fn run_block<T: ReadStorage, PS: PreimageSource, TS: TxSource, TR: TxResultCallback>(
+    fn run_block<
+        Storage: ReadStorage,
+        PreimgSrc: PreimageSource,
+        TrSrc: TxSource,
+        TrCallback: TxResultCallback,
+        Tracer: EvmTracer,
```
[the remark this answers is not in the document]
0xVolosnikov: Let's just rename it for now, wdyt? `EvmTracer` is too specific for the interface imo

PR #313, opened 2025-10-02
feat!: Add mint and withdrawal events for L2BaseToken

0xVolosnikov:
## What ❔

This PR adds events expected from L2BaseToken system contract: `Withdrawal`, `WithdrawalWithMessage` and `Mint`.

## Why ❔

Better observability and integration with indexers/explorers.

## Is this a breaking change?
- [X] Yes
- [ ] No

## Checklist

### DIFF
--- a/basic_bootloader/src/bootloader/run_single_interaction.rs
+++ b/basic_bootloader/src/bootloader/run_single_interaction.rs
@@ -4,7 +4,7 @@ use arrayvec::ArrayVec;
 use errors::BootloaderSubsystemError;
 use system_hooks::addresses_constants::L2_BASE_TOKEN_ADDRESS;
 use system_hooks::HooksStorage;
-use zk_ee::kv_markers::MAX_EVENT_TOPICS;
+use zk_ee::storage_types::MAX_EVENT_TOPICS;
 use zk_ee::system::errors::subsystem::SubsystemError;
 use zk_ee::system::errors::{runtime::RuntimeError, system::SystemError};
 use zk_ee::system::CallModifier;
--- a/system_hooks/src/l2_base_token.rs
+++ b/system_hooks/src/l2_base_token.rs
@@ -18,7 +18,7 @@ use core::fmt::Write;
 use ruint::aliases::{B160, U256};
 use zk_ee::execution_environment_type::ExecutionEnvironmentType;
 use zk_ee::internal_error;
-use zk_ee::kv_markers::MAX_EVENT_TOPICS;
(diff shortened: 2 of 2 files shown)


system_hooks/src/l2_base_token.rs
```diff
@@ -140,206 +155,317 @@ where
...
+    // - nominal_token_value (32)
+
+    // So the padded message will be 64 bytes long.
+    // Total length of the encoded message will be 32 + 32 + 64 = 128 bytes.
+    let mut l1_messenger_calldata = [0u8; 128];
+    l1_messenger_calldata[31] = 32; // offset
+    l1_messenger_calldata[63] = 56; // length
+    l1_messenger_calldata[64..68].copy_from_slice(FINALIZE_ETH_WITHDRAWAL_SELECTOR);
+    // check that first 12 bytes in address encoding are zero
+    // TODO: should we fail or silently truncate? Solidity would truncate iirc.
```
0xVolosnikov, line 226: @AntonD3 probably we should change it. Wdyt?

system_hooks/src/l2_base_token.rs
```diff
@@ -140,206 +155,317 @@ where
...
+    };
+    if calldata_len < message_end {
+        return Ok(Err(
+            "L2 base token failure: withdrawWithMessage called with invalid calldata",
+        ));
+    }
+    let additional_data = &calldata[(length_encoding_end as usize)..message_end as usize];
+
+    // check that first 12 bytes in address encoding are zero
+    // TODO: should we fail or silently truncate? Solidity would truncate iirc.
```
0xVolosnikov, line 341: @AntonD3 same here

basic_bootloader/src/bootloader/run_single_interaction.rs
```diff
@@ -55,6 +64,22 @@ where
                 }
             })?;
 
+        // Emit mint event
```
0xVolosnikov, line 67: I don't like it and it would be very nice to not emit this event. Basically this is completely unexpected by EVM - this is more or less okay I guess, since it only happens in our specific type of transactions (l1->l2, upgrade).

0xVolosnikov:
Note: this is a breaking change, should not be merged for now

system_hooks/src/l2_base_token.rs
AntonD3: I think it makes sense to follow solidity abi decoder behavior here, worth to check

antoniolocascio (approved):
LGTM, the cost of copying in the event for withdrawWithMessage needs to be accounted for in #318

PR #317, opened 2025-10-06
fix: Replace gas_per_pubdata with pubdata_price in block context

antoniolocascio:
## What ❔
Replaces the gas_per_pubdata field of the block context with pubdata_price

## Why ❔
It simplifies the charging for pubdata, as before we were considering the priority fee in the calculation of native_per_pubdata.

## Is this a breaking change?
- [x] Yes
- [ ] No

## Checklist


basic_bootloader/src/bootloader/transaction/mod.rs
```diff
@@ -244,6 +244,16 @@ impl<'a> ZkSyncTransaction<'a> {
             _ => return Err(()),
         }
 
+        // gas_per_pubdata_limit should be zero for non L1 transactions
+        match tx_type {
+            Self::UPGRADE_TX_TYPE | Self::L1_L2_TX_TYPE => {}
+            _ => {
+                if self.gas_per_pubdata_limit.read() != 0 {
+                    return Err(());
```
0xVolosnikov, line 252: We should reflect it in docs, not obvious during debugging

0xVolosnikov (approved):

PR #318, opened 2025-10-07
fix!: charge gas in system contract hooks

antoniolocascio:
## What ❔
This PR changes how gas is charged in system hooks. Now we charge an amount of gas proportional to the native consumed. 

## Is this a breaking change?
- [x] Yes
- [ ] No

## Checklist

### DIFF
--- a/system_hooks/src/l2_base_token.rs
+++ b/system_hooks/src/l2_base_token.rs
@@ -436,7 +436,7 @@ where
         abi_encoded_event_length
     };
 
-    // First we charge for copying the event data
+    // Now we charge for copying the event data
     let native_copy_cost = evm_interpreter::native_resource_constants::COPY_BASE_NATIVE_COST
         .saturating_add(
             evm_interpreter::native_resource_constants::COPY_BYTE_NATIVE_COST


system_hooks/src/lib.rs
```diff
@@ -319,3 +320,58 @@ fn make_return_state_from_returndata_region<S: SystemTypes>(
...
+///
+/// Average value for the ratio between native and gas.
+/// We do not rely on it for anything critical,
+/// If a call to a given hook consumes N native, we also charge
+/// N / AVERAGE_NATIVE_PER_GAS_FOR_HOOKS gas. That way,
+/// we ensure that the caller can bound the computation cost of the call.
+///
+const AVERAGE_NATIVE_PER_GAS_FOR_HOOKS: u64 = 60;
+
+// Base cost for calling into a system hook
```
AntonD3, line 333:
```suggestion
/// Base cost for calling into a system hook
```

basic_system/Cargo.toml
```diff
@@ -19,6 +19,7 @@ either = { version = "*", default-features = false }
 storage_models = { path = "../storage_models", default-features = false }
 modexp = { path = "../supporting_crates/modexp" }
 cycle_marker = { path = "../cycle_marker" }
+system_hooks = { path = "../system_hooks" }
```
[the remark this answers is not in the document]
antoniolocascio: I wanted to avoid doing that, as `charge_native_and_proportional_gas` is a bit hacky. But sure, I can do that for now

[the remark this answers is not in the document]
antoniolocascio: Ah it also depends on ERGS_PER_GAS, which is defined on evm_interpreter, so I can't move to zk_ee...

evm_interpreter/src/lib.rs
```diff
@@ -474,3 +475,48 @@ impl From<InternalError> for ExitCode {
...
+pub fn post_charge_proportional_gas<R: Resources>(
+    resources: &mut R,
+    resources_before_hook: R,
+) -> Result<(), SystemError> {
+    use zk_ee::system::Computational;
+    let native_used = resources
+        .native()
+        .diff(resources_before_hook.native())
+        .as_u64();
+    let proportional_ergs_from_native = native_used
```
0xVolosnikov: Probably we have signed division here for some reason and binary verification fails because of that

system_hooks/src/contract_deployer.rs
```diff
@@ -192,3 +198,21 @@ where
         _ => Ok(Err("Contract deployer hook: unknown selector")),
     }
 }
+
+///
+/// We add some ergs cost to account for work charged in native only.
```
AntonD3: Can we include it during `set_bytecode_details` call?

system_hooks/src/l1_messenger.rs
```diff
@@ -273,3 +282,23 @@ pub(crate) fn send_to_l1_inner<S: EthereumLikeTypes>(
 
     Ok(Ok(message_hash))
 }
+
+///
+/// Ergs cost of emitting an L1 message.
```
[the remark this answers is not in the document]
antoniolocascio: Same as the other reply, happy to change it, but wanted to make it as easy as possible to determine the gas cost from the hook implementation.

0xVolosnikov:
To not forget: https://github.com/matter-labs/zksync-os/pull/313#pullrequestreview-3363538267

system_hooks/src/contract_deployer.rs
antoniolocascio: Yes, see line 216

system_hooks/src/l1_messenger.rs
```diff
@@ -264,8 +273,8 @@ pub(crate) fn send_to_l1_inner<S: EthereumLikeTypes>(
     topics.push(message_hash);
 
     system.io.emit_event(
-        ExecutionEnvironmentType::parse_ee_version_byte(caller_ee)
-            .map_err(SystemError::LeafDefect)?,
+        // Use EVM to charge gas for this operation
```
0xVolosnikov: Note: here it's becoming obvious why we need to replace this hook with EVM contract

PR #325, opened 2025-10-15
fix: add assertion checking initial slot is empty

antoniolocascio:
## What ❔
Add assertion checking initial slot is empty

## Is this a breaking change?
- [ ] Yes
- [X] No

## Checklist


0xVolosnikov (changes requested):
We need a regression test

PR #327, opened 2025-10-16
fix!: get rid of side effects produced by invalid transactions

antoniolocascio:
## What ❔
This PR fixes 2 side-effects caused by invalid transactions:
1. Native charging for IO: rollbacks didn't purge slots from the cache even if the initial access was reverted. This meant that if an invalid tx touched an address than then was touched by a valid one, the latter wouldn't perform some native charging. This was fixed by purging the slots after "uninitializing". In addition, this PR also makes invalid transactions roll back to the handler taken before the tx.
2. The tx counter (used to index logs) was being increased even for invalid transactions, this PR fixes that.

## Why ❔

## Is this a breaking change?
- [X] Yes
- [ ] No

## Checklist

### DIFF
--- a/tests/instances/transactions/src/lib.rs
+++ b/tests/instances/transactions/src/lib.rs
@@ -632,6 +632,149 @@ fn test_cold_in_new_tx() {
     assert!(result2.is_ok_and(|o| !o.is_success()));
 }
 
+#[test]
+fn test_invalid_tx_does_not_affect_native() {
+    let wallet = PrivateKeySigner::from_str(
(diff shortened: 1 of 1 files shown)


antoniolocascio:
@0xVolosnikov yeah, but I don't think we have much of a choice. We can in the future redesign the charging to make it always the same no matter the cache, but for now I would accept the performance hit

0xVolosnikov:
Can we split this PR into 2 PRs? 
- IO
- Transaction number

PR #328, opened 2025-10-16
fix!: don't charge warm gas for selfdestruct with precompile as beneficiary

antoniolocascio:
## What ❔

Precompiles are always warm, but we have a corner case when we charge for the cold access (during selfdestruct)

## Is this a breaking change?
- [X] Yes
- [ ] No

## Checklist


0xVolosnikov (changes requested):
We need a regression test

0xVolosnikov (approved):

PR #332, opened 2025-10-17
fix!: Extcodehash for empty accounts

vladbochok:
## What ❔

Per [EIP-161](https://eips.ethereum.org/EIPS/eip-161): An account is considered empty when it has no code and zero nonce and zero balance.
Per [EIP-1052](https://eips.ethereum.org/EIPS/eip-1052): In case the account does not exist or is empty (as defined by [EIP-161](https://eips.ethereum.org/EIPS/eip-161)) 0 is pushed to the stack.

Currently empty account may have either `0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470` or 0 bytecodehash. But according to EIPs it should be zero. I found this inconsistency by loading extbytecodehash of `address(0)` and revm/zksync os state mismatched.

## Why ❔

It is a small inconsistency in ZKsync OS EVM execution.

## Is this a breaking change?
- [X] Yes
- [ ] No

## Checklist


0xVolosnikov:
Also requires a regression test

PR #345, opened 2025-10-22
fix!: remove initial access from cache during rollbacks

antoniolocascio:
## What ❔
This PR fixes a side-effect caused by invalid transactions related to native charging for IO. Rollbacks didn't purge slots from the cache even if the initial access was reverted. This meant that if an invalid tx touched an address than then was touched by a valid one, the latter wouldn't perform some native charging. This was fixed by purging the slots after "uninitializing". In addition, this PR also makes invalid transactions roll back to the handler taken before the tx.

## Why ❔

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


antoniolocascio:
Closed in favor if #366

PR #347, opened 2025-10-22
fix: add UARTPrintResponder processor in generate_proof_input

perekopskiy:
## What ❔

add UARTPrintResponder processor in generate_proof_input

## Why ❔

generate_proof_input should work with logging_enabled bins

## Is this a breaking change?
- [ ] Yes
- [X] No

## Checklist


0xVolosnikov:
lgtm, but probably it would be better to make it configurable - `if debugging { oracle.add_external_processor ... }`, wdyt?

perekopskiy:
Do you want to add flag because this processor affects performance or smth else?
idk, `debugging` looks a bit extra, currently server just passes a path to server_app_logging_enabled and that's it, but it can pass a bool as well if needed

0xVolosnikov:
> because this processor affects performance or smth else?

I don't think that it affects performance, but maybe we can end up in some strange situations when we have the processor in one place and don't have in actual proving, so it'll diverge etc.

0xVolosnikov:
Probably this is not a real problem honestly, so feel free to decide

PR #355, opened 2025-10-27
fix: Reimplement EVM storage refunds (H4)

0xVolosnikov:
## What ❔

Fix for EVM gas storage refunds calculation

## Why ❔

Previous implementation was incorrect

## Is this a breaking change?
- [ ] Yes
- [X] No

## Checklist


0xVolosnikov:
Unfortunately we do not check EVM refunds in spec tests at all. And we hardly check gas as well.

https://github.com/matter-labs/zksync-os/blob/8f01c1fad0bfd253634d9f1813716f0d5f833d78/tests/evm_tester/src/test/case/mod.rs#L536-L543

PR #206, opened 2025-08-12
perf: copy .rodata to RAM

yoaveshel:
## What ❔
Copy all of .rodata section to RAM

## Why ❔
So we can remove ROM checks in https://github.com/matter-labs/zksync-os/pull/131

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.
- [ ] Code has been formatted.

PR #358, opened 2025-10-28
feat: remove `init` functions and statics from crypto crate

yoaveshel:
## What ❔
Previously we needed to keep all constants used for delegation in mutable statics and init them at runtime. Since https://github.com/matter-labs/zksync-os/pull/206 this is no longer needed. This PR removes the init functions

## Why ?
free performance!

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/crypto/src/bigint_delegation/u256.rs
+++ b/crypto/src/bigint_delegation/u256.rs
@@ -172,7 +172,7 @@ pub fn is_zero(a: &U256) -> bool {
 
 #[inline(always)]
 /// it takes `a` as mutable for the purposes of delegation calls, but doesn't mutate it
-pub fn is_one(a: &mut U256) -> bool {
+pub fn is_one(a: &U256) -> bool {
     delegation::eq(a, &ONE) != 0
 }
 
--- a/crypto/src/bls12_381/fields/fr.rs
+++ b/crypto/src/bls12_381/fields/fr.rs
@@ -147,7 +147,7 @@ fn __gcd_inverse(a: &Fr) -> Option<Fr> {
     let mut c = Fp::zero();
     let modulus = Fr::MODULUS;
 
-    while !u256::is_one(&mut u) && !u256::is_one(&mut v) {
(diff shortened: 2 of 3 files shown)


crypto/src/bigint_delegation/u256.rs
```diff
@@ -1,28 +1,17 @@
...
-        SCRATCH.as_mut_ptr().write(U256::zero());
-    }
-}
+static mut COPY_PLACE_0: U256 = U256::zero();
+static mut COPY_PLACE_1: U256 = U256::zero();
+static mut COPY_PLACE_2: U256 = U256::zero();
+static mut COPY_PLACE_3: U256 = U256::zero();
+static mut SCRATCH: U256 = U256::zero();
+
+const ONE: U256 = U256::one();
```
shamatar: Technically we only use `&ONE`, so we can put it as static here to avoid relying on Rust rules for consts vs statics (for refs to consts)

crypto/src/bigint_delegation/delegation.rs
```diff
@@ -55,7 +30,8 @@ pub(super) fn mul_high(a: &mut U256, b: &U256) {
 }
 
 #[inline(always)]
-pub(super) fn eq(a: &mut U256, b: &U256) -> u32 {
+pub(super) fn eq(a: &U256, b: &U256) -> u32 {
```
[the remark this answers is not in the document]
yoaveshel: in this case I don't think it matters too much, and this signature makes more sense as `eq` shouldn't mutate `a`. But nothing heavily depends on it so I can change it back

crypto/src/bigint_delegation/u256.rs
yoaveshel: good catch

shamatar:
Interesting behavior of point-eval

shamatar:
One more comment above, otherwise looks fine

PR #361, opened 2025-10-28
fix: Fix allocated capacity in from_big_endian_with_double_capacity_or_min_capacity

0xVolosnikov:
## What ❔

Fixes small issue

## Why ❔

The min_capacity should be used

## Is this a breaking change?
- [ ] Yes
- [X] No

## Checklist

### DIFF
--- a/basic_system/src/system_functions/modexp/delegation/bigint.rs
+++ b/basic_system/src/system_functions/modexp/delegation/bigint.rs
@@ -151,7 +151,7 @@ impl<A: Allocator + Clone> BigintRepr<A> {
         allocator: A,
     ) -> Self {
         if bytes.is_empty() {
-            let backing = Vec::new_in(allocator);
+            let backing = Vec::with_capacity_in(min_capacity, allocator);
             return Self { backing, digits: 0 };
         }
         let (remainder, digits_bytes) = bytes.as_rchunks::<32>();


basic_system/src/system_functions/modexp/delegation/bigint.rs
```diff
@@ -151,7 +151,7 @@ impl<A: Allocator + Clone> BigintRepr<A> {
         allocator: A,
     ) -> Self {
         if bytes.is_empty() {
-            let backing = Vec::new_in(allocator);
+            let backing = Vec::with_capacity_in(min_capacity, allocator);
```
shamatar: It is mechanical, but not sure that it makes sense or reachable. If it’s used for modulus or base, then no computation happens and it’s not used

0xVolosnikov: Also not sure if it's used. Just an edge case pointed by auditors - `from_big_endian_with_double_capacity_or_min_capacity` implies that we should use `min_capacity` here.

antoniolocascio (approved):

basic_system/src/system_functions/modexp/delegation/bigint.rs
antoniolocascio: I think it makes sense to merge. Even if not used now, the function might be used in the future for something that expects the backing vector to be allocated with min capacity.

PR #365, opened 2025-10-28
fix: Add fuzz tests for Forward-Proving divergency

mikhail-egorov:
## What ❔
This PR adds new fuzz targets for precompiles (ecadd and sha256) to detect Forward–Proving divergences without relying on the RISC-V simulator (which slows fuzzing). It also introduces small changes to the crypto and basic_system crates to allow the fuzzer to run without compiling for the RISC-V architecture. For example, the `cfg` is updated from:

```rust
#[cfg(not(all(feature = "single_round_with_control", target_arch = "riscv32")))]
```

to:

```rust
#[cfg(not(any(
    all(feature = "single_round_with_control", target_arch = "riscv32"),
    all(feature = "proving", fuzzing)
)))]
```

To run Forward and Proving modes from the same fuzz target, this PR adds lightweight wrappers for ZKSync OS crates (crypto, basic_system, zk_ee, etc.). These wrappers live under `fuzz/wrappers/`.

## Is this a breaking change?
- [ ] Yes
- [X] No

## Checklist

### DIFF
--- a/basic_system/src/system_functions/point_evaluation/consts.rs
+++ b/basic_system/src/system_functions/point_evaluation/consts.rs
@@ -56,10 +56,7 @@ pub const G2_BY_TAU_POINT:
         infinity: false,
     };
 
-#[cfg(any(
-    all(target_arch = "riscv32", feature = "proving"),
-    all(feature = "proving", fuzzing)
-))]
+#[cfg(all(feature = "proving", any(target_arch = "riscv32", fuzzing)))]
 pub const G2_BY_TAU_POINT:
     <crypto::bls12_381::curves::Bls12_381 as crypto::ark_ec::pairing::Pairing>::G2Affine =
     crypto::bls12_381::curves::g2::G2Affine {


basic_system/src/system_functions/point_evaluation/consts.rs
```diff
@@ -53,7 +56,10 @@ pub const G2_BY_TAU_POINT:
         infinity: false,
     };
 
-#[cfg(all(target_arch = "riscv32", feature = "proving"))]
+#[cfg(any(
+    all(target_arch = "riscv32", feature = "proving"),
+    all(feature = "proving", fuzzing)
+))]
```
antoniolocascio: I feel like this is simpler to understand, no? Not sure I added the right amount of closing ) btw
```suggestion
#[cfg(all(feature = "proving", any(target_arch = "riscv32", fuzzing)))]
```

antoniolocascio:
@mikhail-egorov should we add these targets to the fuzz.sh script? Or is that a follow-up? Otherwise LGTM

antoniolocascio (approved):

PR #366, opened 2025-10-29
fix!: make native charging independent of caches state

antoniolocascio:
## What ❔

This PR fixes a side-effect caused by invalid transactions related to native charging for IO. If an invalid tx touched an address than then was touched by a valid one, the latter wouldn't perform some native charging. This PR fixes it by making native charging in the caches independent of the actual state of the caches. In addition, this PR also makes invalid transactions roll back to the handler taken before the tx.

## Why ❔

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/basic_system/src/system_implementation/flat_storage_model/account_cache.rs
+++ b/basic_system/src/system_implementation/flat_storage_model/account_cache.rs
@@ -90,7 +90,7 @@ impl<
         }
     }
 
-    fn charge_for_cold_access(
+    fn charge_ergs_for_cold_access(
         ee_type: ExecutionEnvironmentType,
         resources: &mut R,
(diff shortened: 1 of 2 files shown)


0xVolosnikov (approved):
lgtm. But would be nice to think about and maybe add some tests for potential edge cases (maybe with selfdestruct?) - maybe in a separate PR. 

Side note: maybe some more comments in this block could be useful for reviewers (it's not trivial to follow): https://github.com/matter-labs/zksync-os/pull/366/files#diff-5fde3be654a4bd8e9c1f22099afe7a40d7e95dd8b1b3270ea7ba2fde712e8107L131

PR #367, opened 2025-10-29
feat: support blobs da commitment scheme

AntonD3:
## What ❔

This PR implements support for EIP-4844 blob-based data availability (DA) commitment scheme in ZKsync OS.

Key Changes:
- New DA Commitment Scheme: Introduces `DACommitmentScheme::BlobsZKsyncOS` for blob-based data availability
- Blob Commitment Generator: Implements [BlobCommitmentGenerator](https://github.com/matter-labs/zksync-os/blob/103b7796068b475b34c7b0e0f31d7f10b7718320/basic_system/src/system_implementation/system/da_commitment_generator/blob_commitment_generator/mod.rs#L48) that encodes pubdata into EIP-4844 blobs using 31-byte chunks
- KZG Point Evaluation: the point evaluation logic system function was restructured to be reused during blob commitment verification
- Oracle Integration: New blob commitment and proof oracle for external blob advice. It's used to get the commitment and the proof from the external source, to not spend cycles inside ZKsync OS on calculations
- Testing: Test coverage including unit tests, fuzz testing, and integration tests

Key Features for BlobsZKsyncOS DA commitment scheme
- **Blob Encoding**: Encodes pubdata into blobs with 31-byte chunks, supporting up to 9 blobs
- **EIP-4844 Versioned Hashes**: Generates KZG versioned hashes with version byte 0x01
- **Cryptographic Verification**: Full KZG proof verification using BLS12-381 curve
- **Performance Optimization**: [Pre-computed bit-reversed roots of unity](https://github.com/matter-labs/zksync-os/blob/103b7796068b475b34c7b0e0f31d7f10b7718320/basic_system/src/system_implementation/system/da_commitment_generator/blob_commitment_generator/brp_roots_of_unity.rs) for polynomial operations

Breaking changes:
- Updated proof run interfaces to support different DA commitment schemes

## Is this a breaking change?
- [ ] Yes
- [X] No (backward compatible with previous versions)

## Checklist

### DIFF
--- a/forward_system/src/run/mod.rs
+++ b/forward_system/src/run/mod.rs
@@ -134,7 +134,9 @@ pub fn generate_proof_input<T: ReadStorageTree, PS: PreimageSource, TS: TxSource
     oracle.add_external_processor(preimage_responder);
     oracle.add_external_processor(tree_responder);
     oracle.add_external_processor(callable_oracles::arithmetic::ArithmeticQuery::default());
-    oracle.add_external_processor(callable_oracles::blob_kzg_commitment::BlobCommitmentAndProofQuery::default());
+    oracle.add_external_processor(
+        callable_oracles::blob_kzg_commitment::BlobCommitmentAndProofQuery::default(),
+    );
     oracle.add_external_processor(UARTPrintResponder);
 
     // We'll wrap the source, to collect all the reads.
@@ -364,7 +366,9 @@ pub fn run_block_with_oracle_dump_ext<
     oracle.add_external_processor(zk_proof_data_responder);
     oracle.add_external_processor(da_commitment_scheme_responder);
     oracle.add_external_processor(callable_oracles::arithmetic::ArithmeticQuery::default());
-    oracle.add_external_processor(callable_oracles::blob_kzg_commitment::BlobCommitmentAndProofQuery::default());
+    oracle.add_external_processor(
+        callable_oracles::blob_kzg_commitment::BlobCommitmentAndProofQuery::default(),
+    );
     oracle.add_external_processor(UARTPrintResponder);
 
     let mut result_keeper = ForwardRunningResultKeeper::new(tx_result_callback);
@@ -403,7 +407,9 @@ pub fn run_block_from_oracle_dump<
     oracle.add_external_processor(zk_proof_data_responder);
     oracle.add_external_processor(da_commitment_scheme_responder);
     oracle.add_external_processor(callable_oracles::arithmetic::ArithmeticQuery::default());
-    oracle.add_external_processor(callable_oracles::blob_kzg_commitment::BlobCommitmentAndProofQuery::default());
+    oracle.add_external_processor(
+        callable_oracles::blob_kzg_commitment::BlobCommitmentAndProofQuery::default(),
+    );
 
     let mut result_keeper = ForwardRunningResultKeeper::new(NoopTxCallback);
 


basic_system/src/system_implementation/system/da_commitment_generator/blob_commitment_generator/commitment_and_proof_advice.rs
```diff
@@ -0,0 +1,63 @@
...
+#[cfg(not(target_arch = "riscv32"))]
+pub fn blob_commitment_and_proof_advice(
+    data: &[u8],
+    _oracle: &mut impl zk_ee::oracle::IOOracle,
+) -> [u8; 96] {
+    use crypto::MiniDigest;
+
+    let sidecar_builder: alloy_consensus::SidecarBuilder<alloy_consensus::SimpleCoder> =
+        alloy_consensus::SidecarBuilder::from_slice(data);
+    // TODO: at this step we compute also kzg proof, which is not needed in fact
```
antoniolocascio: Should we avoid it now? Or as a follow-up?

basic_system/src/system_implementation/system/da_commitment_generator/blob_commitment_generator/commitment_and_proof_advice.rs
```diff
@@ -0,0 +1,63 @@
...
+    hasher.update(data);
+    let mut challenge_point = hasher.finalize();
+    // truncate hash to 128 bits
+    // NOTE: it is safe to draw a random scalar at max 128 bits because of the schwartz zippel lemma
+    for byte in challenge_point[0..16].iter_mut() {
+        *byte = 0;
+    }
+    let blob =
+        unsafe { core::mem::transmute::<&alloy_consensus::Blob, &c_kzg::Blob>(&sidecar.blobs[0]) };
+    let p = c_kzg::ethereum_kzg_settings(8)
```
antoniolocascio: Should we cache the call to `ethereum_kzg_settings(8)`. IIUC this is called for every block, right? I guess for forward it's not that bad, but maybe we can add a TODO for it?

basic_system/src/system_implementation/system/da_commitment_generator/blob_commitment_generator/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+/// Number of field elements per EIP-4844 blob
+///
+pub const ELEMENTS_PER_4844_BLOCK: usize = 4096;
+
+///
+/// Number of bytes we can encode in one blob
+///
+pub const ENCODABLE_BYTES_PER_BLOB: usize = BLOB_CHUNK_SIZE * ELEMENTS_PER_4844_BLOCK;
+
+pub const MAX_NUMBER_OF_BLOBS: usize = 9;
```
antoniolocascio: This will be confusing when we merge #381 , can we rename to something like `L1_MAX_NUMBER_OF_BLOBS`?

basic_system/src/system_implementation/system/da_commitment_generator/blob_commitment_generator/polynomial_evaluation.rs
```diff
@@ -0,0 +1,85 @@
...
+        poly_iter
+            .next()
+            .unwrap()
+            .write(crypto::bls12_381::Fr::from_bigint(parse_u256_be(chunk)).unwrap());
+    }
+    poly_iter
+        .next()
+        .unwrap()
+        .write(crypto::bls12_381::Fr::from_bigint(parse_u256_be(&last_chunk)).unwrap());
+    let poly = unsafe { MaybeUninit::array_assume_init(poly) };
```
antoniolocascio: Also, is it ok to assume init? Should we 0 out the poly? Do we really need to use MaybeUninit here?

callable_oracles/src/blob_kzg_commitment/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+        Box::new(n)
+    }
+}
+
+///
+/// Calculate kzg commitment and proof at the point `blake2s(versioned_hash & data)` for blob created from passed data.
+///
+/// For encoding, we chunk `data` by 31 bytes and interpret each chunk as BE blob element.
+///
+pub fn blob_kzg_commitment_and_proof(data: &[u8]) -> [u8; 96] {
```
antoniolocascio: I'm a bit confused, why do we need both this and the `blob_commitment_and_proof_advice` for x86?

basic_system/src/system_implementation/system/da_commitment_generator/blob_commitment_generator/commitment_and_proof_advice.rs
AntonD3: Oh, wait, this TODO shouldn't be here indeed, I have to update this function generally, thanks for flagging

basic_system/src/system_implementation/system/da_commitment_generator/blob_commitment_generator/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+    //     self.pubdata_buffer.try_extend_from_slice(filling_part).unwrap();
+    //
+    //     cycle_marker::wrap!("blob_versioned_hash", {
+    //         self.versioned_hashes_hasher.update(&blob_versioned_hash(self.pubdata_buffer.as_slice()));
+    //     });
+    //     self.pubdata_buffer.clear();
+    //     // theoretically remainder can be still bigger than buffer_capacity,
+    //     // so we are making call to the `write` again to handle it recursively
+    //     self.write(remainder);
+    // }
```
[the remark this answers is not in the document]
AntonD3: Yes, we already decided to postpone input generation change, so makes sense to clean it up

AntonD3:
@antoniolocascio, agree about testing, can you point me to some fuzz tests as an example?

Regarding integration tests I added a e2e test in the multiblock-batch, also I was thinking about unit tests to cover many blobs. Do think it's enough? Any other ideas?

antoniolocascio:
Yes, unit tests for encoding would be nice. For fuzzing: not sure how many implementations we'll end up having. In the current style, I would fuzz consistency of the `blob_commitment_and_proof_advice` on native vs the callable oracle one. If we unify this, then maybe just smoke fuzzing the encode (with different size datas) or even take random data -> encode -> check the point eval. Basically just feed random data to `blob_versioned_hash()` and see if it panics. The random data should cover interesting cases for lengths, but maybe we can have those as unit tests

zk_ee/src/common_structs/da_commitment_scheme.rs
```diff
@@ -0,0 +1,69 @@
...
+            4 => Ok(DACommitmentScheme::BlobsZKsyncOS),
+            _ => Err(()),
+        }
+    }
+}
+
+impl UsizeSerializable for DACommitmentScheme {
+    const USIZE_LEN: usize = <u8 as UsizeSerializable>::USIZE_LEN;
+
+    fn iter(&self) -> impl ExactSizeIterator<Item = usize> {
```
0xVolosnikov: Maybe it would be better to always explicitly convert it to/from `u8`? This code duplication here looks ugly

crypto/src/lib.rs
```diff
@@ -95,3 +95,46 @@ pub trait MiniDigest: Sized {
...
+pub fn parse_u256_be<const N: usize>(input: &[u8; N]) -> BigInt<4> {
+    assert!(N <= 32);
+    // Arkworks has strange format for integer serialization, so we do manually
+    let mut repr = [0u64; 4];
+    let mut repr_iter = repr.iter_mut();
+    let (remainder, chunks) = input.as_rchunks::<8>();
+    for chunk in chunks.iter().rev() {
+        *repr_iter.next().unwrap() = u64::from_be_bytes(*chunk);
+    }
+    if remainder.len() != 0 {
```
0xVolosnikov: Clippy

tests/instances/system_hooks/src/lib.rs
```diff
@@ -109,7 +109,7 @@ fn test_set_deployed_bytecode_evm_unauthorized() {
     };
     let transactions = vec![encoded_tx];
 
-    let output = chain.run_block(transactions, None, None);
+    let output = chain.run_block(transactions, None, None, None);
```
0xVolosnikov: We should fix other tests too. Probably some new tests after rebase

basic_system/src/system_functions/point_evaluation/mod.rs
```diff
@@ -77,7 +77,7 @@ fn parse_g1_compressed(input: &[u8]) -> Result<crypto::bls12_381::G1Affine, ()>
     crypto::bls12_381::G1Affine::deserialize_compressed(input).map_err(|_| ())
 }
 
-fn point_evaluation_as_system_function_inner<D: ?Sized + TryExtend<u8>, R: Resources>(
+pub fn point_evaluation_as_system_function_inner<D: ?Sized + TryExtend<u8>, R: Resources>(
```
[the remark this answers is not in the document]
AntonD3: Yes, also thought about it, will take a look

zk_ee/src/common_structs/da_commitment_scheme.rs
AntonD3: > I don't think that this is a good reason honestly

I mean, this is the approach we are using usually in the codebase, even for a similar enum `TxEncodingFormat`. So I made the same to make it consistent.

Although this approach allows us to avoid intermediate types, I don't have a strict opinion on it, but IMHO, if we want to change it, it's worth changing it everywhere, and it's somewhat a separate issue.

forward_system/src/run/mod.rs
```diff
@@ -138,6 +144,64 @@ pub fn generate_proof_input<T: ReadStorageTree, PS: PreimageSource, TS: TxSource
     Ok(std::rc::Rc::try_unwrap(items).unwrap().into_inner())
 }
 
+// TODO: in future we should generate input per batch
```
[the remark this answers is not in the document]
AntonD3: https://linear.app/matterlabs/issue/EVM-1184/generate-proof-input-per-batch

basic_system/src/system_implementation/system/da_commitment_generator/blob_commitment_generator/mod.rs
antoniolocascio: I would prefer something like `DA_MAX_NUMBER_OF_BLOBS`, but we can discuss it when we discuss 4844. What I want to avoid is that we accidentally use this one for the max number of blobs in L2, which can be different (Cancun on L2 vs Pectra L1). I'll resolve for now.

basic_system/src/system_implementation/system/da_commitment_generator/blob_commitment_generator/polynomial_evaluation.rs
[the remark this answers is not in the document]
AntonD3: Yes, you are right, fixed

basic_system/src/system_implementation/system/public_input.rs
```diff
@@ -166,7 +165,9 @@ impl BatchOutput {
         hasher.update(self.chain_id.to_be_bytes::<32>());
         hasher.update(&self.first_block_timestamp.to_be_bytes());
         hasher.update(&self.last_block_timestamp.to_be_bytes());
-        hasher.update(self.used_l2_da_validator_address.to_be_bytes::<20>());
+        // Encode DA commitment scheme as U256 BE
+        hasher.update([0u8; 31]);
+        hasher.update([self.da_commitment_scheme as u8]);
```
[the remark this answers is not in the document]
AntonD3: Updated public input doc

zk_ee/src/common_structs/da_commitment_scheme.rs
0xVolosnikov: Also, to understand better: we don't want to move it into zksync-os-interfaces, right? Corresponding value is passed only by proof input generator?

0xVolosnikov (approved):
Lgtm. Would be nice to check the compatibility of our encoding with Alloy.
Also we can additionally improve fuzzing/test coverage later.

basic_system/src/system_implementation/system/da_commitment_generator/blob_commitment_generator/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+
+fn versioned_hash_for_kzg(data: &[u8]) -> [u8; 32] {
+    use crypto::sha256::Digest;
+    let mut hash: [u8; 32] = crypto::sha256::Sha256::digest(data).into();
+    hash[0] = 1; // KZG_VERSIONED_HASH_VERSION_BYTE
+
+    hash
+}
+
+fn calculate_evaluation_point(data: &[u8], versioned_hash: &[u8]) -> crypto::bls12_381::Fr {
```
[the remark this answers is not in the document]
AntonD3: Going to resolve and merge, but feel free to raise it once again, we can make follow up PR

AntonD3:
> Point evaluation native usage increased by 1.5m, will check

I am merging PR to not block the release; performance degradation can be investigated later

PR #418, opened 2025-11-21
feat!: prover input run on native arch

antoniolocascio:
## What ❔
This PR changes the way we compute prover input. Before, we needed to run the slow risc-v simulator to do so.
Now, we can do on native architecture.
This run is also responsible for computing the pubdata.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


basic_system/src/system_functions/modexp/mod.rs
```diff
@@ -221,12 +229,23 @@ fn modexp_as_system_function_inner<
     );
 
     #[cfg(not(any(all(target_arch = "riscv32", feature = "proving"), test)))]
-    let output = ::modexp::modexp(
-        base.as_slice(),
-        exponent.as_slice(),
-        modulus.as_slice(),
-        allocator,
-    );
+    let output = if USE_DELEGATION {
```
AntonD3: Can we run delegation implementation depending on proving feature, so we can remove this extra argument?

And actually maybe it makes sense to rename delegation implementation to something like "proving implementation"/"advice_implementation"? As I understand it uses delegation depending on target

forward_system/src/run/output.rs
```diff
@@ -134,12 +135,21 @@ impl<TR: TxResultCallback> From<ForwardRunningResultKeeper<TR>> for BlockOutput
             storage_writes,
             account_diffs,
             published_preimages,
-            pubdata,
+            // TODO: should we define 2 types of output?
```
AntonD3: I would remove pubdata from block output, and for now return pubdata as separate value(not in block output). In future we'll probably have `BatchOutput`

proof_running_system/src/system/bootloader.rs
```diff
@@ -179,17 +178,13 @@ pub fn run_proving_inner<
     let _ = L::default().write_fmt(format_args!("IO implementer init is complete"));
 
     // Load all transactions from oracle and apply them.
-    let (mut oracle, public_input) = ProvingBootloader::<O, L>::run_prepared::<
-        BasicBootloaderProvingExecutionConfig,
-    >(oracle, &mut NopResultKeeper, &mut NopTracer::default())
-    .expect("Tried to prove a failing batch");
-
-    // disconnect oracle before returning, if some other post-logic is needed that doesn't use Oracle trait
```
AntonD3: So, we don't need disconnect now?

AntonD3 (commented):
Overall this looks great, left a few minor comments and questions

basic_system/src/system_functions/modexp/mod.rs
antoniolocascio: I wanted to avoid using the feature, as we want to use the basic_system crate in both modes for the same binary. Otherwise, we'll need to somehow import the crate twice for server to run both the sequencing system and the prover input gen. 

The naming is just because it uses the modexp::delegation implementation, but can rename to advice, it makes more sense.

callable_oracles/src/lib.rs
```diff
@@ -59,3 +59,23 @@ impl UsizeDeserializable for MemoryRegionDescriptionParams {
...
 }
+
+#[inline(always)]
+unsafe fn read_u64_words(ptr_u64: u64, len_words_u64: u64) -> Vec<u64> {
+    if ptr_u64 == 0 || len_words_u64 == 0 {
+        return vec![];
+    }
+    let addr = ptr_u64 as usize;
+    let len_words = len_words_u64 as usize;
+    core::slice::from_raw_parts(addr as *const u64, len_words).to_vec()
```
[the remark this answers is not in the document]
antoniolocascio: Let me try, I agree

forward_system/src/run/output.rs
antoniolocascio: Removed it, had to use a new branch for interface

basic_system/src/system_functions/mod.rs
```diff
@@ -29,9 +29,9 @@ fn bytereverse(input: &mut [u8]) {
 /// No std system functions implementations.
 /// All of them are following EVM specs(for precompiles and keccak opcode).
 ///
-pub struct NoStdSystemFunctions;
+pub struct NoStdSystemFunctions<const USE_ADVICE: bool>;
```
AntonD3: nit: maybe worth to mentioned that `USE_ADVICE` affects only forward system

basic_system/src/system_functions/modexp/mod.rs
```diff
@@ -25,29 +25,32 @@ pub const MODEXP_ADVICE_QUERY_ID: u32 = ADVICE_SUBSPACE_MASK | 0x10;
...
 #[derive(Debug, Default)]
-pub struct ModExpAdviceParams {
-    pub op: u32,          // Operation type (0 = division)
-    pub a_ptr: u32,       // Pointer to dividend
-    pub a_len: u32,       // Length of dividend in words
-    pub b_ptr: u32,       // Pointer to divisor
-    pub b_len: u32,       // Length of divisor in words
-    pub modulus_ptr: u32, // Pointer to modulus
-    pub modulus_len: u32, // Length of modulus in words
+pub struct ModExpAdviceParamsGeneric<W> {
```
AntonD3: can we just use usize?

callable_oracles/src/arithmetic/mod.rs
```diff
@@ -1,12 +1,57 @@
...
+                } else {
+                    break;
+                }
+            }
+            &input[..digits]
+        }
+        let quotient = strip_leading_zeroes(&self.quotient);
+        let remainder = strip_leading_zeroes(&self.remainder);
+
+        // account for usize being u64 here
```
AntonD3: nit: should we assert that usize indeed is 64 bit?

AntonD3 (commented):
Overall looks good, there are some things that are not very clear for me around arithmetic advice(asked questions in comments), and please consider that it requires non-trivial integration from the platform side - pubdata is now not available from forward run

Cargo.toml
```diff
@@ -61,8 +61,8 @@ keywords = ["blockchain", "zksync", "zk", "risc-v"]
 categories = ["cryptography"]
 
 [workspace.dependencies]
-zksync_os_evm_errors = { version = "0.0.14", default-features = false }
-zksync_os_interface = { version = "0.0.14", default-features = false }
+zksync_os_evm_errors = { git = "https://github.com/matter-labs/zksync-os-interface", branch = "alocascio-prover-input-gen", default-features =  false }
+zksync_os_interface = { git = "https://github.com/matter-labs/zksync-os-interface", branch = "alocascio-prover-input-gen" }
```
antoniolocascio: Thread to prevent merging without merging this first

0xVolosnikov:
@vv-dev-ai ptal

vv-dev-ai:
[AI-review] Reviewed commit `26ac689f4a9e226bfdac8ac3f8e97ee093d0dbf5` — **COMMENT** (4 inline comments).

callable_oracles/src/lib.rs
```diff
@@ -60,12 +66,55 @@ impl UsizeDeserializable for MemoryRegionDescriptionParams {
...
+    let bytes = unsafe { core::slice::from_raw_parts(ptr, len_bytes) };
+    bytes.to_vec()
+}
+
+#[inline(always)]
+pub(crate) fn read_host_struct<T>(ptr_u64: u64) -> T {
+    let ptr = validate_host_pointer(ptr_u64, core::mem::align_of::<T>());
+
+    // Safety: the pointer was validated to be non-null and aligned for `T`.
+    unsafe { ptr.cast::<T>().read() }
```
vv-dev-ai, line 131: [AI-review] **Medium:** The `// SAFETY:` comment is incomplete. `ptr.read()` requires not just non-null + aligned, but also that the pointed-to memory is initialized, valid for `size_of::<T>()` bytes, and that `T`'s validity invariant is satisfied by the bit pattern at that address. The pointer originates from the caller's own address space (stack-allocated `ModExpAdviceParams64` / `FieldOpsHint64`), so this is currently sound, but the safety comment should document these full preconditions.

Additionally, `read_host_struct` is generic over any `T` with no trait bounds. Adding `T: Copy` (or a custom marker trait) would prevent accidental use with types that have drop glue or non-trivial validity invariants.

Suggested safety comment:
```rust
// Safety: the caller guarantees that `ptr_u64` points to a fully
// initialized value of type `T` within the current process's
// address space. The pointer was validated to be non-null and
// aligned for `T`. The memory is not concurrently mutated.
```

callable_oracles/src/lib.rs
```diff
@@ -60,12 +66,55 @@ impl UsizeDeserializable for MemoryRegionDescriptionParams {
...
+    if len_words_u64 == 0 {
+        return vec![];
+    }
+    let ptr = validate_host_pointer(ptr_u64, core::mem::align_of::<u64>());
+    let len_bytes = checked_byte_len(len_words_u64, core::mem::size_of::<u64>(), max_bytes);
+    let len_words = len_bytes / core::mem::size_of::<u64>();
+
+    // Safety: the caller supplied a non-null pointer aligned for `u64`, and
+    // `len_words` was derived from a checked byte-length computation.
+    let words = unsafe { core::slice::from_raw_parts(ptr.cast::<u64>(), len_words) };
```
vv-dev-ai, line 101: [AI-review] **Low:** Same incomplete safety comment pattern as `read_host_struct`. The comment mentions non-null + alignment + bounds but omits the critical requirement that the memory region `[ptr, ptr + len_bytes)` must contain initialized data and remain valid for the duration of the read. This applies to `read_u8_words` at line 110 as well.

oracle_provider/src/lib.rs
```diff
@@ -322,3 +328,85 @@ impl<M: MemorySource> NonDeterminismCSRSource<M> for ReadWitnessSource<M> {
...
+    where
+        I: UsizeSerializable + UsizeDeserializable,
+    {
+        let inner = self.original_source.raw_query(query_type, input)?;
+        // First add the length of the iterator.
+        let len = inner.len();
+        {
+            let mut read_items = self.read_items.borrow_mut();
+            // Len is multiplied by 2 to account for 32/64-bit mismatch
+            let len_u32 = u32::try_from(len * 2).expect("iterator length does not fit into u32");
```
vv-dev-ai: [AI-review] **Low:** `len * 2` could overflow on a pathological oracle response before the `try_from` check catches it. Consider:
```rust
let len_u32 = u32::try_from(len.checked_mul(2).expect("response length overflow"))
    .expect("iterator length does not fit into u32");
```
This makes the multiplication overflow explicit rather than relying on `usize` wrapping (which wouldn't wrap on 64-bit in practice, but is cleaner).

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/post_tx_op_proving_singleblock_batch.rs
```diff
@@ -246,9 +247,18 @@ where
                     state_diffs_hasher.update(value.current_value.as_u8_ref());
                 });
             let state_diffs_hash = state_diffs_hasher.finalize().into();
-            Ok((io.oracle, state_diffs_hash))
+
+            let _ = io
+                .oracle
+                .raw_query_with_empty_input(DISCONNECT_ORACLE_QUERY_ID)
+                .expect("must disconnect an oracle before performing arbitrary CSR access");
+            Ok((io.oracle, state_diffs_hash, batch_output))
```
vv-dev-ai: [AI-review] **Nit:** The `.expect()` on `DISCONNECT_ORACLE_QUERY_ID` is consistent with the previous code in `proof_running_system`, but in proving mode a panic produces an undiagnosable abort. Since the `post_op` function already returns `Result`, consider propagating via `?`:
```rust
let _ = io.oracle
    .raw_query_with_empty_input(DISCONNECT_ORACLE_QUERY_ID)?;
```
Same pattern applies in `post_tx_op_proving_multiblock_batch.rs` and `post_tx_op_proving.rs`.

vv-dev-ai (commented):
[AI-review] PR #418 — **COMMENT**

### Unsafe code documentation

The new host-memory reading helpers in `callable_oracles/src/lib.rs` (`read_host_struct`, `read_u64_words`, `read_u8_words`) have incomplete `// SAFETY:` comments. Per the codebase's unsafe code checklist, every unsafe block must document: pointer validity (points to live, initialized memory of the correct type), alignment, initialization, bounds, and lifetime (memory not freed during read). The current comments mention only non-null and alignment, omitting that the caller must guarantee the pointed-to memory is initialized and remains valid for the duration of the read.

All call sites are sound (pointers originate from stack-allocated structs within the same native process), but the documentation should be complete for future maintainers.

### Suggested Tests

- **Differential test: native vs RISC-V prover input for modexp with large inputs** — The modexp advice path has the most complex 32/64-bit adaptation (packed lengths, digit size conversion). A test with multi-limb modexp inputs (e.g., 2048-bit modulus) exercising `NativeArithmeticQuery` would strengthen coverage of the bigint digit size conversion logic. The existing `native_arithmetic_query_processes_valid_query` test uses only single-limb inputs.
- **Edge case: modexp with `q_len` or `r_len` that is odd in u32 words** — The native path asserts `q_len.is_multiple_of(2) && r_len.is_multiple_of(2)`. A test confirming this assertion fires when the oracle returns odd word counts would document this invariant.

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/post_tx_op_proving_singleblock_batch.rs
antoniolocascio: This moves in the next PR

basic_system/src/system_functions/modexp/advice/bigint.rs
```diff
@@ -857,37 +856,79 @@ impl<'a, O: IOOracle> ModexpAdvisor for OracleAdvisor<'a, O> {
...
-                a_ptr: a_ptr.addr() as u32,
-                a_len: a_len as u32,
-                b_ptr: 0,
-                b_len: 0,
-                modulus_ptr: modulus_ptr.addr() as u32,
-                modulus_len: modulus_len as u32,
-            }
+        // We use different advice params depending on architecture
+        // Both are mostly the same, main difference is the width of pointers
+        #[cfg(target_arch = "riscv32")]
```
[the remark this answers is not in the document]
AntonD3: I understand that in practice we'll not run it on other 32-bit platform, but still technically this code is incorrect, maybe someone else will want to reuse this code, etc. Ofc, it's all weak arguments, but I saw we use `cfg(target_pointer_width = "64")` in other places, so why just not use it here?

crypto/src/bigint_delegation/u256.rs
```diff
@@ -1,5 +1,5 @@
 use super::{delegation, DelegatedBarretParams, DelegatedModParams, DelegatedMontParams};
-use crate::ark_ff_delegation::{BigInt, BigInteger};
+use crate::{BigInt, BigInteger};
```
AntonD3, line 2: Isn't it supposed to use ark_ff_delegation BigInt?

callable_oracles/src/arithmetic/mod.rs
AntonD3: Same here. While it’s practically 99.9% fine, adding a small assertion helps maintain better code quality.

callable_oracles/src/arithmetic/mod.rs
```diff
@@ -1,11 +1,60 @@
...
 use oracle_provider::OracleQueryProcessor;
 use risc_v_simulator::abstractions::memory::MemorySource;
 
 use crate::utils::{
     evaluate::{read_memory_as_u64, read_struct},
     usize_slice_iterator::UsizeSliceIteratorOwned,
 };
+use crate::{read_host_struct, read_u64_words};
+
+const HOST_MODEXP_ADVICE_MAX_BYTES: usize = 32 * 1024 * 1024;
```
AntonD3: Where this const comes from?

crypto/src/bigint_delegation/u256.rs
antoniolocascio: It is whenever we're in a proving environment, see lib.rs in crypto:
```
#[cfg(any(
    all(target_arch = "riscv32", feature = "bigint_ops"),
    feature = "proving",
    test
))]
pub use self::ark_ff_delegation::{BigInt, BigInteger};

#[cfg(not(any(
    all(target_arch = "riscv32", feature = "bigint_ops"),
    feature = "proving",
    test
)))]
pub use self::ark_ff::{BigInt, BigInteger};
```

callable_oracles/src/arithmetic/mod.rs
antoniolocascio: AI convinced me with the other uses cases of this limit, but actually it makes no sense to check this max bytes. This one I ignored as it's big enough for modexp in fusaka, but thinking more about it makes no sense to check this. Removed the checks altogether.

callable_oracles/src/arithmetic/mod.rs
antoniolocascio: Yeah, added a strong assert for this in the lib.rs of callable_oracles, so it should be enough

basic_system/src/system_functions/modexp/advice/bigint.rs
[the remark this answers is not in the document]
antoniolocascio: Done, makes sense

basic_bootloader/src/bootloader/block_flow/ethereum/post_tx_op_proving.rs
```diff
@@ -144,7 +145,12 @@ where
             &metadata.block_level.computed_header_hash
         );
 
-        Ok((io.oracle, metadata.block_level.computed_header_hash))
+        let _ = io
```
0xVolosnikov: Would be nice to use something more abstract - implement SimpleQuery or smth, to not touch internal implementation details here

callable_oracles/src/lib.rs
```diff
@@ -60,12 +66,52 @@ impl UsizeDeserializable for MemoryRegionDescriptionParams {
     }
 }
 
+fn validate_host_pointer(ptr_u64: u64, alignment: usize) -> *const u8 {
+    assert!(ptr_u64 != 0);
```
0xVolosnikov, line 75: Can you explain why?

0xVolosnikov:
@vv-dev-ai do an aggressive and paranoid review of this PR. This change is critical - you can use subagents to check it from different angles, with different context and different knowledge. Also take into account comments in this PR

vv-dev-ai:
[AI-review] Reviewed commit `c4c3a21069a2a6ad8f0b13fc7aa39efc934911cb` — **COMMENT** (3 inline comments).

vv-dev-ai (commented):
[AI-review] PR #418 — **COMMENT**

### Findings

**1. `read_host_struct<T>` lacks `T: Copy` bound** (Medium, defense-in-depth)

The function is `pub(crate)` and accepts arbitrary `T`. All current callers pass `Copy` types (`ModExpAdviceParams64`, `FieldOpsHint64`), so this is sound today. But the API permits future callers to pass types with drop glue, which would create double-free UB via `ptr::read()`. Adding `T: Copy` is a zero-cost compile-time guard. This was flagged in the prior review and remains open.

**2. `len * 2` can silently wrap to 0** (Low, oracle_provider/src/lib.rs:350)

If an oracle response ever has `len > usize::MAX / 2` elements, `len * 2` wraps to a small value and `u32::try_from` succeeds — recording incorrect witness data without panicking. Unreachable in practice (gas budget caps response sizes), but `len.checked_mul(2).expect(...)` would make the intent explicit. This was also flagged in the prior review.

**3. `arg.a_len * 4` wrapping multiplication** (Low, callable_oracles/src/arithmetic/mod.rs:161)

Same pattern: `u64` multiplication without overflow check. In practice `a_len` comes from `BigintRepr.digits` which is bounded by gas-metered EVM input parsing, but `arg.a_len.checked_mul(4).unwrap()` would be strictly correct.

### Verified Correct (via subagent deep analysis)

- **Oracle query ordering preserved**: `DISCONNECT_ORACLE_QUERY_ID` correctly moved from `proof_running_system` (after `run_prepared`) into shared `post_op` functions in `basic_bootloader`. Both native and RISC-V modes now issue DISCONNECT at the same point.
- **Witness recording format**: `ReadWitnessSource` IOOracle impl produces a u32 stream byte-for-byte identical to the RISC-V simulator path — length prefix `len * 2`, LE word splitting confirmed by tests.
- **Arithmetic query packing**: Header `(q_len | r_len << 32)` round-trips correctly on both 32-bit and 64-bit consumers. Odd-count assertion guards against malformed oracle responses.
- **`USE_ADVICE` const generic**: `ForwardSystemTypes<O, true>` resolves to `NoStdSystemFunctions<true>`, matching `proof_running_system`. Both modes use oracle delegation for ecrecover/modexp.
- **`BasicBootloaderProvingExecutionConfig`**: Native prover input run uses the same execution config as RISC-V proving. Same `BlockDataKeeper`, `PostTxLoopOp`, and storage model parameters.
- **`bigint_op_delegation_with_carry_bit` cfg removal**: Correct and necessary — the native prover input path now exercises this code outside of tests.
- **Cross-mode equivalence testing**: `assert_block_outputs_match` thoroughly validates header, storage writes, account diffs, tx results, gas/native usage. `assert_eq!(prover_input_forward, proof_input)` verifies witness-level equivalence.
- **`compile_error!` guards**: Prevent compilation on unsupported (non-64-bit-LE) targets.
- **`write_bigint` MaybeUninit safety**: All digits fully initialized before `set_num_digits` — verified for both production and naive-advisor paths.

### Suggested Tests

- **Edge case: modexp with zero-length quotient on 64-bit path** — verify `q_len / 2 = 0` produces correct empty `BigintRepr` when dividend < modulus.
- **Property test: witness equivalence for transactions exercising all three oracle types** (arithmetic, blob KZG, field ops) — the current test only covers simple transactions. A test with a contract calling the modexp precompile would exercise the full native oracle path end-to-end.

callable_oracles/src/lib.rs
```diff
@@ -60,12 +66,52 @@ impl UsizeDeserializable for MemoryRegionDescriptionParams {
...
+    let len_bytes = checked_byte_len(len_words_u8, core::mem::size_of::<u8>());
+
+    // Safety: the caller supplied a non-null pointer, and `len_bytes` was
+    // derived from a checked byte-length computation.
+    let bytes = unsafe { core::slice::from_raw_parts(ptr, len_bytes) };
+    bytes.to_vec()
+}
+
+#[inline(always)]
+pub(crate) fn read_host_struct<T>(ptr_u64: u64) -> T {
```
vv-dev-ai: [AI-review] **Medium:** `read_host_struct<T>` has no trait bounds on `T`. Since `ptr::read()` does a bitwise copy, passing a type with drop glue would create a second owner — double-free UB. All current callers use `Copy` types, so this is sound today, but the API permits misuse.

Fix: `pub(crate) fn read_host_struct<T: Copy>(ptr_u64: u64) -> T`

Also, the safety comment should document the full precondition: the pointed-to memory must be initialized and valid for `size_of::<T>()` bytes within the current process's address space.

callable_oracles/src/arithmetic/mod.rs
```diff
@@ -58,37 +106,265 @@ impl<M: MemorySource> OracleQueryProcessor<M> for ArithmeticQuery<M> {
...
+        let arg: ModExpAdviceParams64 = read_host_struct(arg_ptr as u64);
+
+        assert!(arg.a_ptr > 0);
+        assert!(arg.a_len > 0);
+        assert_eq!(arg.b_ptr, 0);
+        assert_eq!(arg.b_len, 0);
+        assert!(arg.modulus_ptr > 0);
+        assert!(arg.modulus_len > 0);
+
+        let a_len_u64_words = arg.a_len * 4;
```
vv-dev-ai: [AI-review] **Low:** `arg.a_len * 4` and `arg.modulus_len * 4` (line 162) use wrapping `u64` multiplication. While unreachable in practice (gas metering bounds `a_len`), consider `arg.a_len.checked_mul(4).expect("a_len overflow")` for defensive correctness.

oracle_provider/src/lib.rs
vv-dev-ai: [AI-review] **Low:** `len * 2` can silently wrap to 0 if `len > usize::MAX / 2`, causing `u32::try_from` to succeed with an incorrect value. Unreachable under current gas budgets, but `len.checked_mul(2).expect("response length overflow")` makes the intent explicit and prevents silent data corruption.

callable_oracles/src/lib.rs
antoniolocascio: Added some doc, but the tldr is to validate non-null and alignment. In theory not needed, but just to harden the implementation

basic_bootloader/src/bootloader/block_flow/ethereum/post_tx_op_proving.rs
antoniolocascio: Good point, done

vv-dev-ai:
[AI-review] Reviewed commit `088e1424017a144a52b90b5069c3991dba9d7b45` — **APPROVE** (0 inline comments).

vv-dev-ai (approved):
[AI-review] PR #418 — **APPROVE**

Reply-review: re-evaluated the 3 open issues from the prior review against the latest commit.

- ✅ **RESOLVED** `readhoststructt-has-no-trait-bounds`: `read_host_struct<T: Copy>` bound added. Safety comment now documents all preconditions (non-null, alignment, initialization, liveness, no concurrent mutation).
- ✅ **RESOLVED** `len-times-2-overflow`: `len * 2` replaced with `len.checked_mul(2).expect("response length overflow")` in `oracle_provider/src/lib.rs:350`.
- ✅ **RESOLVED** `a-len-times-4-overflow`: Both `a_len` and `modulus_len` multiplications now use `.checked_mul(4).expect(...)` in `callable_oracles/src/arithmetic/mod.rs:161-165`.

callable_oracles/src/arithmetic/mod.rs
antoniolocascio: Done

oracle_provider/src/lib.rs
antoniolocascio: Done

AntonD3 (approved):

PR #430, opened 2025-12-02
feat: minimal set of features to support gateway

AntonD3:
## What ❔

2 features to support gateway:
- read multichain root from `MesageRoot` contract and include it in the full logs root
- support free(0 gas price) l1 -> l2 transactions

## Is this a breaking change?
- [x] Yes
- [ ] No

It's backward compatible for forward run(taking into account that there are no 0 gas price l1 -> l2 transactions).
But full logs root changed(changes in the sequencer required)

## Checklist


antoniolocascio (approved):
Code LGTM, but it would be nice to add tests in instances for:
* L1 (upgrade) txs using 0 fee
* multichain_root is what we expect in the PI (for this, we probably want the native prover input run)

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/mod.rs
```diff
@@ -162,3 +162,134 @@ pub fn read_settlement_layer_chain_id<
...
+        SF,
+        N,
+        O,
+        FlatTreeWithAccountsUnderHashesStorageModel<A, R, P, SF, N, PROOF_ENV>,
+        PROOF_ENV,
+    >,
+) -> Bytes32 {
+    use zk_ee::system::IOSubsystem;
+
+    const MESSAGE_ROOT_ADDRESS: B160 = B160::from_limbs([0x10005, 0, 0]);
```
0xVolosnikov: Should be moved to constants / addresses I believe

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -386,17 +386,23 @@ where
...
-                .unwrap_or_else(|| {
-                    system_log!(
-                        system,
-                        "Native per gas calculation for L1 tx simulation overflows, using saturated arithmetic instead"
-                    );
-                    u64::MAX
-                })
+        if gas_price.is_zero() {
+            if Config::SIMULATION {
+                // TODO: do we even need a separate flow for L1 tx simulation?
```
0xVolosnikov: @antoniolocascio PTAL, it's not clear why we handle simulation case separately here - I'm not sure that we even use it somehow

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/mod.rs
```diff
@@ -162,3 +162,133 @@ pub fn read_settlement_layer_chain_id<
         .expect("must read SystemContext SL chain id");
     U256::from_be_bytes(chain_id.as_u8_array())
 }
+
+///
+/// Reads multichain root from the L2MessageRoot(0x10005) contract.
```
0xVolosnikov, line 167: Would be nice to expand this comment a bit: why do we need it and how it's used

0xVolosnikov (changes requested):
> would be nice to add tests in instances for: ... L1 (upgrade) txs using 0 fee ...

Still not added, we should cover this case here @AntonD3

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/mod.rs
```diff
@@ -162,3 +162,133 @@ pub fn read_settlement_layer_chain_id<
...
+            ExecutionEnvironmentType::NoEE,
+            &mut inf_resources,
+            &MESSAGE_ROOT_ADDRESS,
+            &Bytes32::from_array(SHARED_TREE_HEIGHT_STORAGE_SLOT),
+        )
+        .expect("must read MessageRoot shared tree height");
+
+    let root_slot = calculate_multichain_root_slot(tree_height);
+
+    io.storage_read::<false>(
```
0xVolosnikov, line 212: Should we do it even if a chain isn't a SL? Maybe this value will be zero in this case or smth?

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
antoniolocascio: This I'm not sure about. Can L1 txs even reach gas estimation? I guess not, so we can remove the special case

AntonD3: Do you want to do it in this PR?

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/mod.rs
AntonD3: This value indeed will be zero in such a case, but I didn't come up with an easy way to trustlessly distinguish in runtime if the state transition being executed is for SL or not.

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
0xVolosnikov: I think it makes sense since we touch this part anyway. Otherwise this logic is complicated

0xVolosnikov (approved):

PR #452, opened 2025-12-11
feat: add support for updating SL chain id

antoniolocascio:
## What ❔

This PR adds support for setting the settlement layer chain id via a service transaction.
Depends on https://github.com/matter-labs/era-contracts/pull/1926

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


basic_system/src/system_implementation/system/public_input.rs
```diff
@@ -159,6 +162,8 @@ pub struct BatchOutput {
     pub upgrade_tx_hash: Bytes32,
     /// Linear keccak256 hash of interop roots
     pub interop_roots_rolling_hash: Bytes32,
+    /// New settlement layer chain id, if updated. 0 otherwise.
```
0xVolosnikov: Should we explicitly forbid setting it to 0?

antoniolocascio: No longer applicable in the new version

PR #478, opened 2026-01-07
feat: Add Stage 1 support

vladbochok:
## What ❔

- Added `number_of_layer_2_txs` to the batch output/commitment 
    - It is needed for the Priority Mode, if it is activated smart contract will only accept batches with no L2 txs processed in the batch
- Changed the way L1 -> L2 transactions spends base tokens. Now they can spend base tokens from L2 balance, and not only the amount deposited from L1 in the same tx.
	- Previously, an L1 → L2 transaction minted `msg.value` before the call and minted the remaining deposited amount to the refund recipient after the call. The transaction then used that minted `msg.value` during execution.
	- Now, an L1 → L2 transaction mints the deposited amount minus the required fee. The `msg.value` can be set arbitrarily by the requester on L1. If it exceeds the account’s balance, the transaction fails and the deposited amount is credited to the refund recipient.

## Why ❔

- These are two features required for the Stage 1. 

## Is this a breaking change?
- [x] Yes
- [ ] No

## Other PRs

- https://github.com/matter-labs/era-contracts/pull/1943
- https://github.com/zksync-association/zk-governance/pull/32

## Checklist

### DIFF
--- a/basic_bootloader/src/bootloader/block_flow/zk/batch_data.rs
+++ b/basic_bootloader/src/bootloader/block_flow/zk/batch_data.rs
@@ -103,7 +103,7 @@ impl<A: alloc::alloc::Allocator, O: IOOracle> ZKBatchDataKeeper<A, O> {
     }
 
     ///
-    /// Returns if the batch has had an upgrade tx in the first block
+    /// Returns if the batch has had an upgrade tx
     ///
     pub fn has_upgrade_tx(&self) -> bool {
         self.upgrade_tx_hash
--- a/basic_bootloader/src/bootloader/block_flow/zk/batch_data.rs
+++ b/basic_bootloader/src/bootloader/block_flow/zk/batch_data.rs
@@ -106,7 +106,8 @@ impl<A: alloc::alloc::Allocator, O: IOOracle> ZKBatchDataKeeper<A, O> {
     /// Returns if the batch has had an upgrade tx
     ///
     pub fn has_upgrade_tx(&self) -> bool {
-        self.upgrade_tx_hash
+        !self
+            .upgrade_tx_hash
             .is_none_or(|hash| hash == Bytes32::ZERO)
     }
 
@@ -297,5 +298,45 @@ impl<A: alloc::alloc::Allocator, O: IOOracle> TxHashesAccumulator for ZKBatchDat
     // used to write l1 txs in tx loop
     fn add_tx_hash(&mut self, tx_hash: &Bytes32) {
         self.enforced_txs_accumulator.add_tx_hash(tx_hash);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use alloc::alloc::Global;
+    use zk_ee::oracle::usize_serialization::{UsizeDeserializable, UsizeSerializable};
+    use zk_ee::system::errors::internal::InternalError;
+
+    struct DummyOracle;
+
+    impl IOOracle for DummyOracle {
+        type RawIterator<'a> = core::iter::Empty<usize>;
+
+        fn raw_query<'a, I: UsizeSerializable + UsizeDeserializable>(
+            &'a mut self,
+            _query_type: u32,
+            _input: &I,
+        ) -> Result<Self::RawIterator<'a>, InternalError> {
+            Ok(core::iter::empty())
+        }
+    }
+
+    #[test]
+    fn has_upgrade_tx_is_false_for_none_and_zero_hash() {
+        let mut keeper = ZKBatchDataKeeper::<Global, DummyOracle>::new();
+
+        assert!(!keeper.has_upgrade_tx());
+
+        keeper.upgrade_tx_hash = Some(Bytes32::ZERO);
+        assert!(!keeper.has_upgrade_tx());
+    }
+
+    #[test]
+    fn has_upgrade_tx_is_true_for_non_zero_hash() {
+        let mut keeper = ZKBatchDataKeeper::<Global, DummyOracle>::new();
+        keeper.upgrade_tx_hash = Some(Bytes32::from_byte_fill(1));
+
+        assert!(keeper.has_upgrade_tx());
     }
 }
--- a/basic_bootloader/src/bootloader/block_flow/zk/batch_data.rs
+++ b/basic_bootloader/src/bootloader/block_flow/zk/batch_data.rs
@@ -26,6 +26,8 @@ pub struct ZKBatchDataKeeper<A: alloc::alloc::Allocator, O: IOOracle> {
     pub da_commitment_generator: Option<alloc::boxed::Box<dyn DACommitmentGenerator<O>, A>>,
     pub logs_storage: ArrayVec<Bytes32, 16384>,
     enforced_txs_accumulator: TransactionsRollingKeccakHasher,
+    // Includes all transactions
+    pub tx_count: U256,
     upgrade_tx_hash: Option<Bytes32>,
     interop_roots_rolling_hash: Bytes32,
     settlement_layer_chain_id: Option<U256>,
@@ -45,6 +47,7 @@ impl<A: alloc::alloc::Allocator, O: IOOracle> ZKBatchDataKeeper<A, O> {
             logs_storage: ArrayVec::new(),
             // keccak256([])
             enforced_txs_accumulator: TransactionsRollingKeccakHasher::empty(),
+            tx_count: U256::ZERO,
             upgrade_tx_hash: None,
             interop_roots_rolling_hash: Bytes32::ZERO,
             settlement_layer_chain_id: None,
@@ -64,6 +67,7 @@ impl<A: alloc::alloc::Allocator, O: IOOracle> ZKBatchDataKeeper<A, O> {
         upgrade_tx_hash: Bytes32,
         interop_roots: impl Iterator<Item = &'a InteropRoot>,
         settlement_layer_chain_id: U256,
+        number_of_txs_in_block: u32,
     ) {
         if self.is_first_block {
             self.initial_state_commitment = Some(state_commitment_before);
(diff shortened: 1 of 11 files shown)


basic_system/src/system_implementation/system/public_input.rs
```diff
@@ -222,6 +225,7 @@ pub struct BatchPublicInputBuilder<A: alloc::alloc::Allocator, O: IOOracle> {
     pub da_commitment_generator: Option<alloc::boxed::Box<dyn DACommitmentGenerator<O>, A>>,
     pub logs_storage: ArrayVec<Bytes32, 16384>,
     pub number_of_layer_1_txs: U256,
+    pub number_of_layer_2_txs: U256,
```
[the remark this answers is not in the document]
antoniolocascio: Done it

basic_system/src/system_implementation/system/io_subsystem.rs
```diff
@@ -721,6 +721,12 @@ impl<
         full_root_hasher.update([0u8; 32]); // aggregated root 0 for now
         let full_l2_to_l1_logs_root = full_root_hasher.finalize();
         let l1_txs_commitment = self.logs_storage.l1_txs_commitment();
+        // Number of L2 transactions can be calculated as:
+        // Total txs - (l1 txs + service txs)
+        let mut number_of_layer_2_txs = self.tx_number - l1_txs_commitment.0;
```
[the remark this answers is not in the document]
antoniolocascio: Note that for now we consider service txs as L2 transactions, for the reason I commented above

0xVolosnikov:
`test_treasury_based_token_distribution_regression` failed for some reason

basic_bootloader/src/bootloader/block_flow/zk/batch_data.rs
```diff
@@ -89,18 +93,29 @@ impl<A: alloc::alloc::Allocator, O: IOOracle> ZKBatchDataKeeper<A, O> {
...
+
         self.interop_roots_rolling_hash = calculate_interop_roots_rolling_hash(
             self.interop_roots_rolling_hash,
             interop_roots,
             &mut crypto::sha3::Keccak256::new(),
         );
     }
 
+    ///
+    /// Returns if the batch has had an upgrade tx in the first block
```
0xVolosnikov: > in the first block

It is not checked/enforced here, so I'd prefer to remove this part

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/post_tx_op_proving_singleblock_batch.rs
```diff
@@ -101,12 +101,20 @@ where
 
         let (priority_operations_hash, number_of_layer_1_txs) =
             block_data.enforced_transaction_hashes_accumulator.finish();
+        // Number of L2 transactions can be calculated as:
```
0xVolosnikov, line 104: We could just implement separate counters for each tx type. Was it considered?

0xVolosnikov: With the current implementation we will have to remember that we calculate `number_of_layer_2_txs` somewhere (in several places even) based on an assumption that we have only l1, l2 and upgrade txs. In future this part of the logic can change and details like this one can create some additional headache

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -299,15 +296,15 @@ where
...
-                .checked_add(pay_to_operator)
-                .ok_or(internal_error!("v+pto"))?;
-            total_deposited
-                .checked_sub(value_plus_fee)
+            let tx_internal_cost = gas_price
+                .checked_mul(U256::from(transaction.gas_limit.read()))
+                .ok_or(internal_error!("gp*gl"))?;
+            tx_internal_cost
+                .checked_sub(pay_to_operator)
                 .ok_or(internal_error!("td-vpf"))
```
0xVolosnikov: nit: error message is outdated

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -299,15 +296,15 @@ where
...
+            // If the transaction succeeds, then it is assumed that the
+            // mint to `from` address was transferred correctly too.
             // However, the remaining value deposited will be given to
             // the refund recipient.
-            let value_plus_fee = value
-                .checked_add(pay_to_operator)
-                .ok_or(internal_error!("v+pto"))?;
-            total_deposited
-                .checked_sub(value_plus_fee)
+            let tx_internal_cost = gas_price
```
0xVolosnikov: Imo it would be nice to add some comments here to explain what is happening. Because names `tx_internal_cost` and `pay_to_operator` are not very descriptive

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -407,13 +404,21 @@ where
     // Start a frame, to revert minting of value if execution fails
     let rollback_handle = system.start_global_frame()?;
 
-    // First we transfer value from treasury
-    if value > U256::ZERO {
+    let tx_internal_cost = gas_price
```
0xVolosnikov: Again, would be nice to explain this part

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/post_tx_op_proving_singleblock_batch.rs
0xVolosnikov: Actually, as I see, we already had this confusion with service transactions

basic_system/src/system_implementation/system/io_subsystem.rs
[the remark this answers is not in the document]
0xVolosnikov: This means that interop is essentially blocked/disabled in priority mode. In theory, some funds can be blocked because of this - imagine if some contracts will rely purely on the interop logic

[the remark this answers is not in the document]
0xVolosnikov: I think that this may be actually a problem and we should enforce inclusion of interop roots on the protocol level. Alternatively, we may allow an operator to submit interop roots using normal L1->L2 tx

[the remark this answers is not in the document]
0xVolosnikov: Also, we may consider moving L1 transactions to the service blocks. Let's briefly touch this topic on Monday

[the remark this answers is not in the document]
antoniolocascio: Let's discuss it, sure, but I still think that priority mode should be reserved for priority txs.

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/post_tx_op_proving_singleblock_batch.rs
antoniolocascio: > Actually, as I see, we already had this confusion with service transactions

Well, I consider service txs to be L2 txs, so that's consistent with the code here.
We can introduce a new counter for L2 txs, but I'm not convinced that it's worth it. L1 txs (priority and upgrade) don't change/get extended frequently

basic_bootloader/src/bootloader/block_flow/zk/batch_data.rs
antoniolocascio: Done

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
antoniolocascio: Hope it's clearer now

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
antoniolocascio: I think it's clearer now

basic_system/src/system_implementation/system/io_subsystem.rs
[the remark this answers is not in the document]
0xVolosnikov: So after the discussion it looks like we can keep it as is for now

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/post_tx_op_proving_singleblock_batch.rs
0xVolosnikov: In my opinion an explicit counter would a better and cleaner solution, because this way we won't rely on any assumptions about our transactions. E.g. now we assume that we can have only 1 upgrade tx per block. I agree, it doesn't change often, but in general we should try to decouple different parts of our logic, if we can - this way our code will be easier to understand and reason about. 

Anyway, this exact case is not very significant, so we can keep it as is if you prefer it

antoniolocascio: > because this way we won't rely on any assumptions about our transactions.  E.g. now we assume that we can have only 1 upgrade tx per block. 

But this is the current spec. Adding a new counter for upgrade txs will only add more validation (e.g. checking that it never goes over 1 for a single block). But I don't feel too strongly about it

0xVolosnikov (approved):

basic_bootloader/src/bootloader/block_flow/zk/batch_data.rs
```diff
@@ -89,18 +93,29 @@ impl<A: alloc::alloc::Allocator, O: IOOracle> ZKBatchDataKeeper<A, O> {
...
         );
     }
 
+    ///
+    /// Returns if the batch has had an upgrade tx
+    ///
+    pub fn has_upgrade_tx(&self) -> bool {
+        self.upgrade_tx_hash
+            .is_none_or(|hash| hash == Bytes32::ZERO)
+    }
```
vladbochok, line 111:
```suggestion
    pub fn has_upgrade_tx(&self) -> bool {
        !self.upgrade_tx_hash
            .is_none_or(|hash| hash == Bytes32::ZERO)
    }
```

PR #489, opened 2026-01-14
feat: introduce callable oracle for ecrecover

antoniolocascio:
## What ❔

Reimplement the hints for ecrecover from ethproofs.
Instead of having a new full implementation of ecrecover, this PR introduces "hooks" for 3 secp256k1 field operations. These hooks have two implmentations: default, where the implementation is straightforward (same implementation as before this PR) and oracle-based, where we use the new callable oracles.
This way, most of the logic for ecrecover is reused.

Given that this is a critical part of the system, the PR includes:
* PBT for "good" case (showing equivalence when using the right oracle)
* Tests for the "bad" case (showing panics if the oracle lies)
* Modifies the fuzz target for ecrecover to compare the forward and oracle runs. I've ran the fuzzer for over 1h without finding issues.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/basic_system/src/system_functions/ecrecover.rs
+++ b/basic_system/src/system_functions/ecrecover.rs
@@ -12,9 +12,11 @@ use zk_ee::system::{Computational, SystemFunctionExt};
 ///
 /// ecrecover system function implementation.
 ///
-pub struct EcRecoverImpl;
+pub struct EcRecoverImpl<const USE_ADVICE: bool = { cfg!(target_arch = "riscv32") }>;
 
-impl<R: Resources> SystemFunctionExt<R, Secp256k1ECRecoverErrors> for EcRecoverImpl {
+impl<R: Resources, const USE_ADVICE: bool> SystemFunctionExt<R, Secp256k1ECRecoverErrors>
+    for EcRecoverImpl<USE_ADVICE>
+{
     /// If the input size is less than expected - it will be padded with zeroes.
     /// If the input size is greater - redundant bytes will be ignored.
     /// If the input is invalid(v != 27|28 or failed to recover signer) returns `Ok(0)`.
@@ -31,8 +33,14 @@ impl<R: Resources> SystemFunctionExt<R, Secp256k1ECRecoverErrors> for EcRecoverI
         Ok(cycle_marker::wrap_with_resources!(
             "ecrecover",
             resources,
-            // TODO: reconsider if we actually want to use the oracle based version here
-            { ecrecover_as_system_function_inner(input, output, resources, Some(oracle)) }
+            {
+                ecrecover_as_system_function_inner::<_, _, _, _, USE_ADVICE>(
+                    input,
+                    output,
+                    resources,
+                    Some(oracle),
+                )
+            }
         )?)
     }
 }
@@ -43,6 +51,7 @@ fn ecrecover_as_system_function_inner<
     S: ?Sized + MinimalByteAddressableSlice,
     D: ?Sized + TryExtend<u8>,
     R: Resources,
+    const USE_ADVICE: bool,
 >(
     src: &S,
     dst: &mut D,
@@ -77,12 +86,7 @@ fn ecrecover_as_system_function_inner<
             return Ok(());
         }
 
-        // on forward run we do not support oracle-based hooks
-        let oracle = if cfg!(target_arch = "riscv32") {
-            oracle
-        } else {
-            None
-        };
+        let oracle = if USE_ADVICE { oracle } else { None };
 
         let Ok(pk_bytes) = ecrecover_inner(digest, r, s, rec_id, oracle) else {
             return Ok(());
@@ -173,7 +177,7 @@ mod test {
 
         let mut resources = <BaseResources<DecreasingNative> as Resource>::FORMAL_INFINITE;
 
-        ecrecover_as_system_function_inner::<TestingTree<false>, _, _, _>(
+        ecrecover_as_system_function_inner::<TestingTree<false>, _, _, _, false>(
             input.as_slice(),
             &mut pubkey,
             &mut resources,
@@ -194,7 +198,7 @@ mod test {
 
         let mut resources = <BaseResources<DecreasingNative> as Resource>::FORMAL_INFINITE;
 
-        ecrecover_as_system_function_inner::<TestingTree<false>, _, _, _>(
+        ecrecover_as_system_function_inner::<TestingTree<false>, _, _, _, false>(
             input.as_slice(),
             &mut pubkey,
             &mut resources,
@@ -216,7 +220,7 @@ mod test {
 
         let mut resources = <BaseResources<DecreasingNative> as Resource>::FORMAL_INFINITE;
 
-        ecrecover_as_system_function_inner::<TestingTree<false>, _, _, _>(
+        ecrecover_as_system_function_inner::<TestingTree<false>, _, _, _, false>(
             input.as_slice(),
             &mut pubkey,
             &mut resources,
@@ -238,7 +242,7 @@ mod test {
 
         let mut resources = <BaseResources<DecreasingNative> as Resource>::FORMAL_INFINITE;
 
-        ecrecover_as_system_function_inner::<TestingTree<false>, _, _, _>(
+        ecrecover_as_system_function_inner::<TestingTree<false>, _, _, _, false>(
             input.as_slice(),
             &mut pubkey,
             &mut resources,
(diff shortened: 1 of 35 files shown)


shamatar (commented):
Looks fine for me in general. Did you get any fuzzer results already?

basic_system/src/system_functions/ecrecover.rs
```diff
@@ -1,44 +1,53 @@
...
         resources: &mut R,
+        oracle: &mut O,
+        _logger: &mut L,
         _allocator: A,
     ) -> Result<(), SubsystemError<Secp256k1ECRecoverErrors>> {
         Ok(cycle_marker::wrap_with_resources!(
             "ecrecover",
             resources,
-            { ecrecover_as_system_function_inner(input, output, resources) }
+            // TODO: reconsider if we actually want to use the oracle based version here
```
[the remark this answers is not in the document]
0xVolosnikov: Note: we should not forget about it in v0.4.0 @AntonD3

shamatar (approved):

shmatar:
Bytes32 had at least usize alignment, but we can put it at alignment 8, or compile time assert that usize is u64

0xVolosnikov:
@vv-dev-ai ptal

vv-dev-ai:
[AI-review] Review started for commit `c7783f0f057c3733d34c558b8200f0b1d5370a70`. I'll post my findings shortly.

callable_oracles/src/lib.rs
```diff
@@ -58,3 +59,13 @@ impl UsizeDeserializable for MemoryRegionDescriptionParams {
...
 }
+
+#[inline(always)]
+unsafe fn read_u64_words(ptr_u64: u64, len_words_u64: u64) -> Vec<u64> {
+    if ptr_u64 == 0 || len_words_u64 == 0 {
+        return vec![];
+    }
+    let addr = ptr_u64 as usize;
+    let len_words = len_words_u64 as usize;
+    core::slice::from_raw_parts(addr as *const u64, len_words).to_vec()
```
vv-dev-ai, line 70: [AI-review] Missing `// SAFETY:` comment on this `unsafe` function. Every unsafe block/function in this codebase requires a SAFETY comment explaining why the operation is sound.

The function dereferences a raw pointer derived from `ptr_u64` (cast from an oracle-provided address) with length `len_words_u64`. The zero-check on line 65 guards against null pointers, but there is no documented invariant guaranteeing that:
1. The address points to a valid, live allocation of at least `len_words * 8` bytes
2. The pointer is properly aligned for `u64` reads
3. The memory will not be mutated concurrently

Suggested fix: add a `// SAFETY:` comment above the `core::slice::from_raw_parts` call documenting which caller invariants make this sound (likely: "the pointer originates from a live Rust allocation within the same process, passed by the RISC-V program via the oracle query").

callable_oracles/src/field_hints/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+        query_id: u32,
+        query: Vec<usize>,
+        _memory: &M,
+    ) -> Box<dyn ExactSizeIterator<Item = usize> + 'static + Send + Sync> {
+        debug_assert!(self.supports_query_id(query_id));
+
+        let mut it = query.into_iter();
+
+        let arg_ptr = it.next().expect("A u64 should've been passed in.");
+
```
vv-dev-ai, line 124: [AI-review] Missing `// SAFETY:` comment on this `unsafe` block. The `core::ptr::read_unaligned` call reinterprets `arg_ptr` (a `usize` from the oracle query) as a `*const FieldOpsHint64` and reads from it.

This is sound only if `arg_ptr` is a valid pointer to a `FieldOpsHint64` in the current process's address space. In the `NativeFieldOpsQuery` context (host-side, x86), the pointer comes from the RISC-V program's oracle query which encodes a native pointer — so it points to live memory owned by the calling code.

Suggested fix: add a `// SAFETY:` comment explaining the trust model, e.g.:
```rust
// SAFETY: `arg_ptr` is a native pointer passed by the calling code within
// the same process. `read_unaligned` is used because `FieldOpsHint64` may
// not be naturally aligned at the source address.
```

Also: the second `unsafe` call to `read_u64_words` on line ~133 needs the same treatment.

callable_oracles/src/field_hints/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+                let t = impls::secp256k1_base_field_sqrt(n);
+                DynUsizeIterator::from_constructor(t, UsizeSerializable::iter)
+            }
+            FieldHintOp::Secp256k1BaseFieldInverse => {
+                let t = impls::secp256k1_base_field_inverse(n);
+                DynUsizeIterator::from_constructor(t, UsizeSerializable::iter)
+            }
+            FieldHintOp::Secp256k1ScalarFieldInverse => {
+                let t = impls::secp256k1_scalar_field_inverse(n);
+                DynUsizeIterator::from_constructor(t, UsizeSerializable::iter)
```
vv-dev-ai, line 87: [AI-review] The wildcard arm `_ => { panic!("Unknown field hint op {}", arg.op); }` on lines 83-85 (in `FieldOpsQuery`) and the equivalent on lines 153-155 (in `NativeFieldOpsQuery`) are unreachable dead code. `FieldHintOp::parse_u32` already returns `None` for unknown ops, and the `let Some(op) = ... else { panic!(...) }` on line 53/line 126 handles that case.

Since `FieldHintOp` is `#[non_exhaustive]`, the wildcard arm is required by the compiler, but it can never execute. Consider replacing the panic with `unreachable!("parse_u32 already validated the op")` for clarity.

basic_system/src/system_functions/field_ops.rs
```diff
@@ -0,0 +1,701 @@
...
+        }
+
+        // ============ fe_invert tests ============
+
+        #[test]
+        #[should_panic]
+        fn test_fe_invert_rejects_zero_answer() {
+            let mut oracle = create_lying_oracle(Corruption::ReturnZero);
+            let mut fe = test_field_element();
+            Secp256k1HooksWithOracle::new(&mut oracle).fe_invert_and_assign(&mut fe);
```
vv-dev-ai, line 407: [AI-review] Minor: the `.unwrap()` on `FieldElement::from_bytes(sqrt_candidate.as_u8_array_ref())` will panic with a generic "called `Option::unwrap()` on a `None` value" if the oracle returns bytes >= the secp256k1 field modulus p. The subsequent `assert!` on line 417 would catch a *wrong but valid* field element, but bytes >= p fail earlier with no diagnostic.

Since this only runs in proving mode (where panic = proof rejection, which is correct), this is not a correctness bug. But for debuggability, consider:
```rust
let fe = FieldElement::from_bytes(sqrt_candidate.as_u8_array_ref())
    .expect("oracle returned value >= field modulus");
```

Same applies to `fe_invert_and_assign` (line ~438) and `scalar_invert_and_assign` (line ~465, the `assert!(inverse < ORDER)` already covers this case there, which is good).

vv-dev-ai (changes requested):
[AI-review] PR #489 — **REQUEST_CHANGES**

### Findings

The architectural approach is sound: introducing a `Secp256k1Hooks` trait to decouple field operations (sqrt, inverse) from their implementations, then providing both a `DefaultSecp256k1Hooks` (direct computation) and a `Secp256k1HooksWithOracle` (oracle hint + algebraic verification). The `USE_ADVICE` const generic on `EcRecoverImpl` correctly defaults to `cfg!(target_arch = "riscv32")`, ensuring oracle hints are only used in proving mode while forward mode uses direct computation.

The verification math is correct:
- `fe_sqrt_and_assign`: verifies `candidate² == x` (residue) or `candidate² == -x` (non-residue)
- `fe_invert_and_assign`: verifies `x * inv == 1`
- `scalar_invert_and_assign`: verifies `x * inv == 1` after range check against `ORDER`

Resource accounting is unchanged — costs are charged before computation in `ecrecover_as_system_function_inner`, independent of the oracle path.

Cross-mode consistency is maintained: forward mode uses `DefaultSecp256k1Hooks` (oracle not consulted even though passed), proving mode uses `Secp256k1HooksWithOracle`. PBT tests verify equivalence.

Test coverage is excellent — PBT for oracle/default equivalence, malicious oracle rejection tests, zero-input regression tests, and the fuzzer now compares forward vs oracle paths.

Two `unsafe` blocks are missing required `// SAFETY:` comments — see inline comments.

Adversarial analysis: no additional failure paths found beyond those noted in inline comments. The `assert!` calls in the oracle verification hooks are intentional proof-failure triggers (panic in proving mode = proof rejection), which is the correct behavior for ZK oracle hint verification.

### Suggested Tests

- **Differential forward/proving test via `zksync_os_runner`**: execute a transaction calling ecrecover with a known-good signature through both `forward_system::run_block` and the RISC-V runner — verify identical state roots. This would exercise the full `USE_ADVICE=true` path end-to-end.
- **Edge case PBT for `fe_sqrt_and_assign` with values near the field modulus** (p-1, p-2): these are the most likely to expose normalization bugs in the sqrt verification.

callable_oracles/src/lib.rs
antoniolocascio: Done in future PR

callable_oracles/src/field_hints/mod.rs
antoniolocascio: Fixed in next PR

PR #510, opened 2026-02-04
fix!: handle validation errors for l1 transactions

antoniolocascio:
## What ❔
Handle "validation errors" that should be already enforced by L1 contracts, as a fallback

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


0xVolosnikov (approved):
I really like this implementation. Did not check that new tests properly catch corresponding panics though.

PR #511, opened 2026-02-05
feat: add salt field to service txs

antoniolocascio:
## What ❔
Add a salt field to service txs for the server to identify transactions (avoiding "equal" service txs having the same hash).

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


0xVolosnikov (approved):
LGTM. I think it's ok to not enforce salt uniqueness since this is needed only for internal usage

PR #514, opened 2026-02-10
fix: add reduction to u512 montgomerry mul

yoaveshel:
## What ❔
add a final reduction step to montgomerry multiplication to make sure we don't get non-canonical representation.
(the scratch space test infrastructure would sometimes be unaligned, so this PR fixes that as well)

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/crypto/src/bigint_delegation/u512.rs
+++ b/crypto/src/bigint_delegation/u512.rs
@@ -3,10 +3,7 @@ use super::{
     u256::{self, U256},
     DelegatedModParams, DelegatedMontParams,
 };
-use crate::{
-    ark_ff_delegation::{BigInt, BigInteger},
-    bigint_delegation::u256::sub_assign,
-};
+use crate::ark_ff_delegation::{BigInt, BigInteger};
 
 pub(super) type U512 = BigInt<8>;
 
--- a/crypto/src/bigint_delegation/u512.rs
+++ b/crypto/src/bigint_delegation/u512.rs
@@ -356,11 +356,10 @@ pub unsafe fn mul_assign_montgomery<T: DelegatedMontParams<8>>(a: &mut U512, b: 
         // // Final reduction: result is in [0, 2*modulus). Subtract modulus if >= modulus.
         // // For BLS12-381, p ≈ 2^381 and R = 2^512, so the probability of a
         // // non-canonical result is p²/R / p ≈ 2^{-131}
-        // // so we check to avoid unnecessary delegation calls
-        // if carry || !u256::lt(a1, as_high(T::modulus())) {
-        //     sub_mod_with_carry::<T>(a, carry);
-        // }
-        sub_mod_with_carry::<T>(a, carry);
+        // so we check to avoid unnecessary delegation calls
+        if carry || !u256::lt(a1, as_high(T::modulus())) {
+            sub_mod_with_carry::<T>(a, carry);
+        }
 
         debug_assert!(a.0[6..8].iter().all(|&x| x == 0));
     })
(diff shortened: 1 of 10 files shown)


crypto/src/bigint_delegation/u512.rs
```diff
@@ -110,6 +110,20 @@ fn copy(dst: &mut U512, src: &U512) {
...
+        if a.0[i] > m.0[i] {
+            return true;
+        }
+        if a.0[i] < m.0[i] {
+            return false;
+        }
+    }
+    true // equal
+}
+
```
antoniolocascio: Is this ever used?

crypto/src/bigint_delegation/u512.rs
```diff
@@ -351,7 +365,15 @@ pub unsafe fn mul_assign_montgomery<T: DelegatedMontParams<8>>(a: &mut U512, b:
...
         let carry2 = new_carry_2;
 
-        u256::add_assign(a1, carry2);
+        let carry = u256::add_assign(a1, carry2);
+
+        // Final reduction: result is in [0, 2*modulus). Subtract modulus if >= modulus.
+        // For BLS12-381, p ≈ 2^381 and R = 2^512, so the probability of a
+        // non-canonical result is p²/R / p ≈ 2^{-131}
+        // so we check to avoid unnecessary delegation calls
+        if carry || !u256::lt(a1, as_high(T::modulus())) {
```
antoniolocascio: So, here `!u256::lt(a1, as_high(T::modulus()))`
we allow the case when the high part is > high(p) but the low is < low(p)? Did you mean to use `is_ge_modulus` here?

crypto/src/bigint_delegation/u512.rs
yoaveshel: removed

crypto/src/bigint_delegation/u512.rs
yoaveshel: a1 is the high limb, so we're checking if `high(a) < high(p)` and if not we reduce

antoniolocascio: Yeah, I see, this is just to check I understand. That means that we might end up reducing even if a < p, no? Ofc this is fine

yoaveshel: Yes it might happen. But to check that a < p properly is 2-3 delegation calls, so we might as well skip the check and just call sub_mod_with_carry

crypto/src/bigint_delegation/u512.rs
```diff
@@ -351,7 +351,15 @@ pub unsafe fn mul_assign_montgomery<T: DelegatedMontParams<8>>(a: &mut U512, b:
 
         let carry2 = new_carry_2;
 
-        u256::add_assign(a1, carry2);
+        let carry = u256::add_assign(a1, carry2);
```
shamatar, line 359: For BLS12-381 carry here is impossible. If we consider implementation generic - then we keep carry, but remove comment. Alternative is to add number of bits parameter into Montgomery params trait, and if it's not 256/512 bit modulus exactly, then we do not need to consider carry

crypto/src/bigint_delegation/u512.rs
```diff
@@ -351,7 +351,15 @@ pub unsafe fn mul_assign_montgomery<T: DelegatedMontParams<8>>(a: &mut U512, b:
...
+
+        // // Final reduction: result is in [0, 2*modulus). Subtract modulus if >= modulus.
+        // // For BLS12-381, p ≈ 2^381 and R = 2^512, so the probability of a
+        // // non-canonical result is p²/R / p ≈ 2^{-131}
+        // so we check to avoid unnecessary delegation calls
+        if carry || !u256::lt(a1, as_high(T::modulus())) {
```
shamatar: If we ignore carry (comment above), then we can sub modulus with borrow, and conditionally select based on borrow (or add modulus back). Comparison is subtraction anyway

shamatar: Not to say that it's not enough to compare high part only

shamatar (changes requested):

shamatar (approved):

PR #515, opened 2026-02-10
fix!: fixes from internal audit

antoniolocascio:
## What ❔
This PR contains fixes for the following issues:
* Actually bump tx counter in accumulator
* Disable blob base fee validation on simulation
* Add more validation in event hooks
* Fix error propagation in block hashes deserialization
* Fix create/create2 tracing
* Fix FlatStorageBacking initialization 

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
+++ b/basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
@@ -333,7 +333,7 @@ where
             system
         )?;
 
-        if &block_base_fee_per_blob_gas > tx_max_fee_per_blob_gas {
+        if &block_base_fee_per_blob_gas > tx_max_fee_per_blob_gas && !Config::SIMULATION {
             return Err(TxError::Validation(
                 InvalidTransaction::BlobBaseFeeGreaterThanMaxFeePerBlobGas,
             ));
--- a/system_hooks/src/event_hooks/system_context.rs
+++ b/system_hooks/src/event_hooks/system_context.rs
@@ -24,6 +24,9 @@ pub fn system_context_event_hook<S: EthereumLikeTypes>(
 ) -> Result<(), SystemError>
 where
 {
+    if topics.is_empty() {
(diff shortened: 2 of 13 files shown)


basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -333,7 +333,7 @@ where
             system
         )?;
 
-        if &block_base_fee_per_blob_gas > tx_max_fee_per_blob_gas {
+        if &block_base_fee_per_blob_gas > tx_max_fee_per_blob_gas && !Config::SIMULATION {
```
0xVolosnikov, line 336: We don't have a test for this case. Generally I'm not sure if we test blobs in the simulation mode at all

system_hooks/src/event_hooks/system_context.rs
```diff
@@ -47,6 +50,12 @@ where
             internal_error!("New SL chain id reporter event hook received bad data").into(),
         );
     }
+    // Same if there's a mismatch in expected topics
+    if topics.len() != 2 {
```
0xVolosnikov, line 54: Regression test would be nice here

tests/instances/system_hooks/src/lib.rs
```diff
@@ -810,3 +812,53 @@ fn test_mint_base_token_hook() {
         "Minted amount should match the requested mint amount"
     );
 }
+
+#[test]
+fn test_event_hooks_empty_topics() {
```
0xVolosnikov, line 817: Would be nice to add a separate test that checks that we actually have expected event hooks on these addresses

0xVolosnikov (approved):

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
antoniolocascio: Added one

PR #516, opened 2026-02-10
test: REVM runner integration

0xVolosnikov:
## What ❔

- Added a new workspace crate `tests/revm_runner` (`zksync_os_revm_runner`) to replay blocks in REVM and compare resulting state diffs with ZKsync OS output.
- Integrated REVM consistency checking into the rig `TestingFramework` flow
- Added `RunConfig.check_revm_consistency` and env gating via `ZKSYNC_REVM_CONSISTENCY_CHECK=true`, plus per-test opt-out (`without_revm_consistency_check` / `disable_revm_consistency_check`).
- Added a CI job (`test_revm_consistency`) to run tests with REVM consistency checks enabled.
- Updated test infra/dependencies required for integration (workspace member registration, rig dependency wiring, Alloy pinning updates in tester/fuzzer wrappers, license-deny exception for `aurora-engine-modexp`, interop test crate rename).

## Why ❔

- Add automated cross-validation between ZKsync OS and REVM to catch EVM-compatibility/state-transition regressions early.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


0xVolosnikov:
Requires https://github.com/matter-labs/zksync-os-revm/pull/12

Can be merged now, we will update the dependency after zksync-os-revm is updated

.github/workflows/ci.yml
```diff
@@ -44,6 +44,33 @@ jobs:
...
+                  rustup component add rust-src
+            - name: Compile for RISC-V
+              working-directory: ./zksync_os
+              run: |
+                  rustup target add riscv32i-unknown-none-elf
+                  cargo install cargo-binutils
+                  rustup component add llvm-tools-preview
+                  rustup component add rust-src
+                  ./dump_bin.sh --type for-tests
+                  ./dump_bin.sh --type multiblock-batch
```
antoniolocascio, line 69: Maybe we can skip this step?

antoniolocascio (approved):

antoniolocascio:
Review from Claude/Codex:

### 1. `parse_explicit_bool` narrows accepted env var values
**Consensus: all 4 reviewers flagged this**

Old: `var_os("ZKSYNC_RISC_V_RUN").is_some()` — any non-empty value (`1`, `yes`, `on`) triggered RISC-V run.
New: only `"true"`/`"false"` (case-insensitive) are accepted. `"1"`, `"yes"` → treated as unset → RISC-V silently skipped.

CI is unaffected (sets `true`), but **developers using `ZKSYNC_RISC_V_RUN=1` locally will be surprised**. Consider logging a warning for unrecognized non-empty values, or accepting `1`/`yes`/`on` as truthy.

### 2. `mint_tokens_to_treasury` moved out of `Chain::run_inner` — breaks direct callers

`run_inner` now ignores `skip_minting_tokens_to_treasury` (destructured as `_`). Minting moved to `TestingFramework::execute_block_internal`. Direct `Chain` callers no longer get automatic minting:
- `tests/evm_tester/` — `chain.run_block_no_panic()`
- `tests/instances/eth_runner/` — `chain.run_block_with_extra_stats()`
- `tests/fuzzer/` — `chain.run_block()`

These are currently excluded from workspace tests, so CI won't catch breakage. But it's a **latent API contract change** that should be documented or fixed.

### 5. `zksync_os_revm_runner` is an unconditional dependency of `rig`

This increases compile time for ALL test crates, even when REVM checking is disabled. Consider making it an optional feature-gated dependency.


Removed 3 and 4, one was the thing I pointed out about risc build in ci for revm tests, the other irrelevant. I don't think any of the comments are blocking, maybe 1 and 5 are worth fixing

.github/workflows/ci.yml
0xVolosnikov: Yes, I initially wanted to. I just have to skip multiblock tests then, because they always expect to have the multiblock binary

antoniolocascio: Ah ok, then maybe not worth skipping

0xVolosnikov: Yep, let's keep as is for now

0xVolosnikov:
[AI] Thanks for the consolidated review; I verified points 1/2/5 directly in code.

1) `parse_explicit_bool`: confirmed. Current behavior intentionally requires explicit `true` / `false`; values like `1`/`yes` are treated as unset. CI is unaffected (`CI=true`, and our docs also use `ZKSYNC_RISC_V_RUN=true`). I agree there is local surprise risk for users of `=1`; we should add a warning for unsupported non-empty values (and can optionally accept `1/yes/on` in a follow-up if we want backward-compat behavior).

2) Treasury minting relocation: confirmed. `mint_tokens_to_treasury` now runs in `TestingFramework::execute_block_internal`, and direct `Chain::*run_block*` callers no longer auto-mint. In-tree direct callers currently pre-seed state/balances explicitly, so there is no immediate breakage observed, but I agree this is a latent API contract change and should be documented (or re-aligned in a follow-up).

5) Unconditional `zksync_os_revm_runner` dependency in `rig`: valid point. It is currently unconditional and can increase compile work when REVM checking is not used. I agree this is worth feature-gating in a focused follow-up to keep this PR scope manageable.

0xVolosnikov:
[AI] Addressed the reviewed items in two follow-up commits:

1) `118f33fb` (`fix(rig): accept common boolean env aliases in RunConfig`)
- `RunConfig` env parsing now accepts `true/false`, `1/0`, `yes/no`, `on/off` (case-insensitive, trimmed).
- Unsupported non-empty values now emit a warning instead of silently falling back.
- Added/updated unit tests for aliases and parsing behavior.

2) `7b202f66` (`fix(rig): keep treasury minting control in TestingFramework`)
- Removed treasury minting toggle from `RunConfig` (keeps `RunConfig` runtime-only).
- Kept treasury auto-minting as `TestingFramework` setup behavior.
- Added framework controls: `without_minting_tokens_to_treasury` / `disable_minting_tokens_to_treasury`.
- Migrated the transactions test that previously used `RunConfig.skip_minting_tokens_to_treasury`.
- Added tests to assert:
  - `Chain` does not auto-mint treasury.
  - `TestingFramework` mints by default.
  - `TestingFramework` can disable automatic minting.
- Added in-code docs clarifying `Chain` vs `TestingFramework` responsibilities.

Targeted tests were run and passed for these changes.

PR #521, opened 2026-02-18
feat: implement getProof verifier and testing prover

antoniolocascio:
## What ❔
Add verifier for zks_getProof to the API and a testing prover. Main prover will live in the server.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/basic_system/src/system_implementation/flat_storage_model/get_proof.rs
+++ b/basic_system/src/system_implementation/flat_storage_model/get_proof.rs
@@ -1,439 +1,26 @@
 //! Helpers for the `zks_getProof` API.
 //!
-//! The module defines JSON-shaped proof structs and provides separate prover and verifier helpers.
-//! The prover is only a lightweight testing utility backed by `FlatStorageBacking`.
-//! The verifier is spec-accurate and can be used by external consumers.
+//! This module keeps prover/testing integration for `FlatStorageBacking`.
+//! Shared `zks_getProof` types and verification logic live in the standalone
+//! `zks_get_proof_verifier` crate.
 
-use alloc::vec::Vec;
+use zk_ee::utils::Bytes32 as StorageBytes32;
 
-use ruint::aliases::B160;
-use zk_ee::utils::Bytes32;
+pub use zks_get_proof_verifier::{
+    compute_state_commitment, verify_response, LeafWithProof, StateCommitmentPreimage,
+    StorageProof, StorageProofType, ZksGetProofHasher, ZksGetProofResponse,
+    ZksGetProofVerificationError, ZERO_32_BYTES,
+};
 
-use super::TREE_HEIGHT;
-
-pub use verifier::{compute_state_commitment, ZksGetProofVerificationError};
-
-/// Preimage data required to recompute the L1 batch commitment.
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub struct StateCommitmentPreimage {
-    #[cfg_attr(feature = "serde", serde(with = "serde_hex::u64"))]
-    pub next_free_slot: u64,
-    #[cfg_attr(feature = "serde", serde(with = "serde_hex::u64"))]
-    pub block_number: u64,
-    #[cfg_attr(feature = "serde", serde(with = "serde_hex::bytes32"))]
-    pub last256_block_hashes_blake: Bytes32,
-    #[cfg_attr(feature = "serde", serde(with = "serde_hex::u64"))]
-    pub last_block_timestamp: u64,
+#[inline(always)]
+fn to_proof_bytes32(value: StorageBytes32) -> [u8; 32] {
+    value.as_u8_array()
 }
 
-/// Response envelope for `zks_getProof`.
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub struct ZksGetProofResponse {
-    #[cfg_attr(feature = "serde", serde(with = "serde_hex::b160"))]
-    pub address: B160,
-    pub state_commitment_preimage: StateCommitmentPreimage,
-    pub storage_proofs: Vec<StorageProof>,
-}
-
-/// A proof for a single storage key.
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub struct StorageProof {
-    #[cfg_attr(feature = "serde", serde(with = "serde_hex::bytes32"))]
-    pub key: Bytes32,
-    pub proof: StorageProofType,
-}
-
-/// A leaf and its Merkle path, used to prove non-existence.
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub struct LeafWithProof {
-    #[cfg_attr(feature = "serde", serde(with = "serde_hex::u64"))]
-    pub index: u64,
-    #[cfg_attr(feature = "serde", serde(with = "serde_hex::bytes32"))]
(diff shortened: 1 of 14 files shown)


basic_system/src/system_implementation/flat_storage_model/get_proof.rs
```diff
@@ -0,0 +1,596 @@
...
+
+use super::TREE_HEIGHT;
+
+pub use verifier::{compute_state_commitment, ZksGetProofVerificationError};
+
+/// Preimage data required to recompute the L1 batch commitment.
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct StateCommitmentPreimage {
```
slowli: It would be preferable to reuse these API types in the server code. Unfortunately, `basic_system` seems somewhat too heavy in terms of the dependency tree for this to work. Is there any chance to move these types to a more lightweight crate? (No problem if there isn't; I'd try to deduplicate types anyway, and there aren't *that* many types, so copying them is an option.)

basic_system/src/system_implementation/flat_storage_model/get_proof.rs
```diff
@@ -0,0 +1,596 @@
...
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(tag = "type", rename_all = "camelCase"))]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum StorageProofType {
+    Existing {
+        #[cfg_attr(feature = "serde", serde(with = "serde_hex::u64"))]
+        index: u64,
+        #[cfg_attr(feature = "serde", serde(with = "serde_hex::bytes32"))]
+        value: Bytes32,
+        #[cfg_attr(feature = "serde", serde(rename = "nextIndex"))]
```
slowli: Nit: Can't [`rename_all_fields`](https://serde.rs/container-attrs.html#rename_all_fields) be used on the containing enum?

basic_system/src/system_implementation/flat_storage_model/get_proof.rs
```diff
@@ -0,0 +1,596 @@
...
+        empty_hashes: &[Bytes32; N],
+    ) -> Result<Bytes32, ZksGetProofVerificationError> {
+        use alloc::boxed::Box;
+        if siblings.len() > N {
+            return Err(ZksGetProofVerificationError::SiblingsTooLong {
+                len: siblings.len(),
+            });
+        }
+
+        let mut path = Box::new_in([Bytes32::ZERO; N], Global);
```
slowli: Are allocations a concern? They could be somewhat easily avoided here.

basic_system/src/system_implementation/flat_storage_model/get_proof.rs
```diff
@@ -0,0 +1,596 @@
...
+                    if left_root != right_root {
+                        return Err(ZksGetProofVerificationError::NonExistingRootMismatch);
+                    }
+                    (left_root, Bytes32::ZERO)
+                }
+            };
+
+            let commitment =
+                compute_state_commitment(&state_root, &response.state_commitment_preimage);
+            if &commitment != expected_batch_hash {
```
slowli: Nit: Are extra hash operations not a concern? This fragment could be reworked to call `compute_state_commitment` once and then just compare recovered state root hashes on subsequent iterations. Then again, if the number of hashing operations a concern, having separate Merkle paths for storage slots is suboptimal on its own 🙃

basic_system/src/system_implementation/flat_storage_model/get_proof.rs
antoniolocascio: Right, missed it while checking docs. Thanks

basic_system/src/system_implementation/flat_storage_model/get_proof.rs
antoniolocascio: Not sure how much of a concern, but better get rid of that. Thanks

basic_system/src/system_implementation/flat_storage_model/get_proof.rs
antoniolocascio: Doubt that this extra hashing will be a concern, but I've implemented your suggestion. I think this method will largely be used for single proofs, so I guess I didn't consider the batching optimization too much

zks_get_proof_verifier/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+    Existing {
+        #[cfg_attr(feature = "serde", serde(with = "serde_hex::u64"))]
+        index: u64,
+        #[cfg_attr(feature = "serde", serde(with = "serde_hex::bytes32"))]
+        value: [u8; 32],
+        #[cfg_attr(feature = "serde", serde(rename = "nextIndex"))]
+        #[cfg_attr(feature = "serde", serde(with = "serde_hex::u64"))]
+        next_index: u64,
+        #[cfg_attr(feature = "serde", serde(with = "serde_hex::vec_bytes32"))]
+        siblings: Vec<[u8; 32]>,
```
slowli, line 82: It's possible to reuse these 4 fields with `LeafWithProof` via `#[serde(flatten)]`. I've (temporarily) modeled API types like this in [my PR](https://github.com/matter-labs/zksync-os-server/blob/de038340e1d7203a3e9068b586a2bdcbea65ad5e/lib/merkle_tree_api/src/api.rs).

zks_get_proof_verifier/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+        leaf: &FlatStorageLeaf,
+        path: &[[u8; 32]; N],
+    ) -> [u8; 32] {
+        let leaf_hash = hash_leaf(hasher, leaf);
+
+        let mut current = leaf_hash;
+        let mut index = index;
+        for path in path.iter() {
+            let path: &[u8; 32] = path;
+            let (left, right) = if index & 1 == 0 {
```
slowli, line 353: Bikeshedding: Don't really have time to play *much* with the disaassembler / [Godbolt](https://godbolt.org/), but AFAIU, `index % 2 == 0` (and `index /= 2` below) would be transformed to binary and / shift right anyway, and they would look more human-readable.

zks_get_proof_verifier/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+        response: &ZksGetProofResponse,
+        expected_batch_hash: &[u8; 32],
+        hasher: &mut H,
+    ) -> Result<Vec<[u8; 32]>, ZksGetProofVerificationError> {
+        // Handle case for 0 proofs:
+        if response.storage_proofs.is_empty() {
+            return Ok(alloc::vec![]);
+        }
+
+        let empty_hashes = compute_empty_hashes::<N, H>(hasher);
```
slowli, line 278: JFYI, the server implementation goes one step further and [has](https://github.com/matter-labs/zksync-os-server/blob/de038340e1d7203a3e9068b586a2bdcbea65ad5e/lib/merkle_tree_api/src/hasher.rs#L81-L82) empty subtree hashes precomputed once on program initialization. (It could be possible to go even further and embed the hash values as a constant, e.g. covering its validity via a  dedicated unit test.) Doesn't matter much from the server integration perspective.

zks_get_proof_verifier/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+    NeighborOrderInvalid,
+    NeighborLinkInvalid,
+    StateCommitmentMismatch,
+}
+
+/// Minimal digest interface used by verifier internals.
+pub trait ZksGetProofHasher {
+    fn update(&mut self, input: impl AsRef<[u8]>);
+    fn finalize_reset(&mut self) -> [u8; 32];
+}
```
slowli, line 105: - JFYI, [the server also has a hasher trait](https://github.com/matter-labs/zksync-os-server/blob/de038340e1d7203a3e9068b586a2bdcbea65ad5e/lib/merkle_tree_api/src/hasher.rs#L9-L24), and it's more high-level. Subjectively, it's easier to work with, but I'm obviously biased as the author 🙃
- Bikeshedding: `impl AsRef` looks overly complicated; just `input: &[u8]` would suffice (and make the trait dyn-compatible, although that's not a requirement).

slowli (commented):
Overall, looks good 👍

zks_get_proof_verifier/src/lib.rs
antoniolocascio: I basically copied the one in our crypto crate. Will take a look at it tomorrow, thanks!

antoniolocascio:
All this logic is now implemented by the server

PR #522, opened 2026-02-18
fix: do not send l1 log for upgrade tx result

AntonD3:
## What ❔

Remove l1 -> l2 log for upgrade tx results.

## Why ❔

By protocol from v31 we shouldn't send them, GW asset tracker relies on it.

## Is this a breaking change?
- [x] Yes
- [ ] No

## Checklist


antoniolocascio (approved):
LGTM, can you add an assert to `test_upgrade_tx_succeeds` checking no l1l2 log was emitted? Just to prevent some regression

PR #523, opened 2026-02-19
fix: observe account read if previously touched

antoniolocascio:
## What ❔

Current account cache doesn't set the account as observed on a read if it was previously touched. This PR fixes that. 

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/basic_system/src/system_implementation/caches/cache_element_properties.rs
+++ b/basic_system/src/system_implementation/caches/cache_element_properties.rs
@@ -25,14 +25,14 @@ pub struct CacheElementProperties {
 }
 
 impl CacheElementProperties {
-    pub fn new(is_new_element: bool, is_value_known: bool) -> Self {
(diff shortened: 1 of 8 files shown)


shamatar:
@antoniolocascio I think in the latest part of ethproofs branch I eliminated/renamed obscurely named "mark as known" call to actually reflect that the value is indeed was observed and so should be eventually verified vs state. Did it get lost eventually?

antoniolocascio:
@shamatar I think @0xVolosnikov kept the old name when backporting this part, but nothing else got lost IIRC

shamatar:
Ok, can we batch-rename it then to "mark observed" or something meaningful?

antoniolocascio:
@shamatar sure, done it now

shamatar (approved):

PR #526, opened 2026-02-20
feat: Return back contract deployer

Artemka374:
## What ❔

This is partial revert of #414


perekopskiy (approved):

0xVolosnikov (approved):

0xVolosnikov:
Note that the actual EVM contract on that address will de ignored. Imo this is fine for now

PR #532, opened 2026-02-24
test: Introduce TestingFramework struct

0xVolosnikov:
## What ❔

Additional layer of abstraction between tests and `Chain`. This is an intermediate step - in following PRs I will better separate concerns between `Chain` and `TestingFramework`. Additionally this separation will make RevmRunner integration easier and more effective.

Note that in this PR we'll change how testing works in general - by default locally RISC-V simulation will be disabled, it will be enabled in CI. This change makes testing setup a bit more friendly for AI agents.

## Is this a breaking change?
- [ ] Yes
- [X] No

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -84,9 +84,9 @@ To run rig-based tests without detailed executions logs use:
 cargo test -p <crate-name> --features rig/no_print
 ```
 
-By default, local rig runs skip RISC-V simulation unless `ZKSYNC_PROVING_RUN` or `CI` is set. To force proving-mode simulation locally use:
+By default, local rig runs skip RISC-V simulation unless `ZKSYNC_RISC_V_RUN` or `CI` is set. To force proving-mode simulation locally use:
 ```bash
-ZKSYNC_PROVING_RUN=true cargo test -p <crate-name>
+ZKSYNC_RISC_V_RUN=true cargo test -p <crate-name>
 ```
 
 #### EVM tester
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -72,10 +72,17 @@ cd zksync_os && ./dump_bin.sh --type for-tests
 ### Testing Infrastructure
 
 #### Integration tests
-The project uses a custom testing rig located in `tests/rig/` with the main abstraction being the `Chain` struct for in-memory chain state testing. Tests are organized in `tests/instances/` and follow this pattern:
+The project uses a custom testing rig located in `tests/rig/` with the main abstraction being the `ZKsyncOSTester` struct. Tests are organized in `tests/instances/` and follow this pattern:
 1. Set up initial chain state (predeployed contracts, balances)
 2. Define transactions to execute
-3. Call `run_block` to execute (typically runs both forward and proof systems, unless configured otherwise)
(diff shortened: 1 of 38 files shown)


AGENTS.md
```diff
@@ -72,10 +72,22 @@ cd zksync_os && ./dump_bin.sh --type for-tests
...
+3. Call `execute_block` to execute (typically runs both forward and proof systems, unless configured otherwise)
+
+Avoid direct `Chain` usage in test instances unless a low-level path is explicitly needed.
+
+To run rig-based tests without detailed executions logs use:
+```bash
+cargo test -p <crate-name> --features rig/no_print
+```
+
+By default, local rig runs skip RISC-V simulation unless `ZKSYNC_PROVING_RUN` or `CI` is set. To force proving-mode simulation locally use:
```
antoniolocascio: I'm not sure about this. I would prefer to keep risc run on by default, and tell agents here to always set it to false. I've encountered issues when running tests locally and I'm afraid I'll forget to enable it sometimes.

tests/instances/transactions/src/lib.rs
```diff
@@ -4,62 +4,42 @@
...
-        app: Some("for_tests".to_string()),
-        only_forward: false,
-        check_storage_diff_hashes: true,
-        skip_minting_tokens_to_treasury: false,
-        ..Default::default()
-    })
-}
+const PRIMARY_TEST_PK: &str = "dcf2cbdd171a21c480aa7f53d77f31bb102282b3ff099c78e3118b37348c72f7";
+const SECONDARY_TEST_PK: &str = "a226d3a5c8c408741c3446c762aee8dff742f21e381a0e5ab85a96c5c00100be";
+const TERTIARY_TEST_PK: &str = "abcdebdd171a21c480aa7f53d77f31bb102282b3ff099c78e3118b37348c72f7";
```
antoniolocascio: Maybe we can add these to framework? 
Sometimes we can't use random_signer for reproducibility, so it might be good to have a few easily-accessible wallets.

tests/evm_tester/src/vm/zk_ee/mod.rs
```diff
@@ -101,7 +101,7 @@ impl ZKsyncOS {
 
         let run_config = RunConfig {
             app: Some("evm_tester".to_string()),
-            only_forward: !proof_run,
+            do_riscv_run: proof_run,
```
antoniolocascio, line 104: Do we want to refactor evm_tester / eth_runner to reuse the framework? Ofc we can do as follow up

antoniolocascio (approved):
Generally LGTM, left a couple comments

tests/evm_tester/src/vm/zk_ee/mod.rs
0xVolosnikov: Yes, I want to do this, but it requires some additional changes (e.g. running Eth STF), so I decided to limit the scope of this PR

tests/instances/transactions/src/lib.rs
0xVolosnikov: Makes sense, will do

AGENTS.md
0xVolosnikov: As discussed, we will experiment and try to use it this way

PR #538, opened 2026-02-27
fix: Make contract deployer hook EVM-friendly

0xVolosnikov:
## What ❔

Ppretend to be an empty account if caller is unauthorized

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


0xVolosnikov:
We should check that this change is compatible with our upgrade flow. Alternatively we can on a higher level pretend that ContractDeployer isn't a hook if caller isn't X. But still we have to figure the exact upgrade flow first.

Iiuc, we want to deploy proxy there and then initialize it, so I guess it's already a bit complex (huh): https://github.com/matter-labs/era-contracts/blob/84e5c59386b6b324d6547c9ad551e43f9a22fc93/l1-contracts/contracts/l2-upgrades/L2ComplexUpgrader.sol#L39

UPD: it was decided to keep ContractDeployer hook and remove ContractDeployer contract for now

antoniolocascio (approved):

PR #540, opened 2026-02-27
feat(EVM-1178): add point evaluation test vectors

0xVolosnikov:
## What ❔
- Added point-evaluation cost vector tests for valid and invalid inputs.
- Verified expected ergs/native charging behavior and removed a stale TODO in cost constants.

## Why ❔
- Protects fee-accounting behavior in point evaluation from regressions.
- Makes expected charging behavior explicit via tests.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


0xVolosnikov:
Doesn't actually cover the task

PR #542, opened 2026-02-27
feat(EVM-1173): [AI] adapt fuzzer mutator to current tx schema

0xVolosnikov:
## What ❔
- Adapted the fuzzer transaction mutator to the current ABI-encoded transaction format.
- Added invariant-preserving mutation and serialization flow for mutated transactions.

## Why ❔
- Restores effective fuzzing after transaction schema evolution.
- Improves mutation coverage while preserving transaction validity constraints.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/tests/fuzzer/fuzz/fuzz_targets/bootloader/bootloader_tx_parser.rs
+++ b/tests/fuzzer/fuzz/fuzz_targets/bootloader/bootloader_tx_parser.rs
@@ -26,7 +26,6 @@ fn fuzz(data: &[u8]) {
     let _ = transaction.encoding(transaction.paymaster_input.clone());
     let _ = transaction.encoding(transaction.signature.clone());
     let _ = transaction.len();
-    let _ = transaction.sig_parity_r_s();
     let _ = transaction.calculate_hash(&mut inf_resources);
     let _ = serialize_zksync_transaction(&transaction);
 }
--- a/tests/fuzzer/fuzz/fuzz_targets/bootloader/bootloader_tx_parser.rs
+++ b/tests/fuzzer/fuzz/fuzz_targets/bootloader/bootloader_tx_parser.rs
@@ -1,21 +1,19 @@
 #![no_main]
 #![feature(allocator_api)]
 
-
-use basic_bootloader::bootloader::transaction::AbiEncodedTransaction;
-use common::mutate_transaction;
+use common::{mutate_transaction, parse_abi_encoded_transaction, serialize_zksync_transaction};
 use libfuzzer_sys::{fuzz_mutator, fuzz_target};
 use zk_ee::reference_implementations::{BaseResources, DecreasingNative};
-mod common;
 use zk_ee::system::Resource;
+
+mod common;
 
 fuzz_mutator!(|data: &mut [u8], size: usize, max_size: usize, seed: u32| {
     mutate_transaction(data, size, max_size, seed)
 });
 
 fn fuzz(data: &[u8]) {
-    let mut data = data.to_owned();
-    let Ok(transaction) = AbiEncodedTransaction::try_from_slice(&mut data) else {
+    let Ok(transaction) = parse_abi_encoded_transaction(data) else {
         // Input is not valid
         return;
     };
@@ -25,17 +23,12 @@ fn fuzz(data: &[u8]) {
     let _ = transaction.tx_type.read();
     let _ = transaction.required_balance();
     let _ = transaction.calldata();
-    let _ = transaction.paymaster_input();
-    let _ = transaction.signature();
-    let _ = transaction.tx_body_length();
-
-    let chain_id = 0;
-    let _ = transaction.calculate_signed_hash(chain_id, &mut inf_resources);
-    let _ = transaction.calculate_hash(chain_id, &mut inf_resources);
-
-    let mut transaction = transaction;
-    let _ = transaction.underlying_buffer();
-    let _ = transaction.pre_tx_buffer();
+    let _ = transaction.encoding(transaction.paymaster_input.clone());
+    let _ = transaction.encoding(transaction.signature.clone());
+    let _ = transaction.len();
+    let _ = transaction.sig_parity_r_s();
+    let _ = transaction.calculate_hash(&mut inf_resources);
+    let _ = serialize_zksync_transaction(&transaction);
 }
 
 fuzz_target!(|data: &[u8]| {
--- a/tests/fuzzer/fuzz/fuzz_targets/bootloader/common/mod.rs
+++ b/tests/fuzzer/fuzz/fuzz_targets/bootloader/common/mod.rs
@@ -284,8 +284,8 @@ fn mutate_zksync_transaction(tx: &mut TransactionData, rng: &mut StdRng) {
         }
         1 => mutate_address_inplace(&mut tx.from, rng),
         2 => mutate_address_inplace(&mut tx.to, rng),
-        3 => tx.gas_limit = mutate_u256_vec(tx.gas_limit, rng),
-        4 => tx.gas_per_pubdata_limit = mutate_u256_vec(tx.gas_per_pubdata_limit, rng),
+        3 => tx.gas_limit = mutate_u64_field(tx.gas_limit, rng),
+        4 => tx.gas_per_pubdata_limit = mutate_u32_field(tx.gas_per_pubdata_limit, rng),
         5 => tx.max_fee_per_gas = mutate_u256_vec(tx.max_fee_per_gas, rng),
         6 => tx.max_priority_fee_per_gas = mutate_u256_vec(tx.max_priority_fee_per_gas, rng),
         7 => tx.nonce = mutate_u256_vec(tx.nonce, rng),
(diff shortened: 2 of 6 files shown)


0xVolosnikov:
@vv-dev-ai ptal

tests/fuzzer/fuzz/fuzz_targets/bootloader/common/mod.rs
```diff
@@ -415,16 +342,31 @@ fn mutate_u256_vec(num: U256, rng: &mut StdRng) -> U256 {
...
+}
+
+fn mutate_u32_field(num: U256, rng: &mut StdRng) -> U256 {
+    mutate_low_bytes(num, 4, rng)
+}
+
+fn mutate_low_bytes(num: U256, bytes_to_mutate: usize, rng: &mut StdRng) -> U256 {
+    let mut mutated_bytes = num.to_be_bytes();
+    let start = 32usize.saturating_sub(bytes_to_mutate);
+    let idx = rng.gen_range(start..mutated_bytes.len());
```
vv-dev-ai, line 366: [AI-review] `mutate_low_bytes` panics when `bytes_to_mutate == 0`: `start = 32usize.saturating_sub(0) = 32`, so `rng.gen_range(32..32)` is an empty range and panics at runtime. Currently unreachable -- `mutate_u64_field` passes 8 and `mutate_u32_field` passes 4 -- but the contract is implicit. Adding a guard (`if bytes_to_mutate == 0 { return num; }`) or `debug_assert!(bytes_to_mutate > 0)` at the top would make the invariant explicit and prevent future callers from stumbling into it.

vv-dev-ai (changes requested):
[AI-review] PR #542

Updates the fuzzer's `common/mod.rs` to work with the current `AbiEncodedTransaction` format: introduces `TransactionData` as a mutable mirror, implements `to_zk_bytes()` serialization and `From<&AbiEncodedTransaction<A>>`, adds an invariant-preserving `mutate_transaction` entry point, and switches from `web3` ABI encoding to a hand-rolled encoder. The core `common/mod.rs` changes are correct and the parse-mutate-serialize-parse roundtrip is sound.

**Verdict: REQUEST_CHANGES**

The PR description says it "restores effective fuzzing," but five of the seven bootloader fuzz targets still use `AbiEncodedTransaction::try_from_slice` (which was removed) and several non-existent methods. Those targets will not compile:

- `bootloader_tx_try_from_slice.rs` -- `try_from_slice` call
- `bootloader_tx_parser.rs` -- `try_from_slice`, `tx.paymaster_input()`, `tx.signature()`, `tx.tx_body_length()`
- `bootloader_tx_validate.rs` -- `try_from_slice`, `calculate_hash(chain_id, resources)` (extra arg vs. current signature)
- `bootloader_tx_calculate_signed_hash.rs` -- `try_from_slice`, `calculate_signed_hash` (method does not exist)
- `bootloader_process_transaction.rs` -- `try_from_slice`

The `parse_abi_encoded_transaction` helper introduced in `common/mod.rs` is the right pattern; the five targets above need to be updated to use it instead of the removed `try_from_slice`.

No correctness issues in the code that is actually changed.

tests/fuzzer/fuzz/fuzz_targets/bootloader/common/mod.rs
```diff
@@ -458,16 +400,20 @@ fn pad32(v: &mut Vec<u8>) {
     }
 }
 
+fn clamp_to_u32(value: usize) -> u32 {
+    u32::try_from(value).unwrap_or(u32::MAX)
```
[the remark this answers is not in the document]
0xVolosnikov: [AI] Fixed in commit 1687e075. `clamp_to_u32` was removed in favor of checked `u32` conversion, and `to_zk_bytes` / `serialize_zksync_transaction` / `mutate_transaction` now early-return when a dynamic offset or length does not fit the `AbiEncodedTransaction` encoding.

tests/fuzzer/fuzz/fuzz_targets/bootloader/common/mod.rs
```diff
@@ -279,126 +225,107 @@ pub fn mock_oracle_balance(
...
+                }
+                _ => AbiEncodedTransaction::<Global>::L1_L2_TX_TYPE,
+            };
+        }
+        1 => mutate_address_inplace(&mut tx.from, rng),
+        2 => mutate_address_inplace(&mut tx.to, rng),
+        3 => tx.gas_limit = mutate_u256_vec(tx.gas_limit, rng),
+        4 => tx.gas_per_pubdata_limit = mutate_u256_vec(tx.gas_per_pubdata_limit, rng),
+        5 => tx.max_fee_per_gas = mutate_u256_vec(tx.max_fee_per_gas, rng),
+        6 => tx.max_priority_fee_per_gas = mutate_u256_vec(tx.max_priority_fee_per_gas, rng),
```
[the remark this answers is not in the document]
0xVolosnikov: [AI] Rechecked against commit 1687e075: this concern is already addressed in the current mutator. `gas_limit` goes through `mutate_u64_field` and `gas_per_pubdata_limit` goes through `mutate_u32_field`, so only the validated low-width bytes are mutated and we do not flip the high 24/28 bytes anymore.

tests/fuzzer/fuzz/fuzz_targets/bootloader/common/mod.rs
```diff
@@ -10,161 +10,107 @@ use oracle_provider::ZkEENonDeterminismSource;
...
+        for value in self.reserved {
+            head.push(enc_u256(value));
+        }
+
+        let head_size = 19 * 32;
+        abi_push_bytes(&mut head, &mut tail, &self.data, head_size);
+        abi_push_bytes(&mut head, &mut tail, &self.signature, head_size);
+        abi_push_bytes32_array(&mut head, &mut tail, &self.factory_deps, head_size);
+        abi_push_bytes(&mut head, &mut tail, &self.paymaster_input, head_size);
+        abi_push_bytes(&mut head, &mut tail, &self.reserved_dynamic, head_size);
```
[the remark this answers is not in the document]
0xVolosnikov: [AI] Fixed in commit 1687e075. The shared ABI serializer now encodes dynamic offsets and lengths through checked `u32` conversions, and the mutation path bails out if a value does not fit the ABI schema instead of silently serializing a saturated offset/length.

tests/fuzzer/fuzz/fuzz_targets/bootloader/common/mod.rs
0xVolosnikov: [AI] Fixed in commit 1687e075. `mutate_low_bytes` now returns early for `bytes_to_mutate == 0` and clamps the requested width to the 32-byte backing array before choosing the mutation index, so future callers cannot hit an empty range.

0xVolosnikov:
[AI] @vv-dev-ai Addressed in commit 1687e075. The remaining bootloader ABI fuzz targets now use the shared `parse_abi_encoded_transaction` helper instead of the removed `try_from_slice` path, the stale hash/signing call sites were updated to the current API surface, and the shared serializer/mutator helpers were tightened around `u32` ABI widths and zero-width mutation requests.

tests/fuzzer/fuzz/fuzz_targets/bootloader/bootloader_tx_parser.rs
```diff
@@ -25,17 +23,12 @@ fn fuzz(data: &[u8]) {
...
-    let _ = transaction.calculate_signed_hash(chain_id, &mut inf_resources);
-    let _ = transaction.calculate_hash(chain_id, &mut inf_resources);
-
-    let mut transaction = transaction;
-    let _ = transaction.underlying_buffer();
-    let _ = transaction.pre_tx_buffer();
+    let _ = transaction.encoding(transaction.paymaster_input.clone());
+    let _ = transaction.encoding(transaction.signature.clone());
+    let _ = transaction.len();
+    let _ = transaction.sig_parity_r_s();
```
vv-dev-ai: [AI-review] `transaction.sig_parity_r_s()` will panic on every valid input to this fuzz target.

`parse_abi_encoded_transaction` only succeeds for `L1_L2_TX_TYPE` and `UPGRADE_TX_TYPE` (all other tx types are rejected by `validate_structure`). For both of those types, `validate_structure` requires the signature field to be empty (`!self.signature.range.is_empty()` → `Err`). So `self.signature.range` is always empty here.

`AbiEncodedTransaction::sig_parity_r_s` does `let r = &signature[..32]` with no length guard; on an empty slice this panics immediately (`range end index 32 out of range for slice of length 0`). The fuzz target crashes on every valid transaction it generates, making it useless as a coverage driver.

Fix: guard the call or skip it for L1/upgrade transactions:
```rust
// ABI-encoded (L1/upgrade) transactions carry no ECDSA signature; skip.
// let _ = transaction.sig_parity_r_s();
```

vv-dev-ai (changes requested):
[AI-review] PR #542

Completes the migration of all seven bootloader fuzz targets from the removed `AbiEncodedTransaction::try_from_slice` API to `parse_abi_encoded_transaction`, and activates the `mutate_transaction` / `TransactionData` serialization flow.

**Verdict: REQUEST_CHANGES**

✅ RESOLVED — Five fuzz targets updated to use `parse_abi_encoded_transaction` instead of removed `try_from_slice` / `calculate_signed_hash(chain_id, …)` / `signature()` APIs.
✅ RESOLVED — `mutate_low_bytes` panic on `bytes_to_mutate == 0` fixed; guard added, and `start` computation changed to non-overflowing `mutated_bytes.len() - bytes_to_mutate`.

One new issue blocks this PR (see inline comment on `bootloader_tx_parser.rs` line 29).

**Informational (no change required):** `bootloader_tx_calculate_signed_hash.rs` now wraps the parsed transaction in `Transaction::Abi(tx)` and calls `transaction.signed_hash()`, which always returns `Err(internal_error!(…))` for ABI transactions. The target silently becomes a parse-only exerciser. This is a behaviour change (was previously testing the signing hash path) but given that ABI transactions have no signed hash this is intentional and acceptable.

Adversarial analysis: no additional failure paths found beyond the one noted above.

tests/fuzzer/fuzz/fuzz_targets/bootloader/bootloader_tx_parser.rs
0xVolosnikov: [AI] Fixed in commit 4f29b653. `parse_abi_encoded_transaction` only yields ABI L1/upgrade transactions, so calling `sig_parity_r_s()` here was indeed invalid because their signature field is required to be empty. I removed that call and kept the target on the non-panicking parser/hash/roundtrip surface instead.

0xVolosnikov:
@antoniolocascio looks okey, enables old disabled fuzzying targets

antoniolocascio (approved):

PR #544, opened 2026-03-04
test: overhaul integration testing infrastructure (rig, macros, new test crates, docs)

vv-dev-ai:
## Summary

This PR overhauls the ZKsync OS integration testing infrastructure to make it simpler to write, read, and extend tests — both by humans and AI agents.

### New rig modules (`tests/rig/src/`)

| Module | What was added |
|--------|---------------|
| `assertions.rs` | 12 assertion macros: `assert_tx_success!`, `assert_tx_reverted!`, `assert_tx_failed!`, `assert_all_success!`, `assert_gas_used_lt/gt/between!`, `assert_event_emitted/not_emitted!`, `assert_block_events_count!`, `assert_account_balance!`, `assert_nonce!` |
| `builder.rs` | `ChainBuilder` + `TxBuilder` fluent APIs; adds `.access_list()` so EIP-2930/1559 tests don't need to drop to raw structs |
| `constants.rs` | Named constants for chain IDs, gas limits, fee params, system addresses |
| `run_config.rs` | `forward_only()`, `full_proof()`, `with_profiler()`, `with_witness_dump()` presets — replaces manual `RunConfig { ... }` construction |

### New test crates

**`tests/instances/errors/`** (18 tests)
- OOG rejections (intrinsic gas, mid-execution, deployment)
- Invalid transaction rejections (wrong chain ID, nonce too low, bad signature, insufficient balance)
- EVM-level reverts (explicit, constructor, nested call)
- State isolation (revert must not mutate storage)
- EIP-1153: TSTORE rolled back on REVERT
- SELFDESTRUCT in a reverting sub-frame has no effect

**`tests/instances/edge_cases/`** (16 tests)
- Zero-value transfer, self-transfer, empty calldata
- Multi-tx state dependencies within a block
- Multi-block state persistence
- EIP-1153: TSTORE/TLOAD round-trip within same tx
- EIP-1153: transient storage cleared between transactions (single dispatch contract)
- PREVRANDAO opcode reads `BlockContext.mix_hash`
- COINBASE opcode reads `BlockContext.coinbase`
- L1 tx bypasses chain-ID validation (design invariant)
- LOG4 with four topics (correct EVM stack-pop order)
- `account_diffs` field populated after ETH transfer

### Documentation

`tests/TESTING.md` — comprehensive guide targeting both humans and AI agents:
- All `ChainBuilder` / `TxBuilder` methods with examples
- Every `RunConfig` preset with when-to-use guidance
- All 12 assertion macros with correct import instructions (Rust 2021 edition)
- Full `BlockOutput` and `TxOutput` field reference (including `account_diffs`, `gas_refunded`, `logs`)
- `run_block_no_panic` and `simulate_block` semantics
- Failure-type decision table (`assert_tx_failed!` vs `assert_tx_reverted!`)
- Step-by-step guide for adding a new test crate

## Test plan

- [ ] `cargo check -p errors --release` — passes (verified locally)
- [ ] `cargo check -p edge_cases --release` — passes (verified locally)
- [ ] `cargo check -p rig --release` — passes (all new rig modules compile)
- [ ] `cargo test -p errors --release` — new error-path tests
- [ ] `cargo test -p edge_cases --release` — new edge-case tests
- [ ] Existing test crates unaffected (`cargo test --workspace --exclude zksync_os --release`)

## Known follow-up items

- `tstore_reverts_on_frame_revert` verifies a second contract's transient slot is zero rather than the original contract's slot. A follow-up can strengthen this using the same calldata-dispatch pattern as `tstore_cleared_between_txs`.
- Several HIGH-priority test ideas from the audit (EIP-7702 edge cases, L1 messenger static/delegate call rejection, EXTCODECOPY, pubdata_limit enforcement) are deferred to follow-up PRs.
- Existing test crates (`transactions/`, `erc20/`, etc.) still construct `RunConfig` manually rather than using the new presets — migration can be done incrementally.

### DIFF
--- a/tests/instances/edge_cases/src/lib.rs
+++ b/tests/instances/edge_cases/src/lib.rs
@@ -102,17 +102,41 @@ fn empty_calldata_call_to_contract() {
 
 // ─── Multiple txs with state dependencies ────────────────────────────────────
 
-/// Two transactions in the same block where tx1 writes a slot and tx2 reads it.
-/// Both must succeed and the state produced by tx1 must be visible to tx2.
+/// Two transactions in the same block where tx0 writes a slot and tx1 reads it.
+/// Both must succeed and tx1 must see the value written by tx0.
+///
+/// Both transactions call the SAME contract so that tx1 reads from the exact
+/// storage namespace that tx0 wrote to. Using separate contracts would test
+/// cross-account isolation (always 0), not within-block state visibility.
+///
+/// Contract dispatch (by calldata presence):
+///   - Non-empty calldata → SSTORE slot 0 ← 0xAB, STOP
+///   - Empty calldata     → SLOAD slot 0, MSTORE, RETURN 32 bytes
+///
+/// Bytecode layout:
+///   Offset  Byte  Instruction
+///    0      36    CALLDATASIZE
+///    1      60 00 PUSH1 0
+///    3      14    EQ
+///    4      60 0d PUSH1 13    ← JUMPDEST offset
+///    6      57    JUMPI
+///    7      60 ab PUSH1 0xAB  ← write path
+///    9      60 00 PUSH1 0
+///   11      55    SSTORE
+///   12      00    STOP
+///   13      5b    JUMPDEST    ← read path
+///   14      60 00 PUSH1 0
+///   16      54    SLOAD
+///   17      60 00 PUSH1 0
+///   19      52    MSTORE
+///   20      60 20 PUSH1 32
+///   22      60 00 PUSH1 0
+///   24      f3    RETURN
 #[test]
 fn multi_tx_block_state_dependency() {
-    // Writer contract: PUSH1 0xAB, PUSH1 0, SSTORE, PUSH1 0, PUSH1 0, RETURN
-    let write_bytecode = hex::decode("60ab60005560006000f3").unwrap();
-    // Reader contract: PUSH1 0, SLOAD, PUSH1 0, MSTORE, PUSH1 32, PUSH1 0, RETURN
-    let read_bytecode = hex::decode("60005460005260206000f3").unwrap();
-
-    let writer_addr = address!("0000000000000000000000000000000000000501");
-    let reader_addr = address!("0000000000000000000000000000000000000502");
+    let contract_bytecode =
+        hex::decode("36600014600d5760ab600055005b60005460005260206000f3").unwrap();
+    let contract_addr = address!("0000000000000000000000000000000000000501");
 
     let signer1 = PrivateKeySigner::random();
     let signer2 = PrivateKeySigner::random();
@@ -122,19 +146,21 @@ fn multi_tx_block_state_dependency() {
     let mut chain = ChainBuilder::new()
         .with_balance(sender1, U256::from(DEFAULT_BALANCE))
         .with_balance(sender2, U256::from(DEFAULT_BALANCE))
-        .with_evm_bytecode(b160(writer_addr), write_bytecode)
-        .with_evm_bytecode(b160(reader_addr), read_bytecode)
+        .with_evm_bytecode(b160(contract_addr), contract_bytecode)
         .build();
 
+    // tx0: non-empty calldata → SSTORE slot 0 = 0xAB
     let tx_write = TxBuilder::new()
         .from(signer1)
-        .to(writer_addr)
+        .to(contract_addr)
+        .calldata(vec![0x01])
         .gas_limit(CALL_GAS_LIMIT)
         .build();
 
+    // tx1: empty calldata → SLOAD slot 0 and return 32 bytes
     let tx_read = TxBuilder::new()
         .from(signer2)
-        .to(reader_addr)
+        .to(contract_addr)
         .gas_limit(CALL_GAS_LIMIT)
         .build();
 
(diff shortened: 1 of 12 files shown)


tests/instances/edge_cases/src/lib.rs
```diff
@@ -0,0 +1,645 @@
...
+/// Both must succeed and the state produced by tx1 must be visible to tx2.
+#[test]
+fn multi_tx_block_state_dependency() {
+    // Writer contract: PUSH1 0xAB, PUSH1 0, SSTORE, PUSH1 0, PUSH1 0, RETURN
+    let write_bytecode = hex::decode("60ab60005560006000f3").unwrap();
+    // Reader contract: PUSH1 0, SLOAD, PUSH1 0, MSTORE, PUSH1 32, PUSH1 0, RETURN
+    let read_bytecode = hex::decode("60005460005260206000f3").unwrap();
+
+    let writer_addr = address!("0000000000000000000000000000000000000501");
+    let reader_addr = address!("0000000000000000000000000000000000000502");
```
0xVolosnikov: P1: This test does not validate the stated cross transaction state dependency. The writer and reader are different contract addresses, so tx2 reads storage of 0x...0502 rather than what tx1 wrote to 0x...0501. This can pass even if same block state visibility is broken. Please make tx2 read from the same account and assert the returned value.

tests/instances/errors/src/lib.rs
```diff
@@ -0,0 +1,645 @@
...
+    //   0x5d           = TSTORE
+    //   0x60 0x00      = PUSH1 0 (size)
+    //   0x60 0x00      = PUSH1 0 (offset)
+    //   0xfd           = REVERT
+    let revert_bytecode = hex::decode("61dead60005d60006000fd").unwrap();
+    let revert_contract = address!("0000000000000000000000000000000000000d01");
+
+    // Check contract: PUSH1 0x00  TLOAD  PUSH1 0x00  MSTORE  PUSH1 0x20  PUSH1 0x00  RETURN
+    //   Returns the 32-byte value of transient slot 0 (should be 0 after the revert)
+    let check_bytecode = hex::decode("60005c60005260206000f3").unwrap();
```
0xVolosnikov: P1: This test claims to verify TSTORE rollback on REVERT, but the TSTORE happens in revert_contract while the TLOAD check is executed in check_contract. Because the read is from a different account context, the assertion can pass even if rollback semantics are wrong. Please read from the same contract context that performed TSTORE and then assert returned bytes are zero.

tests/instances/edge_cases/src/lib.rs
```diff
@@ -0,0 +1,645 @@
...
+    let output =
+        chain.run_block(vec![tx_write, tx_read], None, None, Some(run_config::forward_only()));
+
+    assert_tx_success!(output, 0);
+    assert_tx_success!(output, 1);
+}
+
+// ─── Multiple blocks ─────────────────────────────────────────────────────────
+
+/// State written in block N is visible in block N+1.
```
0xVolosnikov, line 183: P2: This test name says state persists across blocks, but the contract only performs SLOAD and the assertions only check success. There is no write and no value assertion, so the persistence invariant is not exercised. Please make block 1 write a known slot value and block 2 read and assert that exact value.

tests/instances/edge_cases/src/lib.rs
vv-dev-ai: Fixed in 866d7f6d. Replaced the two separate contracts (writer at 0x501, reader at 0x502) with a single calldata-dispatch contract at 0x501. tx0 calls with non-empty calldata → SSTORE slot 0 = 0xAB; tx1 calls with empty calldata → SLOAD slot 0 and returns 32 bytes. The assertion now checks that the returned value equals 0xAB, proving within-block state visibility between two transactions calling the same account.

tests/instances/errors/src/lib.rs
vv-dev-ai: Fixed in 866d7f6d. Replaced the two separate contracts (revert_contract at 0xd01, check_contract at 0xd02) with a single calldata-dispatch contract at 0xd01. tx0 (non-empty calldata) does TSTORE slot 0 ← 0xdead then REVERT; tx1 (empty calldata) does TLOAD slot 0 and returns 32 bytes — both call the same address. Both run in the same block. The assertion checks that tx1's returned bytes are 0x00..00, reading from the same account namespace that tx0 wrote to.

tests/instances/edge_cases/src/lib.rs
vv-dev-ai: Fixed in 866d7f6d. Replaced the read-only contract with a calldata-dispatch contract. Block 1 calls with non-empty calldata → SSTORE slot 0 = 0xBE; block 2 calls with empty calldata → SLOAD slot 0 and returns 32 bytes. The assertion checks that the returned value equals 0xBE, actually verifying the cross-block persistence invariant.

tests/instances/edge_cases/src/lib.rs
0xVolosnikov: Rechecked in commit 866d7f6d: this is fixed. The test now uses one dispatch contract, writes in tx0, reads in tx1, and asserts the returned value (0xAB) from the same storage namespace.

tests/instances/errors/src/lib.rs
0xVolosnikov: Rechecked in commit 866d7f6d: this is fixed. Both txs now call the same contract; tx0 does TSTORE+REVERT and tx1 does TLOAD with an explicit zero-value assertion in the same account context.

tests/instances/edge_cases/src/lib.rs
0xVolosnikov: Rechecked in commit 866d7f6d: this is fixed. Block 1 now writes slot 0 (0xBE), block 2 reads the same slot, and the test asserts the returned value, so persistence is now actually exercised.

PR #553, opened 2026-03-09
fix: fix pubdata estimation for l1 transactions

antoniolocascio:
## What ❔
There's an issue when L1 transactions are understimated in pubdata needed.
The issue is caused by the sender balance being considered "empty" after transferring the value to the destination on simulation, but having some leftovers in actual execution. This PR avoids compression of the sender balance in simulation.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
+++ b/basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
@@ -526,6 +526,14 @@ where
         .ok_or(internal_error!("mfc+tic"))?;
 
     // First we transfer from treasury
+    // We want to ensure that the simulation of a transaction
+    // never underestimates gas/pubdata compared to the actual execution
+    // of said transaction.
+    // During simulation the gas price is typically set to 0. So we need
+    // to be conservative about operations that incur in gas/pubdata depending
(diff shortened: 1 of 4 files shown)


perekopskiy (approved):

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -522,15 +526,15 @@ where
         .ok_or(internal_error!("mfc+tic"))?;
 
     // First we transfer from treasury
-    if to_transfer > U256::ZERO {
+    if to_transfer > U256::ZERO || Config::SIMULATION {
```
0xVolosnikov, line 537: It's not clear why do we want to do this here

0xVolosnikov: I mean, we add some new hacky logical pathes for different cases, but this is not very clear from the code alone. Maybe we should add more comments there for now

antoniolocascio: Fair point, added a comment for it

PR #555, opened 2026-03-10
feat: add run_from_bytes and generate_proof_input_from_bytes APIs

itegulov-claude:
## Summary

- Add `run_from_bytes` to `zksync_os_runner` to run the RISC-V simulator from an in-memory byte slice instead of a file path
- Refactor `generate_proof_input` in `forward_system` to delegate to a new `generate_proof_input_from_bytes`, enabling callers to skip file I/O when the binary is already loaded

## Test plan

- [ ] `cargo check -p zksync_os_runner` passes cleanly
- [ ] Existing tests continue to pass

### DIFF
--- a/zksync_os_runner/src/lib.rs
+++ b/zksync_os_runner/src/lib.rs
@@ -53,57 +53,16 @@ pub fn run(
     run_and_get_effective_cycles(img_path, diagnostics, cycles, non_determinism_source).0
 }
 
-pub fn run_from_bytes(
-    img_bytes: &[u8],
-    diagnostics: Option<DiagnosticsConfig>,
-    cycles: usize,
-    non_determinism_source: impl NonDeterminismCSRSource<VectorMemoryImpl>,
-) -> [u32; 8] {
-    println!("ZK RISC-V simulator is starting");
-
-    let config = SimulatorConfig {
-        bin: BinarySource::Slice(img_bytes),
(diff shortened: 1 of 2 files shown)


zksync_os_runner/src/lib.rs
```diff
@@ -53,6 +53,41 @@ pub fn run(
     run_and_get_effective_cycles(img_path, diagnostics, cycles, non_determinism_source).0
 }
 
+pub fn run_from_bytes(
+    img_bytes: &[u8],
+    diagnostics: Option<DiagnosticsConfig>,
+    cycles: usize,
+    non_determinism_source: impl NonDeterminismCSRSource<VectorMemoryImpl>,
+) -> [u32; 8] {
+    println!("ZK RISC-V simulator is starting");
```
antoniolocascio, line 62: Can you refactor it so that both share the logic? IIUC run_and_get_effective_cycles can reuse this function

antoniolocascio (approved):

zksync_os_runner/src/lib.rs
itegulov: Yup, done. TBH I did see that Claude was being lazy in this PR, thanks for keeping me honest :sweat_smile:

antoniolocascio (approved):
LGTM

PR #557, opened 2026-03-12
feat: replicate L2AssetTracker.handleFinalizeBaseTokenBridgingOnL2

Raid5594:
## What ❔

Replicating `handleFinalizeBaseTokenBridgingOnL2` in zksync-os.

## Why ❔

In Era during mint in Base Token we are calling `L2_ASSET_TRACKER.handleFinalizeBaseTokenBridgingOnL2(L1_CHAIN_ID, _amount);`, however in zksync-os counterpart the mint is performed by direct storage manipulation and not SC call, so we have to manually introduce the respective changes in zksync-os. The functionality is not replicated completely, only the paths that are valid for base token.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -693,3 +702,226 @@ where
...
+/// 1. `_registerLegacyTokenIfNeeded` — for ZKOS chains the base token is registered during upgrade
+/// 2. `_needToForceSetAssetMigrationOnL2` → `_forceSetAssetMigrationNumber`
+/// 3. `chainBalance` update — skipped since base token origin is L1 (not this chain)
+/// 4. `interopInfo[assetId].totalSuccessfulDepositsFromL1 += amount` when settling on L1
+///
+/// IMPORTANT: Must be called BEFORE balance changes, matching the Solidity invariant that
+/// L2AssetTracker is notified before totalSupply/balances change.
+fn handle_finalize_base_token_bridging<S: EthereumLikeTypes>(
+    system: &mut System<S>,
+    amount: &U256,
```
antoniolocascio, line 759: Nit: should we add an early exit for amount == 0? To keep it the same as the solidity contract

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -693,3 +702,226 @@ where
...
+    Bytes32::from_array(hasher.finalize().into())
+}
+
+/// Computes keccak256(abi.encode(key, base_slot_bytes)) for nested Solidity mapping slot derivation.
+fn solidity_mapping_slot_nested(key: &[u8; 32], base_slot_bytes: &[u8; 32]) -> Bytes32 {
+    let mut hasher = Keccak256::new();
+    hasher.update(key);
+    hasher.update(base_slot_bytes);
+    Bytes32::from_array(hasher.finalize().into())
+}
```
antoniolocascio, line 740: Could be shared. `solidity_mapping_slot` could be defined as `solidity_mapping_slot_nested(key, &U256::from(base_slot).to_be_bytes::<32>()`

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -693,3 +702,226 @@ where
...
+/// The Solidity `_handleFinalizeBridgingOnL2Inner` (called with tokenOriginChainId=L1_CHAIN_ID,
+/// fromChainId=L1_CHAIN_ID, tokenAddress=L2_BASE_TOKEN) does:
+/// 1. `_registerLegacyTokenIfNeeded` — for ZKOS chains the base token is registered during upgrade
+/// 2. `_needToForceSetAssetMigrationOnL2` → `_forceSetAssetMigrationNumber`
+/// 3. `chainBalance` update — skipped since base token origin is L1 (not this chain)
+/// 4. `interopInfo[assetId].totalSuccessfulDepositsFromL1 += amount` when settling on L1
+///
+/// IMPORTANT: Must be called BEFORE balance changes, matching the Solidity invariant that
+/// L2AssetTracker is notified before totalSupply/balances change.
+fn handle_finalize_base_token_bridging<S: EthereumLikeTypes>(
```
antoniolocascio, line 757: Can you break this function into smaller ones? Ideally following the solidity structure, to make it easier to follow

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -651,6 +655,11 @@ where
 
     let treasury_address = &system_hooks::addresses_constants::BASE_TOKEN_HOLDER_ADDRESS;
 
+    // Replicate L2AssetTracker.handleFinalizeBaseTokenBridgingOnL2 BEFORE changing balances,
+    // matching the Solidity invariant that the asset tracker is notified before
+    // totalSupply/balances change.
+    handle_finalize_base_token_bridging(system, nominal_token_value, resources)?;
```
antoniolocascio: For Era, do we charge gas from the transaction in the bootloader for these operations?

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -693,3 +702,226 @@ where
...
+        let pre_v31_supply_u256 = pre_v31_supply.into_u256_be();
+
+        let holder_balance = system.io.get_nominal_token_balance(
+            ee,
+            resources,
+            &system_hooks::addresses_constants::BASE_TOKEN_HOLDER_ADDRESS,
+        )?;
+
+        let total_supply =
+            pre_v31_supply_u256 + INITIAL_BASE_TOKEN_HOLDER_BALANCE.saturating_sub(holder_balance);
```
antoniolocascio: Nit: the contract uses checked arithmetic (by default). Not sure how the Era bootloader handles a revert in here. But at least we should document why under and overflows are impossible (if they are)

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -693,3 +702,226 @@ where
...
+        let deposits_slot = Bytes32::from_u256_be(&deposits_slot_u256);
+
+        let current_deposits = system.io.storage_read::<false>(
+            ee,
+            resources,
+            &L2_ASSET_TRACKER_ADDRESS,
+            &deposits_slot,
+        )?;
+        let current_deposits_u256 = current_deposits.into_u256_be();
+        let new_deposits = current_deposits_u256 + *amount;
```
antoniolocascio: Same re. under/overflows

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -693,3 +702,278 @@ where
...
+
+    // Read BASE_TOKEN_ASSET_ID from L2AssetTracker
+    let base_token_asset_id = system.io.storage_read::<false>(
+        ee,
+        resources,
+        &L2_ASSET_TRACKER_ADDRESS,
+        &Bytes32::from_u256_be(&U256::from(ASSET_TRACKER_BASE_TOKEN_ASSET_ID_SLOT)),
+    )?;
+    if base_token_asset_id.is_zero() {
+        // Not initialized yet (before genesis upgrade), skip.
```
AntonD3, line 779: I understand that it just follows solidity logic, but to understand better - in practice it shouldn't be reachable, right?

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -693,3 +702,278 @@ where
...
+    // (tokens are transferred out, never minted into the holder). saturating_sub guards
+    // against any unexpected state where holder_balance exceeds the initial value.
+    //
+    // Addition: pre_v31_supply_u256 is a stored uint256 that in practice fits well within
+    // 128 bits (total ETH supply is ~120M << 2^128), and the subtraction result is at most
+    // 2^127 - 1, so the sum cannot overflow a U256. The Solidity contract uses checked
+    // arithmetic (default in Solidity >=0.8) which would revert on overflow; here we rely
+    // on the same economic bounds making overflow impossible.
+    let total_supply =
+        pre_v31_supply_u256 + INITIAL_BASE_TOKEN_HOLDER_BALANCE.saturating_sub(holder_balance);
```
AntonD3: It's incorrect, `holder_balance` can be bigger than `INITIAL_BASE_TOKEN_HOLDER_BALANCE`, you should probably do similar fix to what I did in https://github.com/matter-labs/era-contracts/pull/2079. Btw makes sense also to check total supply implementation in the hook

AntonD3 (changes requested):

0xVolosnikov:
@vv-dev-ai ptal

vv-dev-ai:
[AI-review] Reviewed commit `fd99c207904fd0d3c9c3a4c5c670d06ccf450632` — **COMMENT** (3 inline comments).

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
vv-dev-ai: [AI-review] **Medium:** `handle_finalize_base_token_bridging` is invoked inside `transfer_from_treasury`, which is called three times per L1 tx (deposit, fee, refund). The docstring says "Must be called BEFORE balance changes" but the fee (line 250) and refund (line 296) calls happen AFTER execution, when balance state has already changed from the deposit transfer.

In Solidity, `handleFinalizeBaseTokenBridgingOnL2` is called exactly once during `mint()` with the full deposit amount. Consider extracting this call to the L1 tx orchestration level, invoked once with `total_deposited` before any `transfer_from_treasury` call.

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -693,3 +702,278 @@ where
...
+        ee,
+        resources,
+        &L2_ASSET_TRACKER_ADDRESS,
+        &deposits_slot,
+    )?;
+    let current_deposits_u256 = current_deposits.into_u256_be();
+    // Addition: both current_deposits and amount represent token quantities bounded by total
+    // token supply (well below 2^128), so the sum cannot overflow a U256. The Solidity
+    // contract uses checked arithmetic (default >=0.8); here the same economic bounds apply.
+    let new_deposits = current_deposits_u256 + *amount;
```
vv-dev-ai: [AI-review] **Medium:** `current_deposits_u256 + *amount` uses wrapping addition. The Solidity counterpart uses checked arithmetic (Solidity >=0.8 default), which would revert on overflow. While economically unlikely, in a state transition function the overflow behavior should match the Solidity contract. Both `current_deposits` (from oracle/storage) and `amount` are untrusted in proving mode.

Suggested fix:
```rust
let new_deposits = current_deposits_u256
    .checked_add(*amount)
    .ok_or(internal_error!("deposits overflow"))?;
```

Same concern applies to the `total_supply` addition at line 877.

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
vv-dev-ai: [AI-review] **Low:** `INITIAL_BASE_TOKEN_HOLDER_BALANCE.saturating_sub(holder_balance)` silently clamps to 0 if `holder_balance` exceeds the initial value. The comment explains why this "can't happen," but if it does (e.g., due to a bug in another part of the system), `saturating_sub` masks the invariant violation and produces an incorrect `totalSupply`. Using `checked_sub` with an `InternalError` would surface the issue:
```rust
let spent = INITIAL_BASE_TOKEN_HOLDER_BALANCE
    .checked_sub(holder_balance)
    .ok_or(internal_error!("holder_balance > INITIAL"))?;
let total_supply = pre_v31_supply_u256
    .checked_add(spent)
    .ok_or(internal_error!("totalSupply overflow"))?;
```

vv-dev-ai (commented):
[AI-review] PR #557 — **COMMENT**

**Medium: `handle_finalize_base_token_bridging` is called once per `transfer_from_treasury` invocation — three times per L1 tx.** `transfer_from_treasury` is called for (1) the deposit transfer to sender, (2) fee payment to operator, and (3) gas refund. Each call triggers `handle_finalize_base_token_bridging`, which performs 3–7 storage reads and up to 2 writes. While the sum of the three `amount` values equals `total_deposited` (matching the Solidity semantics where `handleFinalizeBaseTokenBridgingOnL2` is called once with the full mint amount), this creates:
- **Redundant storage operations** — the 2nd and 3rd calls read `base_token_asset_id` and `assetMigrationNumber` again (warm, but still costs native on the deposit path and adds cache entries)
- **Fragile equivalence** — the Solidity contract is called once with `total_deposited`; here the same total is achieved additively across three calls. If the Solidity contract ever introduces non-additive tracking (e.g., deposit count), the behavior diverges
- **Semantic mismatch** — fee-to-operator and refund transfers are internal accounting, not L1 deposits, yet they trigger deposit tracking

Consider moving the call to `process_l1_transaction_inner` (or the equivalent orchestration point), called once with `total_deposited` before the first `transfer_from_treasury`, matching the Solidity flow more closely.

### Suggested Tests

- **Non-zero totalSupply path**: Set `zkosPreV31TotalSupply > 0` or reduce treasury balance below `INITIAL_BASE_TOKEN_HOLDER_BALANCE` so `totalSupply != 0`, then verify `assetMigrationNumber` is NOT force-set.
- **Accumulation across multiple L1 txs in same block**: Execute two L1 deposit txs in one block. Verify `totalSuccessfulDepositsFromL1` equals the sum of both deposit amounts.
- **Revert path**: Execute an L1 tx that reverts during execution. Verify deposits are still tracked correctly (the fee and refund treasury transfers still happen post-revert, so the function is invoked for those amounts).

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
Raid5594: Yes, it's not reachable in practice, as whether it's genesis upgrade or version upgrade, the asset id will be set

Shall we explicitly revert as a precaution?

0xVolosnikov:
Closed in favor of https://github.com/matter-labs/zksync-os/pull/580

PR #561, opened 2026-03-13
feat: [do not merge, experimental] Usize serialization refactoring

0xVolosnikov:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


forward_system/src/run/query_processors/tx_data.rs
```diff
@@ -88,7 +88,7 @@ impl<TS: TxSource, M: MemorySource> OracleQueryProcessor<M> for TxDataResponder<
                     }
                 } as u32;
 
-                DynUsizeIterator::from_constructor(len, UsizeSerializable::iter)
+                DynWordIterator::from_word_serializable(len)
             }
```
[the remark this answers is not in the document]
0xVolosnikov: [AI] Fixed in commit 3de5a9b9. I added `boxed_inline_word_iter` and switched the small fixed-size forward responses over to it (`u32`, `u8`, `B160`, `()`, small length/hash queries), so those paths no longer materialize an intermediate `Vec<usize>`. Large payloads still use the existing boxed iterator path.

zk_ee/src/oracle/word_serialization/mod.rs
```diff
@@ -0,0 +1,442 @@
...
+}
+
+impl<T: WordDeserializable> WordDeserializable for Vec<T> {
+    fn read_words(src: &mut impl ExactSizeIterator<Item = usize>) -> Result<Self, InternalError> {
+        let len = <u64 as WordDeserializable>::read_words(src)?;
+        let len = usize::try_from(len)
+            .map_err(|_| internal_error!("vec deserialization failed: length overflow"))?;
+        let mut out = Vec::with_capacity(len);
+        for _ in 0..len {
+            out.push(T::read_words(src)?);
```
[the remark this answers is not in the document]
0xVolosnikov: [AI] Fixed in commit 3de5a9b9. I removed the generic `Vec<T>: WordSerializable/WordDeserializable` impls entirely, since they had no production users on this branch and only test coverage. That removes the untrusted length-prefix allocation path instead of trying to harden an API surface we do not need.

0xVolosnikov:
@vv-dev-ai ptal

zk_ee/src/oracle/word_serialization/mod.rs
```diff
@@ -0,0 +1,373 @@
...
+    fn word_len(&self) -> usize {
+        <u64 as WordSerializable>::word_len(&0) * 4
+    }
+
+    fn write_words(&self, out: &mut impl WordSink) {
+        cfg_if::cfg_if!(
+            if #[cfg(target_endian = "big")] {
+                compile_error!("unsupported architecture: big endian arch is not supported")
+            } else if #[cfg(target_pointer_width = "32")] {
+                unsafe {
```
vv-dev-ai, line 220: [AI-review] Missing `// SAFETY:` comment. The soundness reasoning that belongs here: `ruint::Uint<256, 4>` is backed by `[u64; 4]` (32 bytes, no padding), and `[u32; 8]` is also 32 bytes with no invalid bit patterns, so the transmute of a copied value is sound under LE layout. Per the unsafe checklist every `unsafe` block must carry this comment—same pattern was missing in the deleted `usize_serialization/mod.rs` and should not be carried forward.

zk_ee/src/oracle/word_serialization/mod.rs
```diff
@@ -0,0 +1,373 @@
...
+            }
+        );
+    }
+
+    fn write_words(&self, out: &mut impl WordSink) {
+        cfg_if::cfg_if!(
+            if #[cfg(target_endian = "big")] {
+                compile_error!("unsupported architecture: big endian arch is not supported")
+            } else if #[cfg(target_pointer_width = "32")] {
+                unsafe {
```
vv-dev-ai, line 300: [AI-review] Missing `// SAFETY:` comment. `ruint::Uint<160, 3>` is backed by `[u64; 3]` (24 bytes, no padding); `[u32; 6]` is also 24 bytes with no invalid bit patterns. Document this before merge.

zk_ee/src/oracle/word_serialization/mod.rs
```diff
@@ -0,0 +1,373 @@
...
+    unsafe fn init_from_words(
+        this: &mut MaybeUninit<Self>,
+        src: &mut impl ExactSizeIterator<Item = usize>,
+    ) -> Result<(), InternalError> {
+        if src.len() < Self::word_len(&B160::ZERO) {
+            return Err(internal_error!("b160 deserialization failed: too short"));
+        }
+        let mut limbs = [0u64; B160::LIMBS];
+        for limb in &mut limbs {
+            *limb = unsafe { <u64 as WordDeserializable>::read_words(src).unwrap_unchecked() };
```
vv-dev-ai, line 334: [AI-review] Missing `// SAFETY:` comment on `unwrap_unchecked()`. The soundness argument: the `src.len() < Self::word_len(&B160::ZERO)` guard on line 328 ensures at least `LIMBS` words remain; `read_words::<u64>` consumes exactly 1 word on 64-bit and 2 words on 32-bit; the guard is expressed in terms of `word_len` which matches—so `next()` cannot return `None`. This must be stated in a `// SAFETY:` comment. Also note: the outer `unsafe fn` signature does not discharge the obligation for the nested `unsafe { }` block—each site needs its own comment.

zk_ee/src/system/metadata/zk_metadata.rs
```diff
@@ -82,49 +89,26 @@ impl<'de> serde::Deserialize<'de> for BlockHashes {
...
-}
-
 // we only need to know limited set of parameters here,
 // those that define "block", like uniform fee for block,
 // block number, etc
 
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[derive(Clone, Copy, Debug, Default, PartialEq)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, WordSerializable, WordDeserializable)]
 pub struct BlockMetadataFromOracle {
```
vv-dev-ai, line 103: [AI-review] `BlockMetadataFromOracle` is oracle-provided on every block. The PR reorders its field declarations so the derived serialization order matches the former explicit `iter()` order:

```
old struct order: chain_id, block_number, block_hashes, timestamp, eip1559_basefee, …
old iter() order: eip1559_basefee, pubdata_price, native_price, block_number, timestamp, chain_id, …  ← now also the field order
```

I verified that the new declaration order matches the old `iter()` / `from_iter()` order, so the wire format is preserved. However, this is a protocol-critical change: any producer (sequencer oracle, test fixture) that was building this struct by field name and relied on the struct field order (not the serialization order) to produce its byte layout would silently produce wrong data after this change. Add a comment above the struct documenting the canonical wire-format order, and confirm with an integration test that round-trips a real sequencer-produced oracle response.

vv-dev-ai (changes requested):
[AI-review] PR #561 — **REQUEST_CHANGES**

This refactor replaces `UsizeSerializable`/`UsizeDeserializable` (const `USIZE_LEN` + `ExactSizeIterator`) with `WordSerializable`/`WordDeserializable` (instance `word_len()` + push-based `write_words()`), backed by a new proc-macro derive crate and dropping the `ExactSizeChain`/`Either` machinery. The overall design is cleaner and makes variable-length structs possible without workarounds.

Three required fixes before this could merge:

### 1. Missing `// SAFETY:` comments (unsafe checklist)
See inline comments on `word_serialization/mod.rs`. Every `unsafe` block must carry one explaining why it is sound—absence is a required fix per the project unsafe checklist regardless of whether the code is otherwise correct.

### 2. `BlockMetadataFromOracle` field reordering — wire format must be verified
See inline comment on `zk_metadata.rs:98`. I cross-checked the old explicit `iter()` order against the new field declaration order and they match, so the wire format *appears* preserved. But this is a protocol-critical struct (oracle-provided, on every block), and the reorder must be accompanied by an explicit comment documenting the canonical serialization order, plus integration-test evidence (e.g., a round-trip test against a sequencer fixture in the known order).

### 3. `EthereumAccountProperties` old `USIZE_LEN` had a spurious `<bool as UsizeSerializable>::USIZE_LEN` term
The old hand-written `USIZE_LEN` was `u64 + U256 + Bytes32 + Bytes32 + bool` (14 words on 64-bit) but `iter()` only emitted 13 words. The new derive correctly computes 13. This is a silent fix—verify that no oracle provider was padding its response to the (incorrect) 14-word length. If the sequencer was ever written against the old constant rather than the actual iterator count, the new deserialization will misalign.

### Smaller observations
- **PR title**: `feat: [do not merge, experimental] Usize serialization refactoring` — requires `feat(<scope>)` per the semantic-prefix convention in `AGENTS.md`.
- **`boxed_inline_word_iter` panic contract**: If an implementation of `write_words` emits more words than `word_len()` claims, and `word_len() ≤ N`, `ArrayVec::push` will panic (in both debug and release). The `debug_assert_eq!` only catches this in debug builds. Given that inconsistent impls would be bugs anyway this is acceptable, but worth noting in the function's doc comment.
- **No `WordDeserializable` for `[T; N]`**: only `WordSerializable` is provided for arrays. The forced manual `WordDeserializable` impl on `BlockHashes` is the consequence. Consider adding it with a length-validated path to reduce future boilerplate.

### Adversarial analysis
No additional failure paths found beyond those noted above. The `CsrWordSink` length-check pair (`assert!(remaining_len != 0)` per write, `assert!(remaining_len == 0)` after) correctly bounds both under- and over-write relative to the announced length. The `create_quasi_rlp` buffer refactor is correct: `MAX_CREATE_RLP_ENCODING_LEN = 31` covers the worst case, `Take<array::IntoIter<…>>` is `ExactSizeIterator`, and both encoding branches produce the same byte sequences as before.

vv-dev-ai:
[AI-review] Already reviewed at this commit (SHA `175fc63c8c3dce0454bedb357e95633595eb30a0`). Nothing new to report.

PR #578, opened 2026-03-19
feat!: batch prover input run

antoniolocascio:
## What ❔
This PR adds native prover-input generation for a whole batch in ZKsync OS.

* Introduces a batch runner that executes all blocks natively, returns canonical batch prover input and canonical batch pubdata, and derives each next block's ProofData internally. The batch now runs against one mutable BatchState, which is updated after each block with storage writes and published preimages.
* Adds pubdata_used to BlockOutput in zksync-os-interface and plumbs it through forward execution. This field is for sealing/accounting only; canonical pubdata bytes still come from prover-input generation.
* Updates the multiblock tests to compare against legacy approach.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


tests/instances/multiblock_batch/src/lib.rs
```diff
@@ -66,48 +110,137 @@ fn run_multiblock_batch_proof_run(da_commitment_scheme: DACommitmentScheme) {
...
+    });
+    assert!(
+        batch_output.prover_input == legacy_batch_input,
+        "batch prover input mismatch at {:?} (batch len {}, legacy len {}, window {:?})",
+        proof_input_mismatch,
+        batch_output.prover_input.len(),
+        legacy_batch_input.len(),
+        mismatch_window,
+    );
+    assert_eq!(
```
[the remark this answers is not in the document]
antoniolocascio: Added checks

forward_system/src/system/system_types/mod.rs
```diff
@@ -128,6 +131,49 @@ impl<O: IOOracle> BasicSTF for ForwardSystemTypes<O, true> {
...
+    type SystemFunctionsExt = NoStdSystemFunctions<true>;
+    type Allocator = Global;
+    type Logger = Logger;
+    type Metadata = zk_ee::system::metadata::zk_metadata::ZkMetadata;
+}
+
+impl<O: IOOracle> EthereumLikeTypes for BatchProverInputSystemTypes<O> {}
+
+impl<O: IOOracle> BasicSTF for BatchProverInputSystemTypes<O> {
+    type BlockDataKeeper = ZKBasicBlockDataKeeper<NopTxHashesAccumulator>;
```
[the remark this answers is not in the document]
antoniolocascio: False, added comment

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/post_tx_op_sequencing.rs
```diff
@@ -45,6 +45,9 @@ where
         _batch_data: &mut Self::BatchDataKeeper,
         result_keeper: &mut impl ResultKeeperExt<EthereumIOTypesConfig, BlockHeader = Self::BlockHeader>,
     ) -> Result<Self::PostTxLoopOpResult, BootloaderSubsystemError> {
+        result_keeper.record_block_pubdata_used(
```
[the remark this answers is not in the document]
antoniolocascio: Nope

AntonD3 (commented):
Generally changes looks good, but wanted to raise one small question - do you think we should store proof input generation entry points in the forward system? In my opinion behaviorally it's a proof run, so maybe we should move it to the proof system?

antoniolocascio:
@AntonD3 I don't think so. We already discussed it, but I just realized that the old method `generate_proof_input` was already in forward system. I think this makes sense, for me the separation is more about host/guest work. proof_running_system is only the instance of the system for the guest. Otherwise oracle responders should also be considered part of proof_running, as they aren't needed for "forward"/sequencing run

basic_bootloader/src/bootloader/block_flow/zk/batch_data.rs
```diff
@@ -19,6 +21,7 @@ pub struct ZKBatchDataKeeper<A: alloc::alloc::Allocator, O: IOOracle> {
     is_first_block: bool,
     initial_state_commitment: Option<Bytes32>,
     current_state_commitment: Option<Bytes32>,
+    current_proof_data: Option<ProofData<FlatStorageCommitment<TREE_HEIGHT>>>,
```
AntonD3, line 30: nit: maybe add a comment that this data is needed for the next block and we add it here, so it can be reused on caller side. It's just a bit contradicts the current comment

basic_bootloader/src/bootloader/block_flow/zk/batch_data.rs
AntonD3: Maybe you can generally update this comment, as now this strucute used not only for PI, also returns batch data, next block proof data

proof_running_system/src/system/bootloader.rs
AntonD3, line 184: nit: maybe then also move disconnect query for singleblock-batch here for consistency. I think it was here before

forward_system/src/run/mod.rs
```diff
@@ -278,38 +305,133 @@ pub fn generate_batch_proof_input(
...
+        oracle = BatchProverInputBootloader::run_prepared::<BasicBootloaderProvingExecutionConfig>(
+            oracle,
+            &mut batch_data,
+            &mut result_keeper,
+            &mut tracer,
+            &mut validator,
+        )
+        .map_err(|e| wrap_error!(e))?;
+
+        let current_forward_result = std::mem::replace(
```
AntonD3, line 404: nit: took a while to understand how we collect results/pubdata here, comments may be useful

AntonD3 (approved):

basic_bootloader/src/bootloader/block_flow/zk/batch_data.rs
antoniolocascio: Updated it

basic_bootloader/src/bootloader/block_flow/zk/batch_data.rs
antoniolocascio: Added comments

forward_system/src/run/mod.rs
antoniolocascio: Added comments

proof_running_system/src/system/bootloader.rs
antoniolocascio: I'm not sure. I tried, but then we need to add the same query both here and in forward system, which isn't great. Anyways the inconsistency will go away once we get rid of singleblock-batch

AntonD3 (approved):

PR #580, opened 2026-03-19
feat: add a separate call for handleFinalizeBaseTokenBridgingOnL2

Raid5594:
## What ❔

Add a separate call to L2_ASSET_TRACKER during L2 base token mint to correspond in logic to era counterpart.

## Why ❔
Alternative approach involves direct manipulation of respective storage slots, thought it is harder to audit / maintain.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


tests/instances/transactions/src/asset_tracker.rs
```diff
@@ -0,0 +1,227 @@
...
+        0x5f, 0x5f, 0xf3,
+    ]
+}
+
+/// Build a TestingFramework with:
+/// - A mock recording contract at L2_ASSET_TRACKER_ADDRESS
+/// - SystemContext slot 0 set to the given chain ID
+fn setup_with_chain_id(sl_chain_id: u64) -> TestingFramework {
+    TestingFramework::new()
+        .without_revm_consistency_check()
```
0xVolosnikov: Should be removed. I'll add this to AGENTS.md

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -525,6 +529,60 @@ where
...
 
+    // Notify L2AssetTracker about the base token bridging from L1.
+    // This must happen before the treasury transfer so the asset tracker
+    // can update its accounting (e.g. deposits, totalSupply) atomically.
+    // We call handleFinalizeBaseTokenBridgingOnL2(uint256 _fromChainId, uint256 _amount)
+    // as L2_BASE_TOKEN_ADDRESS (0x800a) to pass the onlyBaseTokenHolderOrL2BaseToken modifier.
+    // This call uses infinite resources (not charged to the user) since it is
+    // a protocol-level operation that must always succeed.
+    if total_deposited > U256::ZERO {
+        // Read settlement layer chain ID from SystemContext storage slot 0
```
0xVolosnikov: Separate function would be great here imo

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -525,6 +529,60 @@ where
...
+        let sl_chain_id = {
+            let mut inf_resources = S::Resources::FORMAL_INFINITE;
```
0xVolosnikov: Don't we have some helper function for exactly that somewhere?

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -525,6 +529,60 @@ where
...
+            result: asset_tracker_result,
+            ..
+        } = BasicBootloader::<S, ZkTransactionFlowOnlyEOA<S>>::run_single_interaction(
+            system,
+            system_functions,
+            memories.reborrow(),
+            &calldata,
+            &L2_BASE_TOKEN_ADDRESS,
+            &L2_ASSET_TRACKER_ADDRESS,
+            S::Resources::FORMAL_INFINITE,
```
0xVolosnikov: We'd like to charge native resources here. FORMAL_INFINITE uses infinite gas and infinite native resource

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
antoniolocascio: We do, `read_settlement_layer_chain_id`

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -525,6 +529,60 @@ where
...
+                )
+                .map_err(BootloaderSubsystemError::from)?;
+            U256::from_be_bytes(sl_chain_id_bytes.as_u8_array())
+        };
+
+        // Encode calldata: selector 0x03117c8c + abi-encoded (fromChainId, amount)
+        let mut calldata = [0u8; 68];
+        calldata[0..4].copy_from_slice(&[0x03, 0x11, 0x7c, 0x8c]);
+        calldata[4..36].copy_from_slice(&sl_chain_id.to_be_bytes::<32>());
+        calldata[36..68].copy_from_slice(&total_deposited.to_be_bytes::<32>());
```
antoniolocascio: Can you add the source of the selector in the comment? I.e. `handleFinalizeBaseTokenBridgingOnL2(uint256,uint256)`

antoniolocascio:
@Raid5594 had a go at how it would look like with the suggestions: https://github.com/matter-labs/zksync-os/commit/a16e205fdcce16d7a258d7f38abd00e341fd53fe. Feel free to take a look and cherry pick if you want

0xVolosnikov:
@vv-dev-ai ptal (testing the reviewer)

vv-dev-ai:
[AI-review] Reviewed commit `528521f5886adffffd6cf8836f08d994fd3da851` — **REQUEST_CHANGES** (1 inline comments).

vv-dev-ai:
[AI-review] Reviewed commit `528521f5886adffffd6cf8836f08d994fd3da851` — **REQUEST_CHANGES** (1 inline comment).

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -525,6 +529,60 @@ where
...
+            tracer,
+            validator,
+        )?;
+
+        if asset_tracker_result.failed() {
+            return Err(internal_error!(
+                "L2AssetTracker.handleFinalizeBaseTokenBridgingOnL2 failed"
+            )
+            .into());
+        }
```
vv-dev-ai: [AI-review] **High:** This `internal_error!` introduces a new hard failure path in L1 transaction processing. When the L2AssetTracker contract reverts (bug, upgrade mismatch, unexpected state), this error propagates through line 208 of `process_l1_transaction` as a non-recoverable `InternalError` that halts block processing.

For L1→2 transactions, block halt = **priority queue stall** (L1 txs cannot be skipped). The existing L1 tx flow carefully avoids hard failures via `L1ResourcesPolicy` saturation, and `FatalRuntimeError` is caught at lines 195–206 and converted to a graceful revert. This `InternalError` bypasses that safety net.

While the contract is trusted and the call uses `FORMAL_INFINITE` resources (preventing resource exhaustion), Solidity-level reverts (assertion failures, modifier rejections, storage inconsistency) would still trigger this path.

Consider:
- Converting failure to a graceful revert (like the `FatalRuntimeError` handler at lines 195–206): exhaust ergs, roll back the frame, continue processing the L1 tx as a reverted execution. This preserves the deposit flow even if the notification fails.
- Or document this as an intentional liveness assumption with a clear recovery path (e.g., emergency contract upgrade procedure).

vv-dev-ai (changes requested):
[AI-review] PR #580 — **REQUEST_CHANGES**

**Protocol change**: This introduces new oracle queries (storage read on SystemContext slot 0, full EVM execution of asset tracker call) in the L1 tx execution path for deposits. This changes the witness tape layout in proving mode and requires a coordinated prover/sequencer upgrade.

### Suggested Tests

- **Asset tracker call failure**: Deploy a reverting contract at `L2_ASSET_TRACKER_ADDRESS` and send an L1 tx with `total_deposited > 0`. This will confirm whether the current `internal_error!` path actually halts the block (and document the expected behavior).
- **Upgrade tx (0x7e) with deposit**: Verify whether `total_deposited > 0` is possible for upgrade transactions and that the asset tracker call path handles it correctly.
- **Zero/uninitialized chain ID**: Test with SystemContext slot 0 uninitialized (zero). Verify the asset tracker receives `chainId = 0` and the call succeeds.

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
vv-dev-ai: [AI-review] **High:** This `InternalError` halts block processing. For L1 priority transactions, this stalls the priority queue — the most critical liveness threat. L1→L2 deposits are sequentially ordered from L1; if one can't be processed, all subsequent deposits are blocked.

The L1 tx path is specifically designed for maximum resilience: `L1ResourcesPolicy` saturates arithmetic errors, signature/nonce checks are skipped, `FORMAL_INFINITE` resources prevent OOG. This new dependency on an L2 contract call succeeding is a fundamentally weaker guarantee than the existing `InternalError` paths at lines 526/530 (which guard against arithmetic conditions the L1 contract already prevents).

If the L2AssetTracker is ever upgraded with a bug or reverts on unexpected input, this halts the chain.

Consider:
1. Treating asset tracker failure as a soft revert — the L1 tx reverts, deposit is refunded to `refund_recipient`, priority queue advances. This matches how main tx body failures are handled (line 644–648: `reverted = result.failed()` → frame rolled back, tx continues as `Revert`).
2. Or explicitly document that asset tracker availability is a liveness precondition for the L1 bridge, and add a test verifying behavior when the contract reverts.

vv-dev-ai (changes requested):
[AI-review] PR #580 — **REQUEST_CHANGES**

**High:** The `InternalError` at `process_l1_transaction.rs:579` introduces a new priority queue stall vector. If the L2AssetTracker contract at `0x1000f` ever reverts, block processing halts and all subsequent L1 deposits are blocked. See inline comment.

### Suggested Tests
- **Reverting asset tracker**: Deploy a reverting mock at `L2_ASSET_TRACKER_ADDRESS` and send an L1 deposit tx. Verify the system handles the failure gracefully (current behavior: `InternalError` → block halt).
- **Reverted L1 tx atomicity**: Send an L1 deposit where the main tx body reverts. Verify the asset tracker notification is also rolled back (frame at line 514 rollback should undo both).
- **Simulation mode**: Run an L1 deposit through `eth_call` / gas estimation and verify the asset tracker call doesn't interfere.

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
antoniolocascio: I think this is the expected behaviour on Era rn, so I think it's ok to keep it

0xVolosnikov: It actually has more important issues. We have to do this call 3 times in practice, otherwise AssetTracker can get invalid data if tx actually reverts for some reason

antoniolocascio: As I said on the other thread: I think this is the expected behaviour on Era rn, so I think it's ok to keep it

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -309,26 +306,53 @@ where
...
+    // so we use an empty slice. L1 transaction returndata is only
+    // informational (it appears in receipts) and is not consumed by
+    // any on-chain logic.
+    let result = if is_success {
+        ExecutionResult::Success {
+            output: ExecutionOutput::Call(&[]),
+        }
+    } else {
+        ExecutionResult::Revert { output: &[] }
+    };
```
antoniolocascio: This seems weird, can't we save the returndata from the actual call? It might have a useful revert reason or even some data we expose to block explorer

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -240,7 +223,13 @@ where
     let pay_to_operator = U256::from(gas_used)
         .checked_mul(U256::from(gas_price))
         .ok_or(internal_error!("gu*gp"))?;
+    // Use FORMAL_INFINITE for post-execution operations (coinbase transfer,
+    // asset tracker notifications, refund transfer, log emission).
+    // These cannot fail due to resource exhaustion. We measure the native
+    // consumed via the diff on inf_resources and include it in
+    // computational_native_used so that block seal criteria remain accurate.
     let mut inf_resources = S::Resources::FORMAL_INFINITE;
+    let inf_resources_initial = inf_resources.clone();
```
antoniolocascio: I don't think we should do this, we never account for native spent that cannot be charged for. As a rule: native_used = native for pubdata + native used for computation. This change would break this.
Instead, the computation that happens after we commit to the final fee should be accounted for as intrinsic and done with FORMAL_INFINITE.

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -582,27 +622,35 @@ where
...
+                system_log!(
+                    system,
+                    "L1 transaction ran out of native resources or memory {re:?}\n"
+                );
+                resources.exhaust_ergs();
+                system.finish_global_frame(Some(&rollback_handle))?;
+                true
+            }
+            _ => return Err(e),
+        },
```
antoniolocascio: Why move this to inside this function? Probably it's only a formal concern, but I don't like that there's fallible computation outisde of this handling. For example, if we run out of native on the mint above or the new call to asset tracker, that's not going to be handled as a revert, which will make the chain stuck

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -693,3 +742,76 @@ where
...
+///
+/// This is called separately for each token movement (value mint, operator
+/// payment, refund) so that the asset tracker's accounting stays correct even
+/// if the main transaction body reverts.
+///
+/// Uses infinite EVM gas but tracks native resource consumption via the
+/// caller's resource object, so the block seal criteria remain accurate.
+///
+/// Failure is logged but does not halt block processing — L1 transactions
+/// must always be processable to avoid stalling the priority queue.
```
antoniolocascio: Not saying this is wrong per-se, but why is it ok to continue processing in this case? IIUC, Era halts if this ever happens. In general, safety comes before liveness. If it's unsafe to continue after an asset tracker failure, then we should halt

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -44,8 +44,19 @@ pub const L1_TX_INTRINSIC_NATIVE_COST: u64 = 130_000;
...
+// settled on L1), the contract performs a single SSTORE:
+//   interopInfo[assetId].totalSuccessfulDepositsFromL1 += _amount
+// Each storage diff is encoded as 32 bytes (derived key) + compressed value
+// diff. The worst-case compressed value using the Add strategy with a
+// 256-bit amount falls back to Nothing encoding = 33 bytes.
+const ASSET_TRACKER_INTRINSIC_PUBDATA: u64 = 32 + 33;
+
+// Needed to publish the l1 tx log, coinbase balance, and asset tracker state diff.
+pub const L1_TX_INTRINSIC_PUBDATA: u64 =
+    88 + COINBASE_BALANCE_INTRINSIC_PUBDATA + ASSET_TRACKER_INTRINSIC_PUBDATA;
```
antoniolocascio: @AntonD3 IIRC these values need to match those in L1, right? So we cannot change them as we please, no?

tests/instances/transactions/src/asset_tracker.rs
```diff
@@ -0,0 +1,231 @@
...
+/// // Accumulate amount: slot[1] += CALLDATALOAD(36)
+/// PUSH1 1, SLOAD, PUSH1 36, CALLDATALOAD, ADD, PUSH1 1, SSTORE
+/// // Store chain ID: slot[2] = CALLDATALOAD(4)
+/// PUSH1 4, CALLDATALOAD, PUSH1 2, SSTORE
+/// // Store last amount: slot[3] = CALLDATALOAD(36)
+/// PUSH1 36, CALLDATALOAD, PUSH1 3, SSTORE
+/// // Return success
+/// PUSH0, PUSH0, RETURN
+/// ```
+fn mock_accumulating_contract() -> Vec<u8> {
```
0xVolosnikov, line 44: Tell agent to use existing EvmBytecodeBuilder or smth like that

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -150,68 +153,48 @@ where
...
-            tracer,
-            validator,
-        ) {
-            Ok((r, pubdata_used, to_charge_for_pubdata, resources_before_refund)) => {
-                let pubdata_info = match r {
-                    ExecutionResult::Success { .. } => {
+    let (is_success, pubdata_info, resources_before_refund, mut memories) =
+        if !preparation_out_of_resources {
+            // Take a snapshot in case we need to revert due to out of native.
+            let rollback_handle = system.start_global_frame()?;
```
0xVolosnikov, line 163: Need to carefully check frames and behavior in case of reverts

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
Raid5594: I agree, we do not want to allow failed bookkeeping in the L2AssetTracker, so halting seems reasonable

basic_bootloader/src/bootloader/constants.rs
Raid5594: We'll update L1 constants respectively for zksync-os.

AntonD3: Yes, we should update l1 contracts, but at the same time make sure that l1 tx can be included even if not enough gas provided to cover the intrinsic cost

vv-dev-ai:
[AI-review] Reviewed commit `50142a771757f9d4699eb0a5cd53082e451b6ed4` — **REQUEST_CHANGES** (1 inline comment).

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -567,42 +617,51 @@ where
...
+            validator,
+        ) {
+            Ok(CompletedExecution {
+                resources_returned,
+                result,
+            }) => {
+                let reverted = result.failed();
+                // Save the returndata before asset-tracker calls overwrite
+                // the runner memory buffer.
+                let returndata = result.return_values().returndata.to_vec();
```
vv-dev-ai: [AI-review] **High:** `.to_vec()` allocates via the global allocator. In production proving builds (`dump_bin.sh --type singleblock-batch`, features `proving,production` — no `global-alloc`), `OptionalGlobalAllocator::alloc` panics with `"global alloc not allowed"` for any non-empty returndata slice.

Trigger: Any L1 deposit transaction whose target contract returns non-empty data.
Impact: Proving-mode abort — the prover cannot produce a proof for a block containing such a transaction. Violates review rule: "Code in crates shared between forward and proving mode must not use global-allocator APIs unless gated behind `#[cfg(not(target_arch = \"riscv32\"))]` or the `global-alloc` feature."
Fix: Save returndata into a pre-sized region in `RunnerMemoryBuffers` (or another allocator-aware buffer) rather than using `.to_vec()`. Alternatively, if the returndata is not needed for L1 tx receipts, skip saving it and use `&[]` directly.

vv-dev-ai (changes requested):
[AI-review] PR #580 — **REQUEST_CHANGES**

✅ RESOLVED: `internalerror-from-l2assettracker-call-failure` — Retracted. Two org members (antoniolocascio, 0xVolosnikov) confirmed the halt-on-failure behavior matches Era and is the correct design choice: safety (correct token accounting) takes priority over liveness. The updated code includes explicit documentation for this decision.

**New finding (see inline):** The `.to_vec()` global allocation at line 641 can abort production proving builds.

0xVolosnikov:
Note that the point of REVM consistency checker tests is to guarantee that this component will work correctly in zksync-os-server - so ideally it should handle this new functionality out of the box

0xVolosnikov:
@vv-dev-ai ptal the latest commits

vv-dev-ai:
[AI-review] Reviewed commit `2b81c4a24438c7b7de3e246704bb9a65d8d0c919` — **APPROVE** (0 inline comments).

vv-dev-ai (commented):
[AI-review] PR #580 — **APPROVE**

✅ **RESOLVED** `global-alloc-to-vec-proving-panic`: The `.to_vec()` global allocator call has been replaced with `Vec::with_capacity_in(rd.len(), system.get_allocator())` + `extend_from_slice()`. This uses the system allocator (talc in proving mode) with exact pre-sizing — no reallocation possible. Safe in production proving builds.

Verified:
- Frame management: all paths through `execute_l1_transaction_and_notify_result` close the frame exactly once via the `frame_finished` guard. The closure pattern correctly catches both FatalRuntimeError (rollback + continue) and other errors (rollback + propagate).
- Value conservation (I2): `to_transfer + pay_to_operator + to_refund_recipient = total_deposited` on success. On revert, value mint rolls back with the inner frame; `pay_to_operator + to_refund_recipient = total_deposited`.
- Resource flow: value mint notification correctly uses user's resources via `with_infinite_ergs` (native tracked, ergs infinite) since it's inside the execution frame. Post-execution notifications (operator fee, refund) correctly use `FORMAL_INFINITE`, covered by the increased `L1_TX_INTRINSIC_NATIVE_COST`.
- `ASSET_TRACKER_INTRINSIC_PUBDATA` (65 bytes) covers the single storage diff from 3 calls to the same mapping slot.
- REVM consistency checker replay logic mirrors the bootloader: value mint gated on `is_success && to_transfer > 0`, operator/refund gated on amount > 0, amounts match.

### Suggested Tests
- **Revert path with nonzero deposit**: An L1 tx where the main body reverts (e.g., invalid callee) but `total_deposited > 0`. Assert that: (a) accumulated amount in the asset tracker equals `total_deposited` (operator fee + refund, but NOT value mint since it rolled back), and (b) call count is 2 (not 3).
- **OON during tx body**: An L1 tx with a gas limit that causes OON during execution. Verify the asset tracker is still called for operator fee and refund portions after the revert.

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
antoniolocascio: Is this resolved? @Raid5594

tests/revm_runner/src/storage_diff_comp.rs
```diff
@@ -250,7 +258,9 @@ where
     let mut map = HashMap::new();
     let account_properties_storage_address = ACCOUNT_PROPERTIES_STORAGE_ADDRESS.into_alloy();
     for w in zksync_storage_writes {
-        if w.account == account_properties_storage_address {
+        if w.account == account_properties_storage_address
+            || should_skip_consistency_check_for_address(w.account)
```
0xVolosnikov: Should we remove it?

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -309,16 +317,25 @@ where
         })?;
     }
 
+    // Notify asset tracker about the refund portion of the deposit.
+    notify_l2_asset_tracker::<S>(
```
0xVolosnikov: Would be nice to move it into `if to_refund_recipient > U256::ZERO {` block - can save some cycles for free

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -483,152 +536,182 @@ fn execute_l1_transaction_and_notify_result<
...
-        false,
-        tracer,
-        validator,
-    )?;
-    let reverted = result.failed();
-    let return_values = result.return_values();
+        // Notify L2AssetTracker about the value mint portion of the deposit.
+        // This is inside the execution frame, so it gets rolled back if the
+        // main tx body reverts — matching the treasury transfer above.
+        notify_l2_asset_tracker::<S>(
```
0xVolosnikov: Same here, we have `if to_transfer > U256::ZERO || Config::SIMULATION {` block above

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -483,152 +536,182 @@ fn execute_l1_transaction_and_notify_result<
...
+            system,
+            system_functions,
+            memories.reborrow(),
+            to_transfer,
+            resources,
```
0xVolosnikov: We use actual resources for the tx here, don't we? So it will spend actual gas and native, so it may be triggered by some gas griefing attack in theory, isn't it?

tests/revm_runner/Cargo.toml
```diff
@@ -24,7 +24,7 @@ revm = "=34.0.0"
 revm-inspectors = "0.34.0"
 log = { workspace = true }
 
-zksync-os-revm = { git = "https://github.com/matter-labs/zksync-os-revm", rev="237fd5a" } # Current dev branch is vv-new-version
+zksync-os-revm = { git = "https://github.com/matter-labs/zksync-os-revm", rev = "466dcfe831d081577563b8e28c715e6127e460c7" }
```
0xVolosnikov: Comment to not forget to update before merge

vv-dev-ai:
[AI-review] Reviewed commit `f83dcbbbc839dc7db30e7785084abf350b14b589` — **COMMENT** (1 inline comments).

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
vv-dev-ai: [AI-review] **Medium:** Value-mint notification consumes native from user’s resources, unlike operator-fee and refund notifications which use `FORMAL_INFINITE`.

Trigger: L1 tx with nonzero deposit + tight gas limit + cold asset tracker storage (first L1 tx in block).
Impact: (1) Reduces native available for the main tx body. (2) If native exhaustion occurs here, the `FatalRuntimeError` is not caught by the handler at line 659 — it propagates up and halts block processing (priority queue stall). (3) Doc comment on `notify_l2_asset_tracker` (line 791) incorrectly states “Uses FORMAL_INFINITE.”
Fix: Pass `FORMAL_INFINITE` resources here (matching operator-fee/refund pattern) and include the value-mint notification’s native cost in `L1_TX_INTRINSIC_NATIVE_COST`. Alternatively, wrap this call in a `FatalRuntimeError` catch that converts to a revert (matching the main tx body pattern at line 659).

vv-dev-ai (commented):
[AI-review] PR #580 — **COMMENT**

Reply-review evaluating new org-member feedback on the latest code.

### Value-mint notification uses user’s native budget

Validating @0xVolosnikov’s [concern](https://github.com/matter-labs/zksync-os/pull/580#discussion_r3009714509): the value-mint `notify_l2_asset_tracker` call at line 608 passes `resources` (user’s mutable resources) to the function. Inside, `resources.with_infinite_ergs(|inf_ergs| { ... })` makes ergs infinite but **native is still consumed from the user’s budget**. This differs from the operator-fee (line 277) and refund (line 342) notifications, which pass `&mut inf_resources` where `inf_resources = S::Resources::FORMAL_INFINITE` — truly infinite native.

Consequences:
- Native consumed by the value-mint notification **reduces the user’s budget** for the main tx body (line 618: `resources_for_tx = resources.clone()`).
- This native cost is **not covered by `L1_TX_INTRINSIC_NATIVE_COST`** (which only accounts for post-execution operations — see the constant’s comment).
- If the notification exhausts native (cold asset tracker storage on the first L1 tx in a block), the resulting `FatalRuntimeError` is **not caught** by the error handler that converts main-tx-body OON to a graceful revert (that handler is at line 659). Instead it propagates through the closure and up to `process_l1_transaction`, halting block processing.
- The function’s doc comment (line 791) incorrectly states “Uses FORMAL_INFINITE”.

See inline comment for the specific line.

### Minor: guard notifications behind zero-amount checks at call site

Per @0xVolosnikov’s suggestions: `notify_l2_asset_tracker` internally checks `if amount > U256::ZERO`, so calling it with zero amounts is a no-op. But in a ZK context, even the function call overhead matters. Moving the refund notification inside the existing `if to_refund_recipient > U256::ZERO {` block (line 335) and the value-mint notification inside `if to_transfer > U256::ZERO || Config::SIMULATION {` (line 586) would save cycles for free.

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -695,3 +778,81 @@ where
...
+/// if the main transaction body reverts.
+///
+/// Uses FORMAL_INFINITE because this bookkeeping is intrinsic to L1 tx
+/// processing. Its native cost is prepaid via L1_TX_INTRINSIC_NATIVE_COST
+/// and is not included in computational_native_used.
+///
+/// Failure halts block processing — if the asset tracker reverts, the
+/// chain's token accounting would be inconsistent, so we treat it as
+/// fatal rather than silently continuing with incorrect bookkeeping.
+fn notify_l2_asset_tracker<'a, S: EthereumLikeTypes + 'a>(
```
0xVolosnikov, line 825: To note: if nothing is deployed on L2AssetTracker it will silently succeed. Maybe makes sense to add comment that this behavior is expected (for auditors)

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -31,21 +31,41 @@ pub const MAX_BLOCK_GAS_LIMIT: u64 = u64::MAX / ERGS_PER_GAS;
...
+//  - coinbase transfer from the older accounting model
+//
+// The current L1 path adds post-execution intrinsic work:
+//  - operator-fee L2AssetTracker notification: warm path, 25k
+//  - refund transfer: treasury warm write 5k + refund recipient cold new write ~205k
+//  - refund L2AssetTracker notification: warm path, 25k
+//
+// This gives a worst-case incremental cost of ~260k, so we set:
+//   130k + 260k = 390k
+pub const L1_TX_INTRINSIC_NATIVE_COST: u64 = 390_000;
```
0xVolosnikov: I think we should explain why we charge the warm path and not the cold path here. It may confuse the auditors a bit

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -31,21 +31,41 @@ pub const MAX_BLOCK_GAS_LIMIT: u64 = u64::MAX / ERGS_PER_GAS;
...
-pub const L1_TX_INTRINSIC_PUBDATA: u64 = 88 + COINBASE_BALANCE_INTRINSIC_PUBDATA;
+// Pubdata produced by the L2AssetTracker.handleFinalizeBaseTokenBridgingOnL2
+// call that the bootloader makes inside the L1 tx execution frame (value-mint
+// notification). In the steady-state case (base token already registered,
+// settled on L1), the contract performs a single SSTORE:
+//   interopInfo[assetId].totalSuccessfulDepositsFromL1 += _amount
+// Each storage diff is encoded as 32 bytes (derived key) + compressed value
+// diff. The worst-case compressed value using the Add strategy with a
+// 256-bit amount falls back to Nothing encoding = 33 bytes.
+const ASSET_TRACKER_INTRINSIC_PUBDATA: u64 = 32 + 33;
```
0xVolosnikov, line 108: value-mint notification will be reverted if the tx is reverted, so probably we will overcharge for the pubdata in this case. Maybe this is ok

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -695,3 +778,81 @@ where
...
+    validator: &mut impl TxValidator<S>,
+) -> Result<(), BootloaderSubsystemError>
+where
+    S::IO: IOSubsystemExt,
+    S::Metadata: ZkSpecificPricingMetadata
+        + BasicMetadata<S::IOTypes, TransactionMetadata = TxLevelMetadata<S::IOTypes>>,
+{
+    if amount > U256::ZERO {
+        // Read settlement layer chain ID from SystemContext storage slot 0
+        let sl_chain_id = read_settlement_layer_chain_id(&mut system.io);
```
0xVolosnikov: nit: we read it every time

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -31,21 +31,41 @@ pub const MAX_BLOCK_GAS_LIMIT: u64 = u64::MAX / ERGS_PER_GAS;
...
-//  - Transferring fee to coinbase.
-//  - Hashing of tx hash into rolling hash.
-//  - Adding tx hash into l1 tx linear hasher
-pub const L1_TX_INTRINSIC_NATIVE_COST: u64 = 130_000;
+// Covers intrinsic L1 tx work not charged as tx-body computation.
+//
+// Baseline 130k covers:
+//  - storing and hashing the L1 tx log
+//  - hashing tx hash into the rolling hash and linear hasher
+//  - coinbase transfer from the older accounting model
```
AntonD3: Shouldn't it also be repriced same way as refund? At least now it also touches warm treasury address, before it wasn't taken in to account

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -31,21 +31,41 @@ pub const MAX_BLOCK_GAS_LIMIT: u64 = u64::MAX / ERGS_PER_GAS;
...
+// settled on L1), the contract performs a single SSTORE:
+//   interopInfo[assetId].totalSuccessfulDepositsFromL1 += _amount
+// Each storage diff is encoded as 32 bytes (derived key) + compressed value
+// diff. The worst-case compressed value using the Add strategy with a
+// 256-bit amount falls back to Nothing encoding = 33 bytes.
+const ASSET_TRACKER_INTRINSIC_PUBDATA: u64 = 32 + 33;
+
+// Needed to publish the l1 tx log, coinbase balance, and asset tracker state diff.
+pub const L1_TX_INTRINSIC_PUBDATA: u64 =
+    88 + COINBASE_BALANCE_INTRINSIC_PUBDATA + ASSET_TRACKER_INTRINSIC_PUBDATA;
```
AntonD3: Should we take into account treasury balance change as well here? And maybe also refund?

tests/revm_runner/Cargo.toml
0xVolosnikov: (makes sense to update now as well - zksync-os-revm was rebased)

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
0xVolosnikov: @Raid5594 pls double check that out-of-native won't break the block processing

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
0xVolosnikov: @Raid5594 it was discussed, but pls also answer here

basic_bootloader/src/bootloader/constants.rs
Raid5594: What was the assumption the refund hasn't been considered here before?

0xVolosnikov: > assumption

Probably we just missed it

AntonD3: Yeah, refund was not present before, I just remember we discussed during some call, and you mentioned that you can now account everything, though I don't mind fixing in in a follow up PR. Btw, for refund it makes sense to check if it's needed, maybe it's already accounted somewhere else

0xVolosnikov:
Also it makes sense to ask AI to find more edge cases to cover with tests. And generally improve the test coverage. I think it's fine to do as a follow-up

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
Raid5594: Resolved in latest commits, now handled at the end of the function

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -695,3 +802,85 @@ where
...
+/// Notify L2AssetTracker about base token bridging from L1.
+///
+/// Calls handleFinalizeBaseTokenBridgingOnL2(uint256 _fromChainId, uint256 _amount)
+/// as L2_BASE_TOKEN_ADDRESS (0x800a) to pass the onlyBaseTokenHolderOrL2BaseToken modifier.
+///
+/// This is called separately for each token movement (value mint, operator
+/// payment, refund) so that the asset tracker's accounting stays correct even
+/// if the main transaction body reverts.
+///
+/// Uses FORMAL_INFINITE because this bookkeeping is intrinsic to L1 tx
```
AntonD3: nit: it's not true, resources passed from outside, and it's not always infinite

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -695,3 +802,85 @@ where
...
+        })?;
+
+        if failed {
+            system_log!(
+                system,
+                "L2AssetTracker.handleFinalizeBaseTokenBridgingOnL2 failed for amount {amount:?}\n"
+            );
+            // A revert here means the chain's token accounting would be inconsistent.
+            // Treated as a fatal system error — block processing cannot continue.
+            return Err(internal_error!(
```
AntonD3, line 878: But should we treat it as fatal error for value mint? At least in current implementation I think out of native is possible for it

basic_bootloader/src/bootloader/constants.rs
AntonD3: Asset tracker totalSuccessfulDepositsFromL1 will be always modified for the transaction fee deposit

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -31,21 +31,62 @@ pub const MAX_BLOCK_GAS_LIMIT: u64 = u64::MAX / ERGS_PER_GAS;
...
-pub const L1_TX_INTRINSIC_PUBDATA: u64 = 88 + COINBASE_BALANCE_INTRINSIC_PUBDATA;
+// Pubdata needed for the treasury balance diff caused by transfers
+// from treasury. Use the same worst-case balance-diff estimate as
+// for coinbase balance updates.
+const TREASURY_BALANCE_INTRINSIC_PUBDATA: u64 = 32 + 34;
+
+// Pubdata needed for the refund recipient balance diff in the worst case.
+// As with the coinbase/treasury balance updates, price a 32-byte key and
+// 34-byte uncompressed value update.
+const REFUND_RECIPIENT_BALANCE_INTRINSIC_PUBDATA: u64 = 32 + 34;
```
AntonD3, line 98: Coinbase, treasury, refund - can we just define single "BALANCE_DIFF_WORST_CASE_PUBDATA"?

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -31,21 +31,62 @@ pub const MAX_BLOCK_GAS_LIMIT: u64 = u64::MAX / ERGS_PER_GAS;
...
+//  - hashing tx hash into the rolling hash and linear hasher
+//  - coinbase transfer from the older accounting model
+//
+// The current L1 path adds post-execution intrinsic work:
+//  - operator-fee L2AssetTracker notification: warm path, 25k
+//  - refund transfer: treasury warm write 5k + refund recipient cold new write ~205k
+//  - refund L2AssetTracker notification: warm path, 25k
+//  - coinbase transfer: treasury warm write 5k
+//
+// We use the warm-path cost for asset tracker notifications because
```
AntonD3: Actually in the worst case it can be cold access, if resources passed to execution will not cover balance increase(and L2AssetTracker call)

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -31,21 +31,62 @@ pub const MAX_BLOCK_GAS_LIMIT: u64 = u64::MAX / ERGS_PER_GAS;
...
-//  - Adding tx hash into l1 tx linear hasher
-pub const L1_TX_INTRINSIC_NATIVE_COST: u64 = 130_000;
+// Covers intrinsic L1 tx work not charged as tx-body computation.
+//
+// Baseline 130k covers:
+//  - storing and hashing the L1 tx log
+//  - hashing tx hash into the rolling hash and linear hasher
+//  - coinbase transfer from the older accounting model
+//
+// The current L1 path adds post-execution intrinsic work:
```
AntonD3: nit: IMHO such "baseline" +  "current L1 path" separation highlights history of this variable. It's useful for reviewers of this PR, but IMHO may be confusing for future readers of this code. WDYT? Not a hard opinion, but maybe makes sense to just list all the cost components in this comment?

AntonD3 (changes requested):

basic_bootloader/src/bootloader/constants.rs
AntonD3: I think it's actually the opposite, if execution is **not** reverted - then we are overcharging, as the user pays, at least partially, for this diff as part of the execution, then we are charging worst-case for the same slot. We can actually refund `ASSET_TRACKER_INTRINSIC_PUBDATA` - `"interopInfo[assetId].totalSuccessfulDepositsFromL1" storage slot actually spent during execute`, but probably not a big deal

basic_bootloader/src/bootloader/constants.rs
Raid5594: `Asset tracker totalSuccessfulDepositsFromL1 always will be always modified for transaction fee deposit`

answering above. this is true only if chain settles on L1

basic_bootloader/src/bootloader/constants.rs
Raid5594: This is related to discussion we had yesterday. do we assume that L1->L2 can ever fail due to out of gas for pubdata for the mint part

basic_bootloader/src/bootloader/constants.rs
Raid5594: We can recalculate above then to make it precise and list everything (I didn't look into assumptions of 130k)

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
antoniolocascio: The failed comes from asset_tracker_result.failed(), which is true when the result is Ok(revert). Out of native would be Err(), so handled by ? and then captured and converted to a top-level revert.

vv-dev-ai:
[AI-review] Reviewed commit `55b30bff66e32c66e9b0d4992b2c68b178d38a50` — **APPROVE** (1 inline comments).

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
vv-dev-ai: [AI-review] **Nit:** Doc comment says "Uses FORMAL_INFINITE" but the value-mint call site (line 643) passes `resources.with_infinite_ergs(...)`, which tracks native against user resources. Only the operator-fee and refund call sites use `FORMAL_INFINITE`. Consider updating to reflect the actual calling convention (e.g., "Resource usage depends on the caller — value-mint tracks native against user resources; operator-fee and refund use FORMAL_INFINITE").

vv-dev-ai (commented):
[AI-review] PR #580 — **APPROVE** (posted as COMMENT; GitHub disallows self-review)

Reply-review evaluating prior open issues against new commits and author/reviewer responses.

### Prior issue: `value-mint-uses-user-native` — ✅ RESOLVED

The critical sub-concern (out-of-native from the value-mint `notify_l2_asset_tracker` call halting block processing) is fixed. The new commits restructure error handling: `FatalRuntimeError` from `execute_l1_transaction_and_notify_result` is now caught in the caller (`process_l1_transaction`, lines 199–220) and converted to a graceful top-level revert. The priority queue advances normally.

Verified in code: the `with_infinite_ergs` wrapper at line 643 ensures OON propagates as `FatalRuntimeError`, the `?` at line 667 carries it out, and the caller's match at line 200 catches it. Both inner and outer frames are properly rolled back.

**Remaining nit:** The doc comment on `notify_l2_asset_tracker` (line 815) still says "Uses FORMAL_INFINITE" — this is only true for the operator-fee and refund call sites. The value-mint call site uses `resources.with_infinite_ergs(...)`, which tracks native against user resources. AntonD3 confirmed this inaccuracy.

basic_bootloader/src/bootloader/constants.rs
AntonD3: Why not? It can even fail with out of native for mint part

basic_bootloader/src/bootloader/constants.rs
AntonD3: Why? Shouldn't we deposit fee for chains that are settling on top of GW?

Anyway, are we ok with such undercharge, even if it's only affects chains that are settling on l1?

AntonD3: Btw, I think it is answer to this https://github.com/matter-labs/zksync-os/pull/580/changes#r3009873615, right?

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
AntonD3: Ok, I see, makes sense. Not important, but comment about it would be appreciated

basic_bootloader/src/bootloader/constants.rs
AntonD3: It would be nice IMHO

vv-dev-ai:
[AI-review] Reviewed commit `55b30bff66e32c66e9b0d4992b2c68b178d38a50` — **COMMENT** (1 inline comments).

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -31,21 +31,62 @@ pub const MAX_BLOCK_GAS_LIMIT: u64 = u64::MAX / ERGS_PER_GAS;
...
+//  - operator-fee L2AssetTracker notification: warm path, 25k
+//  - refund transfer: treasury warm write 5k + refund recipient cold new write ~205k
+//  - refund L2AssetTracker notification: warm path, 25k
+//  - coinbase transfer: treasury warm write 5k
+//
+// We use the warm-path cost for asset tracker notifications because
+// L2AssetTracker is always accessed earlier in the same tx (value-mint
+// notification), so subsequent accesses hit the warm cache. Using the cold
+// path cost would overcharge and unnecessarily reduce the gas available
+// for user execution.
```
vv-dev-ai: [AI-review] **Low:** The warm-path assumption here does not hold when the L1 tx reverts.

Trigger: L1 tx body reverts → frame rollback clears `last_touched_in_tx` for the L2AssetTracker slot (confirmed by `generic_pubdata_aware_plain_storage.rs:180`: "in case of revert it should become cold again") → post-execution operator-fee and refund notifications hit cold storage.

Impact: `L1_TX_INTRINSIC_NATIVE_COST` underestimates the native cost on the revert path. Not a correctness issue (FORMAL_INFINITE prevents failures), but operator undercharging on reverted L1 txs.

Fix: Either (a) use cold-path cost for worst-case estimate, or (b) document the undercharge as acceptable and explain the trade-off (overcharging on success path vs undercharging on revert path).

vv-dev-ai (commented):
[AI-review] PR #580 — **COMMENT** (posted as COMMENT; GitHub disallows self-review)

Reply-review: evaluating org member discussion since prior APPROVE.

**AntonD3's cold-path observation is valid.** The `L1_TX_INTRINSIC_NATIVE_COST` comment (lines 47–51) assumes asset tracker notifications use warm-path cost (25k) because "L2AssetTracker is always accessed earlier in the same tx (value-mint notification)". However, the value-mint notification is inside the execution frame (`execute_l1_transaction_and_notify_result` line 583–703), and `last_touched_in_tx` is part of the HistoryMap value — not `element_properties` — so it IS rolled back when the frame reverts. The code itself confirms this at `generic_pubdata_aware_plain_storage.rs:180`: *"Since in case of revert it should become cold again."*

When the L1 tx body reverts:
1. Inner frame rollback (line 703) undoes the value-mint notification's storage effects, including warm metadata
2. Outer frame rollback (line 188 in `process_l1_transaction`) reverts to pre-tx state
3. Post-execution notifications (lines 289, 348) hit the L2AssetTracker slot cold

**Impact: economic, not correctness.** Since post-execution notifications use `FORMAL_INFINITE` (line 268), cold-path cost cannot cause failures. But `L1_TX_INTRINSIC_NATIVE_COST` underestimates the actual native cost on the revert path, meaning the operator bears more proving work than what was reserved from the user's gas budget.

Suggestion: update the comment to document the cold-path case and consider whether the constant should use the cold-path cost as the true worst case (since revert is a valid execution path for L1 txs).

AntonD3 (approved):

PR #588, opened 2026-03-23
feat: migrate EVM interpreter to custom U256

0xVolosnikov:
## What ❔

An attempt to revive https://github.com/matter-labs/zksync-os/pull/135

This PR migrates the EVM interpreter-facing path from `ruint::aliases::U256` to the custom non-`Copy` `u256::U256`.

It is split into three reviewable commits:
- shared `zk_ee` helpers and trait-bound updates for the new integer type
- `evm_interpreter` migration of stack, frame state, and opcode implementations
- `forward_system` tracer/logger adaptation at the EVM API boundary

## Why ❔

The proving-oriented custom `U256` needs to be exercised in the EVM execution path without forcing a workspace-wide integer migration in one step.

This keeps the change localized to the interpreter and its tracing boundary, reduces copy-heavy assumptions in the EVM stack/frame code, and makes the migration reviewable in commit-sized pieces.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.
- [ ] Code has been formatted.

PR #601, opened 2026-03-30
refactor: defer cache materialization for access-list touches

antoniolocascio:
## What ❔

This PR defers cache materialization for EIP-2930 access-list touches in both storage and account caches.

It changes the touch paths to:
- record warm, unobserved placeholders instead of immediately querying oracle-backed state
- materialize storage slots and account data only when a real read/write path needs the value
- move access-list charging into the lower storage/account cache layers instead of precharging in the bootloader
- preserve iterator, persistence, and deconstruction behavior by continuing to skip unobserved entries until they are materialized

It also adds focused tests for:
- touched-only storage slots staying hidden from iterators
- touched storage slots materializing on first read
- touched accounts staying unobserved until materialization
- low-level `CacheRecord` / `HistoryMap` invariants used by the lazy materialization flow

## Why ❔

The goal is to make access-list warming actually lazy.

Before this change, access-list touches could force unnecessary materialization and oracle IO even when execution never observed the touched slot/account value. That was inconsistent with geth execution witness, blocking ethproofs integration.

This change keeps the warmness/accounting semantics, avoids unnecessary reads, preserves the persistence invariant that observed state is materialized before it is consumed, and keeps the flat-model native charging path independent of cache status.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


antoniolocascio:
Fixed the major ones, the minors aren't valid/worth fixing

0xVolosnikov:
I see here that worst case for CALLDATACOPY is "15462.7" - it is strange, compare with https://github.com/matter-labs/zksync-os/pull/588#issuecomment-4129773489

Maybe I missed some fixes for benchmarks, you can ask AI to investiagate

0xVolosnikov:
@antoniolocascio do we want to include this one?

antoniolocascio:
@0xVolosnikov no, closing it

PR #604, opened 2026-03-31
feat!: migrate to riscv_transpiler (airbender-v2)

ly0va:
## What ❔

Replace the removed `risc_v_simulator` crate with `riscv_transpiler` from zksync-airbender dev branch (v2). This migrates the entire RISC-V execution and oracle infrastructure to the new VM API.

### Key changes

**Dependencies & toolchain:**
- `risc_v_simulator` removed from workspace, replaced by `riscv_transpiler` + `common_constants` (git deps pointing to airbender `dev` branch)
- Toolchain bumped to `nightly-2026-02-10` in both root and `zksync_os/` (required by airbender v2's `blake2s_u32`)
- `zksync_os/reproduce/Dockerfile` updated to match

**oracle_provider:**
- `ZkEENonDeterminismSource`, `OracleQueryProcessor`, and `ReadWitnessSource` are **no longer generic** over memory type
- Processors take `&dyn RamPeek` instead of `&M` where `M: MemorySource`
- `DummyMemorySource` now implements `RamPeek`
- Implements new `NonDeterminismCSRSource` trait from `riscv_transpiler::vm` (with `write_with_memory_access`, `write_with_memory_access_dyn`)

**callable_oracles:**
- All query processors (`ArithmeticQuery`, `BlobCommitmentAndProofQuery`, `FieldOpsQuery`, `HashToPrimeSource`, and their `Native*` variants) simplified to non-generic structs
- Memory reads use `peek_word(addr)` instead of `memory.get(addr, AccessType, &mut trap)`
- Fixed pre-existing overflow check in `read_memory_as_u64` (now uses byte length for bounds check)

**zksync_os_runner:**
- Rewritten to use `VM::run_basic_unrolled` with `SimpleTape` + `RamWithRomRegion`
- Requires `.bin` + `.text` file pair (already produced by `dump_bin.sh` via `objcopy --only-section=.text`)
- `DiagnosticsConfig` parameter removed; flamegraph support via feature-gated `run_basic_unrolled_with_flamegraph`
- `simulate_witness_tracing` rewritten using `SimpleSnapshotter`

**cycle_marker:**
- Fully wired to new `CycleMarkerHooks` API from [zksync-airbender#237](https://github.com/matter-labs/zksync-airbender/pull/237)
- `print_cycle_markers` now takes a `CycleMarker` argument (scoped via `CycleMarkerHooks::with()`)
- Runner uses `VM::<DelegationsCounters, CycleMarkerHooks>` when feature is enabled
- Re-exports `CycleMarkerHooks`, `CycleMarker`, `Mark` from `riscv_transpiler::cycle`
- Inline asm now uses hardcoded `x0` to match transpiler's strict `csrrw x0, 0x7ff, x0` decoder

**forward_system:**
- All 12 query processors and system type aliases updated to non-generic oracle types

**tests/rig:**
- `flamegraph_output` wired through to `run_default_with_flamegraph_path` (replaces old `ProfilerConfig`)
- Oracle factory trait methods return non-generic `ZkEENonDeterminismSource`
- `run_prover` rewritten to use `execution_utils::unrolled::prove_unrolled_for_machine_configuration_into_program_proof`

**Other:**
- Removed stabilized `ptr_as_ref_unchecked` feature gate from `evm_interpreter`
- Updated `Cargo.lock` for latest `zksync_os_interface` (adds `pubdata_used` to `BlockOutput`)
- Rebuilt blake2s test binaries and committed `.text` files
- Fixed pre-existing clippy lints (`sort_by_key`, `default_constructed_unit_structs`, `manual_checked_ops`, conditional `P256VerifyErrors` import)

## Why ❔

The `risc_v_simulator` crate was removed from zksync-airbender in [zksync-airbender#233](https://github.com/matter-labs/zksync-airbender/pull/233) and [zksync-airbender#235](https://github.com/matter-labs/zksync-airbender/pull/235). The new `riscv_transpiler` provides a more efficient VM with batch execution, optional JIT compilation (x86_64), and integrated witness snapshotting. This migration is required to stay on the airbender dev branch.

## Is this a breaking change?
- [x] Yes
- [ ] No

**Breaking API changes:**
- `zksync_os_runner::run()` and `run_and_get_effective_cycles()` no longer take a `DiagnosticsConfig` parameter
- `run_and_get_effective_cycles_from_bytes()` now requires both `img_bytes` and `text_bytes`
- `ZkEENonDeterminismSource`, `ReadWitnessSource`, `OracleQueryProcessor` are no longer generic over memory type
- `cycle_marker::print_cycle_markers()` now takes a `CycleMarker` argument instead of reading from a global
- `MemorySource` trait replaced by `RamPeek` (re-exported from `riscv_transpiler::vm`)

## Checklist


Cargo.toml
```diff
@@ -65,13 +65,14 @@ categories = ["cryptography"]
...
-gpu_prover = { git = "https://github.com/matter-labs/zksync-airbender", tag = "v0.5.2" }
+riscv_transpiler = { git = "https://github.com/matter-labs/zksync-airbender", rev = "8ef06cf8dc63b04e4b309b501d54bb571e86a1a9" }
+blake2s_u32 = { git = "https://github.com/matter-labs/zksync-airbender", rev = "8ef06cf8dc63b04e4b309b501d54bb571e86a1a9", default-features = false }
+common_constants = { git = "https://github.com/matter-labs/zksync-airbender", rev = "8ef06cf8dc63b04e4b309b501d54bb571e86a1a9" }
+prover_examples = { git = "https://github.com/matter-labs/zksync-airbender", rev = "8ef06cf8dc63b04e4b309b501d54bb571e86a1a9" }
+execution_utils = { git = "https://github.com/matter-labs/zksync-airbender", rev = "8ef06cf8dc63b04e4b309b501d54bb571e86a1a9" }
+prover = { git = "https://github.com/matter-labs/zksync-airbender", rev = "8ef06cf8dc63b04e4b309b501d54bb571e86a1a9" }
+cli = { git = "https://github.com/matter-labs/zksync-airbender", rev = "8ef06cf8dc63b04e4b309b501d54bb571e86a1a9" }
+gpu_prover = { git = "https://github.com/matter-labs/zksync-airbender", rev = "8ef06cf8dc63b04e4b309b501d54bb571e86a1a9" }
```
antoniolocascio: Should we make a tag for pre-release "v2"?

callable_oracles/src/arithmetic/mod.rs
```diff
@@ -2,14 +2,16 @@ use basic_system::system_functions::modexp::{
...
+use oracle_provider::RamPeek;
 
 use crate::utils::{
     evaluate::{read_memory_as_u64, read_struct},
     usize_slice_iterator::UsizeSliceIteratorOwned,
 };
 use crate::{read_host_struct, read_u64_words};
 
+const HOST_MODEXP_ADVICE_MAX_BYTES: usize = 32 * 1024 * 1024;
+
```
antoniolocascio: I think this was removed on base branch

callable_oracles/src/arithmetic/mod.rs
```diff
@@ -158,14 +142,21 @@ impl<M: MemorySource> OracleQueryProcessor<M> for NativeArithmeticQuery<M> {
...
 
-        let mut n: Vec<u64> = read_u64_words(arg.a_ptr, a_len_u64_words);
-        let mut d: Vec<u64> = read_u64_words(arg.modulus_ptr, modulus_len_u64_words);
+        let mut n: Vec<u64> =
+            read_u64_words(arg.a_ptr, a_len_u64_words, HOST_MODEXP_ADVICE_MAX_BYTES);
+        let mut d: Vec<u64> = read_u64_words(
+            arg.modulus_ptr,
+            modulus_len_u64_words,
+            HOST_MODEXP_ADVICE_MAX_BYTES,
+        );
```
antoniolocascio: This part of the diff seems also a bad merge

callable_oracles/src/field_hints/mod.rs
```diff
@@ -129,7 +111,7 @@ impl<M: MemorySource> OracleQueryProcessor<M> for NativeFieldOpsQuery<M> {
         const { assert!(8 == core::mem::size_of::<usize>()) };
         assert!(arg.src_ptr > 0);
         assert_eq!(arg.src_len_u32_words, 8);
-        let n: Vec<u64> = read_u64_words(arg.src_ptr, u64::from(arg.src_len_u32_words / 2));
+        let n: Vec<u64> = read_u64_words(arg.src_ptr, u64::from(arg.src_len_u32_words / 2), 32);
```
antoniolocascio: Same here

callable_oracles/src/utils/evaluate.rs
```diff
@@ -1,13 +1,8 @@
...
-use risc_v_simulator::cycle::status_registers::TrapReason;
-
-pub fn read_memory_as_u8<M: MemorySource>(
-    memory: &M,
-    offset: u32,
-    len: u32,
-) -> Result<Vec<u8>, ()> {
+use oracle_provider::RamPeek;
+
+pub fn read_memory_as_u8(memory: &dyn RamPeek, offset: u32, len: u32) -> Result<Vec<u8>, ()> {
```
antoniolocascio, line 5: This part of the diff seems also a bad merge

callable_oracles/src/lib.rs
```diff
@@ -79,18 +79,21 @@ fn validate_host_pointer(ptr_u64: u64, alignment: usize) -> *const u8 {
...
-    len_units.checked_mul(bytes_per_unit).unwrap()
+    let requested_bytes = len_units.checked_mul(bytes_per_unit).unwrap();
+    assert!(requested_bytes <= max_bytes);
+
+    requested_bytes
 }
 
 #[inline(always)]
-pub(crate) fn read_u64_words(ptr_u64: u64, len_words_u64: u64) -> Vec<u64> {
+pub(crate) fn read_u64_words(ptr_u64: u64, len_words_u64: u64, max_bytes: usize) -> Vec<u64> {
```
antoniolocascio: This part of the diff seems also a bad merge

.github/workflows/bench.yml
```diff
@@ -92,7 +94,7 @@ jobs:
           cargo install cargo-binutils
           rustup component add llvm-tools-preview
           rustup component add rust-src
-          ./dump_bin.sh --type for-tests
+          ./dump_bin.sh --type for-tests-benchmarking || ./dump_bin.sh --type for-tests
```
antoniolocascio, line 118: I'm pretty sure this is also unrelated

tests/rig/src/chain.rs
```diff
@@ -190,7 +187,7 @@ pub struct RunConfig {
     // Runtime execution controls for `Chain` block execution.
     // Setup conveniences (for example, treasury pre-funding) are owned by `TestingFramework`.
     // Config for the profiler
```
antoniolocascio: I guess this isn't a config anymore, just output path

zksync_os/dump_bin.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -e
 
-USAGE="Usage: $0 --type {singleblock-batch|singleblock-batch-logging-enabled|debug-in-simulator|evm-replay|evm-replay-benchmarking|multiblock-batch|multiblock-batch-logging-enabled|evm-tester|for-tests}"
+USAGE="Usage: $0 --type {singleblock-batch|singleblock-batch-logging-enabled|debug-in-simulator|evm-replay|evm-replay-benchmarking|multiblock-batch|multiblock-batch-logging-enabled|evm-tester|for-tests|for-tests-benchmarking}"
```
antoniolocascio, line 4: This makes sense, but is it intentional?

zksync_os_runner/src/lib.rs
```diff
@@ -46,142 +24,218 @@ pub fn run_default_with_flamegraph_path(
...
+    run_inner(&bin_words, &text_words, cycles, non_determinism_source)
+}
+
+fn run_inner(
+    bin_words: &[u32],
+    text_words: &[u32],
+    cycles: usize,
+    mut non_determinism_source: impl NonDeterminismCSRSource,
+) -> ([u32; 8], Option<u64>) {
+    println!("ZK RISC-V transpiler is starting");
```
antoniolocascio: Let's take the opportunity and remove this, maybe change it to a log than can be disabled

zksync_os_runner/src/lib.rs
```diff
@@ -46,142 +24,218 @@ pub fn run_default_with_flamegraph_path(
...
+    let (bin_words, text_words) = load_bin_and_text(&img_path, &text_path);
+
+    let instructions = preprocess_bytecode::<FullUnsignedMachineDecoderConfig>(&text_words);
+    let tape = SimpleTape::new(&instructions);
+    let mut ram =
+        RamWithRomRegion::<{ ROM_SECOND_WORD_BITS }>::from_rom_content(&bin_words, RAM_SIZE);
+    let mut state = State::initial_with_counters(DelegationsCounters::default());
+
+    let mut config = FlamegraphConfig::new(sym_path, output_path);
+    config.frequency_recip = 1;
```
antoniolocascio: Why did we get rid of the diagnostic config? I found it useful several times to change the frequency

zksync_os_runner/src/lib.rs
```diff
@@ -46,142 +24,218 @@ pub fn run_default_with_flamegraph_path(
...
+    (output, Some(cycles_executed))
 }
 
 pub fn simulate_witness_tracing(
     img_path: PathBuf,
-    non_determinism_source: impl NonDeterminismCSRSource<VectorMemoryImplWithRom>,
+    mut non_determinism_source: impl NonDeterminismCSRSource,
 ) {
-    println!("ZK RISC-V simulator is starting");
+    println!("ZK RISC-V transpiler witness tracing is starting");
```
antoniolocascio: Same as the other print

antoniolocascio (commented):
Generally looks good on a (manual) first pass, mostly pointed out some leftovers from a rebase/merge. Will do another pass with AI

antoniolocascio:
Code wise it looks good, but I'm concerned about performance. I did some benching on my machine (running transactions test instance with risc-v enabled) before and after the PR. Before, the test took ~3s on my machine, hitting max memory usage of 0.18GB. After, the test takes ~25s and uses over 4GB of memory. 
We need to understand what's happening, this isn't acceptable for CI imo. Now the bench job takes over 40mins. I'll start a thread on slack

zksync_os_runner/src/lib.rs
ly0va: Diagnostic config was removed upstream in riscv_transpiler. We can add the frequency config in a follow-up PR.

.github/workflows/bench.yml
ly0va: Related to the other comment about cycle markers

zksync_os/dump_bin.sh
ly0va: Yes, this is intentional. New cycle markers panic during proving, so we have to enable them for benchmarking but disable for proving tests. Whereas previously they could be left enabled for proving and use the same `--type` as benchmarking

callable_oracles/src/utils/evaluate.rs
ly0va: This is intentional, replaces MemorySource with RamPeek and re-formats.

Cargo.toml
ly0va: We had previously discussed with @0xVolosnikov and decided that a commit ref is fine for now.

ly0va: once https://github.com/matter-labs/zksync-airbender/pull/253 gets merged, we should update to the new ref

zksync_os_runner/src/lib.rs
antoniolocascio: I don't get it. Here we're setting frequency_recip to 1, this is the only config I care about (alongside output path). Do you mean that the struct is gone? I would ideally define a simple struct here with at least those two fields, but it's ok if you want to do it as follow up.
Feel free to resolve

ly0va: Ah ok I see, I'll make a follow up

ly0va: https://github.com/matter-labs/zksync-os/pull/612

0xVolosnikov:
@vv-dev-ai ptal

vv-dev-ai:
[AI-review] Reviewed commit `b435d998c726b41f57ddbad762414c90b7a2be1e` — **COMMENT** (2 inline comments).

zksync_os_runner/src/lib.rs
```diff
@@ -46,142 +24,240 @@ pub fn run_default_with_flamegraph_path(
...
     // our convention is to return 32 bytes placed into registers x10-x17
+    let output: [u32; 8] = std::array::from_fn(|i| state.registers[10 + i].value);
 
-    // TODO: move to new simulator
-    #[allow(deprecated)]
-    (
-        run_result.state.registers[10..18].try_into().unwrap(),
-        block_effective,
-    )
+    (output, block_effective.or(Some(cycles_executed)))
```
vv-dev-ai, line 106: [AI-review] **Low:** `block_effective.or(Some(cycles_executed))` changes the API contract: previously `run_and_get_effective_cycles` returned `None` when cycle_marker was disabled; now it always returns `Some(cycles_executed)`. Current callers handle both cases (`if let Some(eff) = ...`), so no breakage today, but downstream consumers expecting `None` to mean "no cycle markers" will misinterpret the value.

vv-dev-ai (commented):
[AI-review] PR #604 — **COMMENT**

- Performance regression (3s→25s, 0.18GB→4GB per @antoniolocascio) needs investigation before merging. The `RAM_SIZE = 1 << 30` (1 GiB) in `zksync_os_runner` may contribute to the memory spike if `RamWithRomRegion::from_rom_content` eagerly allocates. Consider profiling with a smaller RAM or verifying the transpiler uses lazy/sparse allocation.

- `DummyMemorySource::peek_word` now calls `unreachable!()` (was: return 0 with `NoTrap`). This is correct — `DummyMemorySource` is only passed to `Native*` query processors that never read memory. But if any future code path passes it to a non-native processor, it will panic instead of silently returning 0. Worth a brief note in the struct doc.

### Suggested Tests
- **Cycle-count equivalence**: For a reference RISC-V binary, compare old simulator cycle count against `(state.timestamp - INITIAL_TIMESTAMP) / TIMESTAMP_STEP` from the transpiler to catch counting discrepancies between the two VMs.

tests/fuzzer/fuzz/Cargo.toml
```diff
@@ -287,14 +287,14 @@ bench = false
 
 [[bin]]
 name = "precompiles_diff_keccak256"
-path = "fuzz_targets/precompiles_diff/sha256.rs"
+path = "fuzz_targets/precompiles_diff/keccak256.rs"
```
vv-dev-ai, line 290: [AI-review] Good catch — `precompiles_diff_keccak256` and `precompiles_diff_ripemd160` were both pointing at `sha256.rs`. This was silently running the wrong fuzz target for those two binaries.

zksync_os_runner/src/lib.rs
0xVolosnikov: @ly0va ptal, if makes sense we can apply additional changes in a follow up PR

0xVolosnikov (approved):

vv-dev-ai:
[AI-review] Reviewed commit `b435d998c726b41f57ddbad762414c90b7a2be1e` — **COMMENT** (0 inline comments, reply-review).

vv-dev-ai (commented):
[AI-review] PR #604 — **COMMENT**

- **`run-inner-return-change`** ❌ STILL OPEN — @0xVolosnikov acknowledged the API contract change (`block_effective.or(Some(cycles_executed))` always returning `Some`) and suggested a follow-up PR. Finding remains open for tracking.
- **`perf-regression`** ✅ RESOLVED — @0xVolosnikov approved the PR after the finding was raised, indicating the team has accepted the regression for now (likely to be addressed by upstream transpiler improvements or a follow-up).

PR #606, opened 2026-04-01
fix: reject non-strict RLP encodings for list lengths

antoniolocascio:
## What ❔
Enforce strict encoding of RLP lengths

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/basic_bootloader/src/bootloader/transaction/rlp_encoded/rlp/minimal_rlp_parser.rs
+++ b/basic_bootloader/src/bootloader/transaction/rlp_encoded/rlp/minimal_rlp_parser.rs
@@ -64,6 +64,10 @@ impl<'a> Rlp<'a> {
         if s.len() > 4 {
             return Err(InvalidTransaction::InvalidStructure);
         }
+        // Reject leading zeros
+        if s.len() >= 1 && s[0] == 0 {
+            return Err(InvalidTransaction::InvalidStructure);
+        }
         let mut v: u32 = 0;
         for &b in s {
             v = (v << 8) | b as u32;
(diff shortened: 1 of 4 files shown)


0xVolosnikov:
@vv-dev-ai PTAL

vv-dev-ai:
[AI-review] Reviewed commit `003084a498ea0cd0f9b9c20c274d0a1fc0e6a404` — **APPROVE** (1 inline comment).

basic_bootloader/src/bootloader/transaction/rlp_encoded/rlp/minimal_rlp_parser.rs
```diff
@@ -601,5 +605,15 @@ mod tests {
 
         let mut rlp = Rlp::new(&[0x82, 0x00, 0x01]); // Represents 1 with leading zero
         assert!(rlp.u256().is_err());
+
+        let mut long_string = vec![0xba, 0x00, 0x38]; // Long-form length for 56 with a leading zero
```
vv-dev-ai: [AI-review] **Nit:** This test passes even without the new check. `0xba - 0xb7 = 3`, so `take_exact(3)` reads `[0x00, 0x38, 0x80]` (= 14464). Then `take_exact(14464)` fails because only 55 payload bytes remain — the error comes from insufficient data, not from the leading-zero rejection.

Use `0xb9` instead (2-byte length encoding) so the length bytes are `[0x00, 0x38]` = 56 with a leading zero. That value passes the existing `len < 56` guard, so the *only* reason for rejection is the new check.

The list test below is correct: `0xf9 - 0xf7 = 2` → reads `[0x00, 0x38]` → properly isolated.

vv-dev-ai (approved):
[AI-review] PR #606 — **APPROVE**

Correct fix — `be_u32` is the single choke-point for both `bytes()` (line 97) and `list()` (line 130) long-form length parsing, so the leading-zero check covers all multi-byte RLP lengths. Consistent with the existing pattern in `u64()` (line 171).

basic_bootloader/src/bootloader/transaction/rlp_encoded/rlp/minimal_rlp_parser.rs
antoniolocascio: Fixed

0xVolosnikov (approved):

zk_ee/src/utils/aligned_vector.rs
```diff
@@ -198,7 +198,7 @@ pub struct UsizeSliceWriter<'a> {
     _marker: core::marker::PhantomData<&'a ()>,
 }
 
-impl<'a> UsizeWriteable for UsizeSliceWriter<'a> {
```
0xVolosnikov, line 201: Purely to make sure life isn't a bed of roses for Lyova (conflicts)

vv-dev-ai:
[AI-review] Reviewed commit `fb9dc5760a27d19127d8ccb2154bb90298b5c2b3` — **APPROVE** (0 inline comments).

vv-dev-ai (commented):
[AI-review] PR #606 — **APPROVE**

✅ RESOLVED: String test now uses `0xb9` (2-byte length encoding) — properly isolates the leading-zero rejection as the sole failure cause.

PR #607, opened 2026-04-01
fix: order of handleFinalizeBaseTokenBridging and mint corresponds to era-contracts

Raid5594:
## What ❔

During L1→L2 deposit processing, the bootloader now calls
`handleFinalizeBaseTokenBridgingOnL2` on L2AssetTracker for each token
movement (value mint, operator fee, refund) before updating the balance, 
matching the call order from era-contracts' `L2BaseTokenEra.sol`.

## Why ❔

Previous order was breaking the logic as base token balance was not 0 before 
the first `handleFinalizeBaseTokenBridging` call. 

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


0xVolosnikov:
Rebase pls

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
0xVolosnikov: Should we remove it? Also in other places

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -268,6 +268,18 @@ where
     let mut inf_resources = S::Resources::FORMAL_INFINITE;
 
     let coinbase = system.get_coinbase();
+    // Notify asset tracker about the operator fee portion of the deposit.
+    notify_l2_asset_tracker::<S>(
```
StanislavBreadless: This function provides sl chain id, but l2basetokenera uses l1 chain id.

StanislavBreadless: the semantic meaning of this var is about "which chain do we receive the token from". we never receive it from sl chain id, we only receive it from either other l2 (during interop) or from l1 (during deposit)

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -612,14 +611,28 @@ where
...
+        // Notify L2AssetTracker about the value mint portion of the deposit.
+        // This is inside the execution frame, so it gets rolled back if the
+        // main tx body reverts — matching the treasury transfer above.
+        // Use with_infinite_ergs so the call cannot fail due to out-of-gas,
+        // but native consumption is still tracked against the user's resources.
+        //
+        // Notify the asset tracker BEFORE changing balances/totalSupply, so that
+        // _needToForceSetAssetMigrationOnL2 can use totalSupply() == 0 consistently.
+        //
+        // Flow in this file should replicate the behaviour of the following call:
```
StanislavBreadless: is it possible to move it out into a separate function? 

like say

```
fn mint_base_token
```

that would be exact equivalent of the l2basetokenera's function. This would remove:
- duplication in multiple places where we handle refund recipient
- make verification a lot easier as there are two functions that must match exactly same

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -738,6 +695,44 @@ where
...
+    resources: &mut S::Resources,
+    tracer: &mut impl Tracer<S>,
+    validator: &mut impl TxValidator<S>,
+) -> Result<(), BootloaderSubsystemError>
+where
+    S::IO: IOSubsystemExt,
+    S::Metadata: ZkSpecificPricingMetadata
+        + BasicMetadata<S::IOTypes, TransactionMetadata = TxLevelMetadata<S::IOTypes>>,
+{
+    notify_l2_asset_tracker::<S>(
```
0xVolosnikov, line 723: @antoniolocascio it won't affect pubdata costs during the simulation, right?

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/mod.rs
```diff
@@ -150,6 +152,33 @@ where
...
+        .storage_read::<false>(
+            ExecutionEnvironmentType::NoEE,
+            &mut inf_resources,
+            &L2_ASSET_TRACKER_ADDRESS,
+            &l1_chain_id_slot,
+        )
+        .expect("must read L2AssetTracker L1_CHAIN_ID");
+    U256::from_be_bytes(chain_id.as_u8_array())
+}
+
```
antoniolocascio: Why is this here?

0xVolosnikov:
Also pls ask AI to generate a better PR title and description, so it will be easier to understand for e.g. auditors

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
0xVolosnikov: Probably it's a completely different story here. Just wanted to make sure that we won't have issues, because these storage slots may change relatively often (if we have several txs for example). Probably not a problem anyway

antoniolocascio: That's a good point, I think it might

0xVolosnikov:
Seems like we should always deploy AssetTracker in tests and add check that it exists in tests. Maybe it makes sense to add a debug assert or smth like that. Does this function in AssetTracker return something (UPD: no)?

Point here is that otherwise we can silently pass some tests that should've failed

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/mod.rs
Raid5594: It's next to `read_settlement_layer_chain_id`
or shall we keep it in the process_l1_transaction file?

0xVolosnikov:
Architectural note: L1 transactions and the related bridging are _**really**_ bad.

0xVolosnikov (changes requested):

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/mod.rs
0xVolosnikov: I don't think that `post_tx_op` uses it in any way, so imo it belongs to `process_l1_transaction`, yes

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
0xVolosnikov: We should investigate/fix it in a follow up

antoniolocascio (approved):

0xVolosnikov:
To not forget: most likely we want to remove / add `|| SIMULATION` here: https://github.com/matter-labs/zksync-os/blob/cabbb0abdf24c01a6156a551db7170bea5bb9b24/basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs#L836

0xVolosnikov:
@vv-dev-ai ptal

vv-dev-ai:
[AI-review] Reviewed commit `6baa480687377b1740739f6dbde08dd84237ce33` — **APPROVE** (0 inline comments).

vv-dev-ai (approved):
[AI-review] PR #607 — **APPROVE**

The reordering of `notify_l2_asset_tracker` before `transfer_from_treasury` is correct per the era-contracts `L2BaseTokenEra.sol` specification, ensuring `_needToForceSetAssetMigrationOnL2` sees `totalSupply() == 0` before the first mint.

The chain ID source change from `read_settlement_layer_chain_id` (SystemContext slot 0) to `read_l1_chain_id` (L2AssetTracker slot 154) is semantically correct — the `_fromChainId` parameter represents the chain tokens are bridged FROM, which for L1→L2 deposits is the L1 chain, not the settlement layer. This distinction matters in Gateway mode.

Rollback behavior verified for all three `mint_base_token` call sites: value mint (inside execution frame — reverts correctly), operator fee and refund (outside frame with `FORMAL_INFINITE` — persist correctly). If `notify_l2_asset_tracker` succeeds but `transfer_from_treasury` fails at the outside-frame call sites, the error propagates fatally, so the inconsistency never becomes visible.

Note: 0xVolosnikov's follow-up item about `|| SIMULATION` behavior in `notify_l2_asset_tracker` (comment #4177196005) remains open.

StanislavBreadless (approved):
I didnt check the tests, the logic is correct though

0xVolosnikov (approved):
Pls update the revm checker commit

PR #608, opened 2026-04-02
feat(ci): parallelize bench workflow

ly0va:
## What ❔

Split benchmark workflow into 2 that can run in parallel, instead of sequentially. Then aggregate the results.

## Why ❔

Shaves off ~15 minutes off CI time.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -15,29 +15,23 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  vm-benchmarks:
-    name: Run benchmarks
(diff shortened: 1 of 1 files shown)


antoniolocascio (approved):

0xVolosnikov:
I'm not sure that this is too different compared to the current implementation tbh

0xVolosnikov:
Probably both base and head share a lot of deps etc, so the sequential setup also benefits a lot from the local cache. Maybe for big PRs changing a lot it doesn't work - and that's why you've had a 40 mins runs. But such big PRs are relatively rare

PR #610, opened 2026-04-02
fix: add extra intrinsic pubdata to account for asset tracker call in simulation

antoniolocascio:
## What ❔
Add extra intrinsic pubdata to account for asset tracker call in simulation for L1 transactions

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
+++ b/basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
@@ -740,7 +740,7 @@ where
     S::Metadata: ZkSpecificPricingMetadata
         + BasicMetadata<S::IOTypes, TransactionMetadata = TxLevelMetadata<S::IOTypes>>,
 {
-    notify_l2_asset_tracker::<S>(
+    notify_l2_asset_tracker::<S, Config>(
         system,
         system_functions,
         memories,
@@ -838,7 +838,7 @@ where
 /// If no contract is deployed at L2AssetTracker, the call succeeds silently
 /// (a call to an empty address returns success with no returndata in EVM).
 /// However, we are certain that L2AssetTracker is available after the upgrade.
-fn notify_l2_asset_tracker<'a, S: EthereumLikeTypes + 'a>(
+fn notify_l2_asset_tracker<'a, S: EthereumLikeTypes + 'a, Config: BasicBootloaderExecutionConfig>(
     system: &mut System<S>,
     system_functions: &mut HooksStorage<S, S::Allocator>,
     memories: RunnerMemoryBuffers<'a>,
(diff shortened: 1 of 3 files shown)


basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -78,6 +78,24 @@ where
     // will be refunded to the user.
     let gas_per_pubdata = transaction.gas_per_pubdata_limit.read();
 
+    // It's important to ensure that the amount of pubdata estimated during
```
0xVolosnikov, line 81: The comments are approaching the length of my university thesis. Probably this is better than no comments anyway. Maybe we should update our docs as well, to explain the simulation mode?

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -78,6 +78,24 @@ where
...
+    // transaction simulation is never less than the amount estimated
+    // during execution of the same transaction.
+    // Since the introduction of the asset tracker calls during the base token
+    // minting, the pubdata for the storage diff of the first of such calls
+    // depends indirectly on the gas price, which can fluctuate from simulation
+    // to execution.
+    // Even though the pubdata for this storage change is already considered in
+    // L1_TX_INTRINSIC_PUBDATA (for the calls related to refunds), we need to
+    // add an extra worst case when in simulation to ensure that simulation
```
0xVolosnikov, line 90: Didn't get the idea here - why should we charge again if it's already charged?

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
antoniolocascio: And I wrote them by hand. I'm not sure what we should add to docs, this is a very small detail.

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
antoniolocascio: So, in both runs we add L1_TX_INTRINSIC_PUBDATA to pubdata, which includes already the worst case for this slot. However, we also add the pubdata needed for the state diffs in tx execution. This includes one of the asset tracker calls. Therefore, we count it twice (worst case for the after fee commitment calls and the one I just described). The issue is that the state diff one can depend on the value of amount, which in turn depends on fee. So for simulation, where fee is 0, the asset tracker call might have no pubdata, but we need to consider the worst case to avoid underestimating pubdata.

vv-dev-ai:
[AI-review] Reviewed commit `48a0c8f09f7ac005c9d0c090e46f72ef10caaf86` — **APPROVE** (0 inline comments).

vv-dev-ai (approved):
[AI-review] PR #610 — **APPROVE**

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
0xVolosnikov: I mean, I don't think this is really clear here, may be confusing for reviewers

0xVolosnikov: other than that lgtm

0xVolosnikov (approved):

PR #616, opened 2026-04-07
feat: l2 tx validation intrinsic native resources

AntonD3:
## What ❔

- Introduce explicit intrinsic-resource formulas for L1/L2 txs (gas, computational native, pubdata) and wire them into `create_resources_for_tx` via new helpers: `calculate_tx_intrinsic_gas`, `calculate_l2_tx_intrinsic_computational_native_resources`, `calculate_l1_tx_intrinsic_computational_native_resources`, `calculate_l2_tx_intrinsic_pubdata`.
- Route validation-phase and post-execution native charges (ecrecover, tx hash keccak, account read, nonce bump, fee prepayment, refund, coinbase payment, access list, authorization list) through a dedicated `intrinsic_resources` tracker instead of `main_resources`. In production this tracker is `FORMAL_INFINITE`, so behavior is unchanged.
- Add opt-in `verify_intrinsic_native` cargo feature (enabled in `tests/rig`) that asserts at runtime the intrinsic computational-native formula is an upper bound on actual consumption.
- Add `api::validate_l2_tx_intrinsic_native_resources` helper so the sequencer can check intrinsic gas + native + pubdata pre-validation without booting the full system.
- Consolidate hash native-cost helpers: move `blake2s_native_cost` from `crypto` to `basic_system::cost_constants` (next to keccak/sha/ripemd); expose `keccak256_native_cost_for_rounds_u64` for reuse in `basic_bootloader` constants.
- Rename `L2_TX_INTRINSIC_GAS` → `TX_INTRINSIC_GAS`, `PER_*_NATIVE_COST` → `PER_*_NATIVE_COMPUTATIONAL_OVERHEAD`.

## Why ❔

Validation-phase native consumption was previously charged per-site from `main_resources`, which made it hard to predict how much gas a transaction needed before it could be accepted, and left the relationship between gas / pubdata / native limits implicit. Moving to a single precharged intrinsic lets the sequencer reject underfunded txs up front using `api::validate_l2_tx_intrinsic_native_resources`, and keeps runtime charges on an isolated tracker so the upper-bound invariant can be verified mechanically under `verify_intrinsic_native`.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -140,8 +140,28 @@ where
         .ok_or(TxError::Validation(InvalidTransaction::PubdataPriceTooHigh))?;
     let native_prepaid_from_gas = native_per_gas.saturating_mul(tx_gas_limit);
 
+    let mut access_list_accounts = 0;
+    let mut access_list_storage_keys = 0;
+    let mut authorization_list_num = 0;
+    if let Some(iter) = transaction.access_list_iter() {
```
AntonD3, line 146: It duplicates a code for access/authorization lists processing, it also affects performance a bit. But I didn't want to change flow significantly in this PR, will be solved in follow up PRs.

vv-dev-ai:
[AI-review] Reviewed commit `0a7ab62a2b42dea1114bb491998f158e442a5054` — **REQUEST_CHANGES** (1 inline comment).

basic_bootloader/src/bootloader/transaction_flow/gas_helpers.rs
```diff
@@ -188,6 +195,62 @@ impl<S: EthereumLikeTypes> core::fmt::Debug for ResourcesForTx<S> {
...
+    intrinsic_computational_native_resources = intrinsic_computational_native_resources
+        .saturating_add(
+            access_list_storages
+                .saturating_mul(L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_ACCESS_LIST_PER_STORAGE_KEY),
+        );
+
+    intrinsic_computational_native_resources = intrinsic_computational_native_resources
+        .saturating_add(
+            authorization_list_num.saturating_mul(L2_TX_INTRINSIC_PUBDATA_PER_AUTHORIZATION),
+        );
```
vv-dev-ai, line 236: [AI-review] **High:** Wrong constant — `L2_TX_INTRINSIC_PUBDATA_PER_AUTHORIZATION` (= 80, a pubdata constant) is used where `L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_AUTHORIZATION` (= 826,380) should be.

Trigger: Any L2 transaction with EIP-7702 authorization entries.

Impact: Each authorization entry is charged 80 native instead of 826,380 — a ~10,300× undercharge. The actual proving work (ecrecover 522k, account read 250k, delegation write 26k, etc.) is not covered by the intrinsic cost. Since validation now runs under `FORMAL_INFINITE`, the undercharge is invisible at runtime but means blocks with many auth entries are underpriced for the prover. Violates resource accounting invariant I10.

Evidence: `L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_AUTHORIZATION` is defined at `constants.rs:104` but never imported in `gas_helpers.rs` — the import list (lines 2-9) imports `L2_TX_INTRINSIC_PUBDATA_PER_AUTHORIZATION` instead.

Fix: Change `L2_TX_INTRINSIC_PUBDATA_PER_AUTHORIZATION` → `L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_AUTHORIZATION` on this line, and update the import.

vv-dev-ai (changes requested):
[AI-review] PR #616 — **REQUEST_CHANGES**

### L2_TX_INTRINSIC_PUBDATA constant

Comment says coinbase = 66 (`32 + 1 + 33`), matching `COINBASE_BALANCE_INTRINSIC_PUBDATA = 32 + 34 = 66` defined in the same file, but the code uses `68 + 64 = 132` instead of `68 + 66 = 134`. 2-byte pubdata undercharge per L2 tx.

### Comment label mismatch in `L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_COST`

Items 5 and 6 label says "37500" but the arithmetic `2500 + 17_500 * 3` evaluates to 55,000. The code correctly uses 55,000. Comment-only issue.

### Suggested Tests

- Test `calculate_l2_tx_intrinsic_computational_native_resources` with `authorization_list_num > 0` and verify the result scales by `L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_AUTHORIZATION` (826,380), not `L2_TX_INTRINSIC_PUBDATA_PER_AUTHORIZATION` (80).
- Test that the API `validate_l2_tx_intrinsic_native_resources` function rejects a tx that passes with the buggy constant but fails with the correct one (a tx with many auth entries and barely sufficient gas).

0xVolosnikov:
It is not fully obvious for me why do we want to follow this approach, please explain benefits and your position tomorrow. 

I'm very concerned that this formula-based calculation will be hard to maintain. We must not forget to change it every time when validation changes for some reason and currently the correctness is not tested. Even with tests it may be a non-trivial task to guarantee correctness, because we need to check with the worst possible case. And this maintainability issue is significant from my PoV - charging is not "local" (charging is located far from the actual changed code) and probability of issues is rather high. So it feels less safe, and based on that I'd prefer "real" charging based on actually executed operations.

Also tbh I don't like that we have to write a long comments explaining how constants are calculated - they are also likely to rot (we already have some mistakes there as I see). Also ofc it would be preferable to use constants instead of magic numbers there.

So let's discuss it, maybe I do not fully understand all benefits or mentioned problems may be easily solved somehow.

0xVolosnikov:
Also it's unfortunate that CI fails - this gets in the way, would be better to have a working PoC when we discuss the rather late changes to this version

antoniolocascio:
@0xVolosnikov I get your concerns. The formula looks scary, indeed, but that's just because of the implementation being very precise about the derivation. I'm pretty sure we can make the definition more programatic, which should improve it a bit. What I would actually do is collect the validation native and pubdata. Instead of using infinite all the time, keep a resource only for validation operations.
We could pick several approaches:
1. Use this data only for testing (and adding test cases for worst case validation) to prevent the constants from going out of sync
2. Also, assign the intrinsic as the limit for this new resource and log if it ever runs out, then monitor for this log 

The benefit is what we discussed last week: make a simple and quick rule to determine if a tx is admitted in the mempool while guaranteeing that the estimateGas always returns a value that will be admitted.

antoniolocascio:
But after looking at the complexity, I think this is better to go in 0.4.0, if we decide to do it. Let's discuss it today

AntonD3:
@0xVolosnikov  TBH I thought we agreed during the discussion last week on this approach, and the only question left was whether we want to include it in 0.3.0, or delay until 0.4.0. But if you still have concerns, I'll try to remember all the benefits, and we can discuss it once again.

Firstly, the main benefit is the predictability of the validation of native resources usage. Obviously, it's especially beneficial for tx validation in the mempool. And IMHO it simplifies the design in some sense as well - charging of native happens only once, not distributed among the whole codebase in the different components - code simpler, and I think it positively impacts debugging, and general "understading" of tx resources usage.

The alternative to this approach is keeping code as it is - charging native resources as they are used during validation. Then, in the mempool, we have to either try to predict native resources using formulas(perhaps easier to test, but not beautiful code), or run zksync os. But even in this case, it will not work without changes in zksync os - currently charging is state-dependent, for example, execution of previous tx in the mempool can change caller balance for the next one, it affects validation pubdata, and transitively may affect validity of the transactions. Perhaps a similar problem can be present for authorization lists. So at least some parts should be changed to charge using the worst case. Btw, if you just change mempool check to account for worst case, it will not work - then it's possible that tx will fail with gas_limit returned from estimateGas. You can also say that it's possible to hack estimateGas somehow, but perhaps it's just easier to change the zksync os at this point. 

For the same reason, we have to charge storage accesses using the worst-case scenario, or account for it in the mempool(which may not be trivial, extra logic, extra state access)

And one more note to the previous paragraph, IMHO, guaranteed validity of the tx in the mempool is really important, e.g., if the transaction was accepted to the mempool, it can not become invalid(at least by user-controllable factors). Otherwise, we are opening a DDOS vector on our sequencer - attacker pays nothing but their txs, spending sequencer time. Of course, it was working fine without it for a long time, and perhaps not so important for 0 gas price chains, but I think in the long term it's not sustainable in any case. And if we don't solve it now, we'll be doing negative work.

AntonD3:
As for complexity, @antoniolocascio  @0xVolosnikov  can you clarify what the most complex part is in the PR, and why you think that formula is scary? I assume you are worried mostly about constants? If so, then as @antoniolocascio  mentioned above, all these long comments with explanations are an artifact of my approach to computing initial values. It was the easiest approach as I had formulas from my previous PR, but indeed, we can try to find a more imperative way to compute these values. We don't have to be really precise here, I was because it was easy using my previous work. Generally I think we should treat it similar way EVM treats its intrinsic cost.

And generally, I tried to make this PR as small as possible. There are 3 main changes:
- introduce formula and consts
- Use this formula
- Do not charge anything else during validation: infinite for computation, remove pubdata charge
Perhaps I can undo some small refactoring of the constants file to minimize the diff(the only thing I've done apart from main changes).

Though I don't have a strong opinion on including it in 0.3.0, just trying to make sure that the implementation and diff are not a bottleneck for a decision.

antoniolocascio:
@AntonD3 for me the scary part is mostly the big comments that then need to correspond to a number. It's really hard to read. But we agree, this is solvable. The question is how much we want to rush this/delay 0.3.0.

0xVolosnikov:
For me, the most scary aspect is that we somehow need to guarantee that, in the future, the calculations will remain correct.

All in all, I am trying to weigh for myself whether the benefits of these changes currently justify the mentioned risks—plus the inherent risks associated with modifying critical code. Let's discuss the possible options on a call and make a decision.

api/src/helpers.rs
```diff
@@ -372,3 +377,83 @@ pub fn encode_set_settlement_layer_chain_id_calldata(new_sl_chain_id: U256) -> V
...
+    pubdata_price: U256,
+    gas_limit: u64,
+    calldata_length: u64,
+    access_list_accounts: u64,
+    access_list_storage_keys: u64,
+    authorization_list_num: u64,
+    max_fee_per_gas: U256,
+    max_priority_fee_per_gas: U256,
+) -> Result<(), ()> {
+    // Validate fee fields
```
0xVolosnikov, line 407: By the name and description this function is supposed to validate **native** resources, so it may be confusing to have purely gas-related checks here. Probably we should rename it better

api/src/helpers.rs
```diff
@@ -372,3 +377,83 @@ pub fn encode_set_settlement_layer_chain_id_calldata(new_sl_chain_id: U256) -> V
...
+    if max_priority_fee_per_gas > max_fee_per_gas {
+        return Err(());
+    }
+    if base_fee > max_fee_per_gas {
+        return Err(());
+    }
+
+    // Compute effective gas price (same as get_gas_price in bootloader)
```
0xVolosnikov: > same as get_gas_price in bootloader

Is it actually possible to reuse the same code in both places?

api/src/helpers.rs
```diff
@@ -372,3 +377,83 @@ pub fn encode_set_settlement_layer_chain_id_calldata(new_sl_chain_id: U256) -> V
...
+
+    let native_prepaid = native_per_gas.saturating_mul(gas_limit);
+
+    // Intrinsic pubdata
+    let intrinsic_pubdata = calculate_l2_tx_intrinsic_pubdata(authorization_list_num);
+    let intrinsic_pubdata_overhead = native_per_pubdata.saturating_mul(intrinsic_pubdata);
+
+    let native_limit = native_prepaid
+        .checked_sub(intrinsic_pubdata_overhead)
+        .ok_or(())?;
```
0xVolosnikov, line 466: Do we want to return the exact failure reason?

basic_bootloader/src/bootloader/transaction/access_list.rs
```diff
@@ -32,7 +32,7 @@ where
             // per-address charge
             resources.charge(&S::Resources::from_ergs_and_native(
```
0xVolosnikov, line 33: Should we replace it with `from_ergs`?

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
0xVolosnikov, line 275: Let's add a comment that it should never happen

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
0xVolosnikov: > follow up PRs

Imo it would be better to not have the follow ups in this version. If we are talking about the next version, pls add a TODO and issue in the task tracker

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -150,9 +170,13 @@ where
         transaction.is_deployment().is_some(),
         calldata.len() as u64,
         calldata_tokens,
-        L2_TX_INTRINSIC_GAS,
-        L2_TX_INTRINSIC_PUBDATA,
-        L2_TX_INTRINSIC_NATIVE_COST,
+        calculate_l2_tx_intrinsic_computational_native_resources(
+            calldata.len() as u64,
+            access_list_accounts,
```
0xVolosnikov: to consider: could we pass the tx reference and count all this values inside instead?

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -233,28 +238,23 @@ where
             }
         }
     };
-    let tx_hash: Bytes32 = transaction.transaction_hash(&mut tx_resources.main_resources)?;
+    let tx_hash: Bytes32 = transaction.transaction_hash(&mut inf_resources)?;
 
     // any IO starts here
 
     // now we can perform IO related parts. Getting originator's properties is included into the
     // intrinsic cost charnged above
```
0xVolosnikov:
```suggestion
    // intrinsic cost charged above
```

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -317,6 +318,8 @@ where
     }
 
     // Access list
+    // we are passing resources, this function should charge only ergs,
```
0xVolosnikov: Note: probably in future we should separate them

vv-dev-ai:
[AI-review] Reviewed commit `934faa230ee1822540eb93ef6cf1ca66dbae6469` — **APPROVE** (1 inline comment).

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -28,9 +28,83 @@ pub const TX_CALLDATA_OFFSET: usize = 0x60;
...
+pub const FREE_L1_TX_NATIVE_PER_GAS: u64 = 10000;
+
+// computational native consts
+/// Constant part of l2 tx intrinsic computational native cost.
+// 1. 30_000 for post validation processing: transferring fee to coinbase, transferring the gas refund, hashing of tx hash into rolling hash.
+// 2. 522000 = 350_000 + 43_000*4 for ecrecover.
+// 3. account read(worst case): 250740 = 500(PREIMAGE_CACHE_GET_NATIVE_COST) + 4000(WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST) + 4000(WARM_STORAGE_READ_NATIVE_COST) + 242240(COLD_NEW_STORAGE_READ_NATIVE_COST)
+// 4. nonce write: 5000 = 4000(WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST) + 1000(WARM_ACCOUNT_CACHE_WRITE_EXTRA_NATIVE_COST)
+// 5. keccak for signing hash(worst case const part - 2 rounds + 1 round precharge for dynamic parts): 37500 = 2500 + 17_500 * 3
+// 6. keccak for full hash(worst case const part - 2 rounds + 1 round precharge for dynamic parts): 37500 = 2500 + 17_500 * 3
```
vv-dev-ai: [AI-review] **Nit:** Comment arithmetic error: `37500 = 2500 + 17_500 * 3` → `2500 + 17_500 * 3 = 55_000`, not 37,500. The code value at line 73 correctly uses `55_000 + 55_000`. The stated total in items 5 and 6 should be updated to match.

vv-dev-ai (approved):
[AI-review] PR #616 — **APPROVE**

✅ **RESOLVED: wrong-auth-native-constant** — `calculate_l2_tx_intrinsic_computational_native_resources` now correctly uses `L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_AUTHORIZATION` (826,380) instead of the pubdata constant (80). Verified in `gas_helpers.rs:228`.

✅ **RESOLVED: pubdata-constant-mismatch** — `L2_TX_INTRINSIC_PUBDATA` is now `68 + 66 = 134`, matching the documented breakdown (sender: 68, coinbase: 66). Test updated accordingly (`test_expensive_pubdata` now uses 134).

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
AntonD3: Yes, in follow-up changes, I am planning to charge intrinsic gas beforehand

api/src/helpers.rs
AntonD3: I thought about this, practically speaking perhaps I will treat this error same way as out of computational native in the mempool. Both computational and pubdata charged from the same value, so IMHO OON for both of them is fine.

api/src/helpers.rs
AntonD3: These checks are done to avoid panic/errors in the computations above, but it's assumed that these values were validated before the function call. I thought about this already, perhaps I can introduce separate errors for such case and/or mention it in the comments

api/src/helpers.rs
AntonD3: Not very trivial, get_gas_price accepts system as argument to access logger. If we want to keep its behavior, we'll need to restructure it somehow. Given that it's 5 lines, decided that it's not worth it

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
AntonD3: This function is used in the helpers as well, and there I wanted to avoid leaking internal bootloader structure, especially given that it will be used in the mempool

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
AntonD3: > Imo it would be better to not have the follow ups in this version.

As we discussed, all the follow ups will go to the 0.4.0

basic_bootloader/src/bootloader/transaction_flow/gas_helpers.rs
AntonD3: fixed

basic_bootloader/src/bootloader/transaction_flow/gas_helpers.rs
```diff
@@ -188,6 +195,62 @@ impl<S: EthereumLikeTypes> core::fmt::Debug for ResourcesForTx<S> {
...
+
+    intrinsic_computational_native_resources = intrinsic_computational_native_resources
+        .saturating_add(
+            access_list_storages
+                .saturating_mul(L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_ACCESS_LIST_PER_STORAGE_KEY),
+        );
+
+    intrinsic_computational_native_resources = intrinsic_computational_native_resources
+        .saturating_add(
+            authorization_list_num.saturating_mul(L2_TX_INTRINSIC_PUBDATA_PER_AUTHORIZATION),
```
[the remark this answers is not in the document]
AntonD3: fixed

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -81,6 +155,29 @@ pub const L1_TX_INTRINSIC_L2_GAS: u64 = 21_000;
...
 
+/// L1 tx calldata byte intrinsic computational native cost.
+// caldata per byte copy: COPY_BYTE_NATIVE_COST = 1
+pub const L1_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_CALLDATA_BYTE: u64 = 1;
+
+// pubdata consts
+/// Constant part of l2 tx intrinsic pubdata.
+// 1. sender account change: 68 = 32(key) + 1(account metadata) + 2(nonce increase) + 33(worst case balance)
+// 2. coinbase: 66 = 32(key) + 1(account metadata) + 33(worst case balance)
+pub const L2_TX_INTRINSIC_PUBDATA: u64 = 68 + 64;
```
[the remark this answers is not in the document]
AntonD3: fixed

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -317,7 +318,7 @@ where
     }
 
     // Access list
-    parse_and_warm_up_access_list(system, &mut tx_resources.main_resources, &transaction)?;
+    parse_and_warm_up_access_list(system, &mut inf_resources, &transaction)?;
```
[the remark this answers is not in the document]
AntonD3: fixed

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -359,7 +360,7 @@ where
         if let Some(authorization_list) = transaction.authorization_list() {
             crate::bootloader::transaction::authorization_list:: parse_authorization_list_and_apply_delegations(
                     system,
-                    &mut tx_resources.main_resources,
+                    &mut inf_resources,
                     authorization_list,
                 )?;
```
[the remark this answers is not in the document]
AntonD3: fixed

vv-dev-ai:
[AI-review] Reviewed commit `65052d866d4e333f118b7971cf3c9dc225239a69` — **APPROVE** (0 inline comments).

vv-dev-ai (commented):
[AI-review] PR #616 — **APPROVE** (posted as COMMENT; GitHub disallows self-review)

New commit since last review (`934faa2` → `65052d8`) contains only a comment typo fix ("charnged" → "charged"). No substantive code changes, no new risk patterns. Prior APPROVE stands.

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -28,9 +38,75 @@ pub const TX_CALLDATA_OFFSET: usize = 0x60;
...
+// TODO (EVM-1157): find a reasonable value for it.
+pub const L1_TX_NATIVE_PRICE: U256 = U256::from_limbs([10, 0, 0, 0]);
+
+// Upgrade, service and gateway mailbox transactions are expected to have ~72 million gas. We will use enough
+// gas to ensure that multiplied by the 72 million they exceed the native computational limit.
+pub const FREE_L1_TX_NATIVE_PER_GAS: u64 = 10000;
+
+// computational native consts
+/// Constant part of l2 tx intrinsic computational native cost.
+pub const L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_COST: u64 = ECRECOVER_NATIVE_COST + // signature verification
```
AntonD3, line 74: @0xVolosnikov @antoniolocascio I redefined the constants as you suggested. It's indeed much more flexible. But at the same time, I feel that it's a bit more fragile and maybe less clear. Though I assume that I am somewhat less objective, since I spent a lot of time on it, I'm curious to hear your opinion? Any other suggestions?

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
AntonD3: Added, but perhaps I'll change this part when I pass non infinite native resources for testing

vv-dev-ai:
[AI-review] Reviewed commit `ea745e0e7f4c07a7e51d0d0b38438d9f8070c512` — **APPROVE** (0 inline comments).

vv-dev-ai (approved):
[AI-review] PR #616 — **APPROVE**

New commits since last review (`65052d8` → `ea745e0`) restructure intrinsic native constants from hardcoded values to symbolic formulas using `ECRECOVER_NATIVE_COST`, `WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST`, `COLD_NEW_STORAGE_READ_NATIVE_COST`, `keccak256_native_cost_for_rounds_u64()`, etc. Verified that numerical values are equivalent — the only intentional difference is dropping the `PREIMAGE_CACHE_GET_NATIVE_COST` (500) from account-read components, which was over-counted since `read_account_properties` does not access the preimage cache. Per-operation native charges for access list and authorization list were already zeroed in the prior version; the cleanup to `from_ergs()` is a no-op.

api/src/helpers.rs
0xVolosnikov: Maybe let's mention it for the future

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
```diff
@@ -562,6 +579,45 @@ where
...
+                actual_used <= formula,
+                "intrinsic computational native formula ({}) is not an upper bound on actual consumption ({})",
+                formula,
+                actual_used
+            );
+            // Skip the overcharging check when authorization-list entries are
+            // present: failed auths (bad sig, wrong chain id, nonce overflow)
+            // consume only PER_AUTH_NATIVE_COMPUTATIONAL_OVERHEAD while the
+            // formula budgets worst-case success cost per entry.
+            if context.authorization_list_num == 0 {
```
AntonD3: Unfortunately for txs with authorization lists it's hard to check upper bound

vv-dev-ai:
[AI-review] Reviewed commit `f44bea288b65de26fddfed64421959037dc00903` — **COMMENT** (0 inline comments).

vv-dev-ai (commented):
[AI-review] PR #616 — **COMMENT**

8 new commits since last review (`ea745e0e` → `f44bea28`) introduce significant resource accounting restructuring:

- **`verify_intrinsic_native` feature** (test-only): runtime validation that the intrinsic native formula upper-bounds actual consumption. Enabled in test rig via `basic_bootloader` feature. Uses `assert!` — properly feature-gated, never active in production.
- **`intrinsic_tracker` pattern**: all validation-phase native charges (ecrecover, keccak, account read, nonce increment, fee prepayment, refund, operator payment, access list warming, auth list processing) now route through a dedicated tracker instead of `main_resources` or standalone `FORMAL_INFINITE`. In production the tracker IS `FORMAL_INFINITE`, so behavior is preserved. Under `verify_intrinsic_native` it starts at `1 << 50` so actual consumption can be recovered and compared against the formula.
- **Service tx differentiation**: new `SERVICE_TX_INTRINSIC_COMPUTATIONAL_NATIVE_COST` and `SERVICE_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_CALLDATA_BYTE`; `is_service` parameter added to `calculate_l2_tx_intrinsic_computational_native_resources` and `calculate_l2_tx_intrinsic_pubdata`.
- **Validation pubdata simplification**: `charge_for_validation_pubdata_using_withheld` removed. Validation pubdata is now only recorded (`system.net_pubdata_used()?`) and excluded from post-execution pubdata charging. Relies on intrinsic pubdata pre-deduction (`L2_TX_INTRINSIC_PUBDATA = 134`) covering actual validation pubdata.
- **Access list / auth list per-item native charges**: `access_list.rs` and `authorization_list.rs` now charge `PER_*_NATIVE_COMPUTATIONAL_OVERHEAD` (2000) per item alongside gas. Both components go against `intrinsic_tracker` (infinite in production; formula pre-deducts the total from `main_resources`).
- **`calculate_tx_intrinsic_gas` consolidation**: new shared function computes total intrinsic gas including EIP-2930 access list and EIP-7702 authorization list costs, used by both L1 and L2 paths.
- **`L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_AUTHORIZATION` now includes keccak contribution**: added `133 * DYNAMIC_PART_KECCAK_COMPUTATIONAL_NATIVE_PER_BYTE * 2` for tx signing + full hash overhead.

These are resource-accounting and state-transition changes that warrant a full deep review. Recommend requesting a full review (`@vv-dev-ai ptal`).

0xVolosnikov:
@vv-dev-ai do a new deep review

vv-dev-ai:
[AI-review] Reviewed commit `f44bea288b65de26fddfed64421959037dc00903` — **APPROVE** (2 inline comments).

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -28,9 +38,98 @@ pub const TX_CALLDATA_OFFSET: usize = 0x60;
...
+    (WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST + WARM_ACCOUNT_CACHE_WRITE_EXTRA_NATIVE_COST) * 2 + keccak256_native_cost_for_rounds_u64(1); // post execution logic: transferring fee to coinbase, transferring the gas refund, hashing of tx hash into rolling hash
+
+/// Service tx intrinsic computational native cost.
+pub const SERVICE_TX_INTRINSIC_COMPUTATIONAL_NATIVE_COST: u64 =
+    WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST + WARM_STORAGE_READ_NATIVE_COST + COLD_NEW_STORAGE_READ_NATIVE_COST + // worst case account read
+    keccak256_native_cost_for_rounds_u64(2) + // keccak for full hash, 1 rounds worst case tx size + 1 round precharge for dynamic parts
+    WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST + WARM_ACCOUNT_CACHE_WRITE_EXTRA_NATIVE_COST + // balance change for fee prepayment
+    (WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST + WARM_ACCOUNT_CACHE_WRITE_EXTRA_NATIVE_COST) * 2 + keccak256_native_cost_for_rounds_u64(1); // post execution logic: transferring fee to coinbase, transferring the gas refund, hashing of tx hash into rolling hash
+
+/// L2 tx calldata byte intrinsic computational native cost.
```
vv-dev-ai: [AI-review] **Low:** `keccak256_native_cost_for_rounds_u64(3) * 2` — the comment for the L2 constant says "2 rounds worst case tx size + 1 round precharge" (3 rounds), applied twice for signing hash + full hash. For `SERVICE_TX_INTRINSIC_COMPUTATIONAL_NATIVE_COST` (line 93), `keccak256_native_cost_for_rounds_u64(2)` is used with comment "1 rounds worst case tx size + 1 round precharge". Service txs only need the full hash (no signing hash) — this is correct. However, service txs from `BOOTLOADER_FORMAL_ADDRESS` may have very large calldata (e.g., upgrade txs wrapped as service). If the tx body exceeds ~272 bytes (2 × KECCAK256_CHUNK_SIZE), the full hash requires more than 2 rounds. The per-byte `SERVICE_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_CALLDATA_BYTE` (line 98) covers additional rounds. Verify the per-byte constant + 2-round precharge correctly upper-bounds worst-case keccak for any service tx calldata length.
Trigger: Service tx with calldata exceeding 2 keccak rounds.
Impact: If per-byte cost + base rounds underestimates, the intrinsic formula undercharges. Low impact because service txs have unlimited native in production (gas_price=0).

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
```diff
@@ -275,14 +289,12 @@ where
...
-        Self::charge_for_validation_pubdata_using_withheld(
-            &mut context.resources,
-            &to_charge_for_pubdata,
-        )?;
+        // we are saving amount of pubdata spent during validation,
+        // it's already covered by intrinsic cost, so it will be excluded
+        // from pubdata payment after execution.
+        // In the future we may consider using intrinsic pubdata here,
+        // so worst case validation pubdata will be "refunded" if not used
+        context.validation_pubdata = system.net_pubdata_used()?;
```
vv-dev-ai, line 297: [AI-review] **Low:** Validation pubdata is now only recorded (`net_pubdata_used()`), not charged from resources. The 134-byte `L2_TX_INTRINSIC_PUBDATA` pre-deduction at resource creation compensates. Since sender account pubdata (~68 bytes for nonce+balance diff) is already captured in `net_pubdata_used()` AND added as intrinsic constant, the total reported pubdata (`context.total_pubdata + L2_TX_INTRINSIC_PUBDATA`) conservatively over-reports by ~68 bytes per L2 tx. This is safe (overcharging), but the future optimization mentioned in the comment (lines 295-296) could recover this by using the intrinsic pubdata as validation budget and "refunding" unused portions.

vv-dev-ai (commented):
[AI-review] PR #616 — **APPROVE**

Deep review of the intrinsic native resource accounting restructuring. Classification: `resource-accounting`, `state-transition`, `public-api-contract`. Risk: deep.

The refactoring consolidates all validation-phase and post-execution native costs (ecrecover, keccak, account read, nonce write, fee prepayment, refund, operator payment) into a single intrinsic formula deducted at resource creation time. Operations that previously charged `main_resources` now charge a separate `intrinsic_tracker` (which is `FORMAL_INFINITE` in production, preserving identical behavior).

Verified invariants:
- **I1 (fee conservation):** fee_prepaid == refund + operator_payment holds on all paths. Refund and operator payment now route through `intrinsic_tracker.with_infinite_ergs()` which is `FORMAL_INFINITE` in production — equivalent to the old standalone `FORMAL_INFINITE`.
- **I4 (post-frame FORMAL_INFINITE):** Post-execution balance mutations still use FORMAL_INFINITE resources in production. The `intrinsic_tracker` substitution is identity.
- **I7 (oracle ordering):** No oracle query changes. Operations happen in the same order.
- **I10 (ergs/native independence):** Access list and authorization list gas is now pre-deducted via `calculate_tx_intrinsic_gas` (from `gas_limit`), and runtime charges go against `intrinsic_tracker.with_infinite_ergs()`. No double-counting: intrinsic_gas reduces ergs at creation; runtime ergs charge hits infinite ergs (no-op).
- **I11 (L1 saturating policy):** L1 path uses `L1ResourcesPolicy`. `calculate_tx_intrinsic_gas(calldata_len, calldata_tokens, false, 0, 0, 0)` = `TX_INTRINSIC_GAS + calldata_tokens*4`, mathematically equivalent to old `L1_TX_INTRINSIC_L2_GAS + calldata_tokens*4`.
- **Feature isolation:** `verify_intrinsic_native` is test-only (`tests/rig/Cargo.toml`). Not in `production` or `for_tests` composite features. Assertions are properly `#[cfg]`-gated.

### Observations (non-blocking)

1. **Pubdata reporting increases by 68 bytes per L2 tx.** `L2_TX_INTRINSIC_PUBDATA` changed from 66 to 134. Since sender account pubdata (nonce + balance diff) is already captured in `net_pubdata_used()` before refund, the 68-byte `SENDER_BALANCE_INTRINSIC_PUBDATA` addition to the constant results in conservative over-reporting. This may cause slightly earlier block sealing under tight pubdata limits. The code comment at `zk/mod.rs:294` acknowledges this as a future optimization target.

2. **`compute_calldata_tokens` now returns 0 as `minimal_gas_used` when EIP-7623 is disabled.** Previously returned `L1_TX_INTRINSIC_L2_GAS` (21000), which was the base intrinsic, not the EIP-7623 floor. The new behavior is more correct — the EIP-7623 floor is 0 without the feature. L1 callers are unaffected (L1ResourcesPolicy saturates any gas shortfall).

### Suggested Tests

- A test with a large access list (many new accounts + storage keys) under `verify_intrinsic_native` to confirm the formula covers per-item native overhead for access list warming.
- A test with a service transaction under `verify_intrinsic_native` to validate `SERVICE_TX_INTRINSIC_COMPUTATIONAL_NATIVE_COST` covers actual consumption (no ecrecover, single keccak).
- An edge case test where `native_per_pubdata * L2_TX_INTRINSIC_PUBDATA` approaches `native_prepaid_from_gas` to verify the resource creation underflow path returns `OutOfNativeResourcesDuringValidation` correctly.

api/src/helpers.rs
```diff
@@ -372,3 +376,82 @@ pub fn encode_set_settlement_layer_chain_id_calldata(new_sl_chain_id: U256) -> V
...
+    native_price: U256,
+    pubdata_price: U256,
+    gas_limit: u64,
+    calldata_length: u64,
+    access_list_accounts: u64,
+    access_list_storage_keys: u64,
+    authorization_list_num: u64,
+    max_fee_per_gas: U256,
+    max_priority_fee_per_gas: U256,
+) -> Result<(), ()> {
```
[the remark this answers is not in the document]
0xVolosnikov: Actually a good point about `zk_ee::utils::u256_try_to_u64,` we import some stuff from zk_ee anyway

api/src/helpers.rs
0xVolosnikov: Pls add a task / todo

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
```diff
@@ -122,6 +121,20 @@ pub struct TxContextForPreAndPostProcessing<S: EthereumLikeTypes> {
...
     pub resources_before_refund: S::Resources,
+    /// Resources used to pay for system operations that are precharged by the
+    /// intrinsic computational native formula (e.g. ecrecover, keccak of the
+    /// signed/tx hash, account read, nonce write, fee prepayment, refund and
+    /// coinbase payment). In production this is `FORMAL_INFINITE` so call sites
+    /// behave exactly as before. With the `verify_intrinsic_native` feature it
+    /// starts at a finite `INTRINSIC_TRACKER_INITIAL_NATIVE` value so that the
+    /// actual native consumption can be recovered by subtracting the residual
+    /// and compared against the formula as an upper bound.
+    pub intrinsic_tracker: S::Resources,
```
0xVolosnikov: Have you considered using a compilation flag?

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
```diff
@@ -122,6 +121,20 @@ pub struct TxContextForPreAndPostProcessing<S: EthereumLikeTypes> {
     pub native_used: u64,
     pub initial_resources: S::Resources,
     pub resources_before_refund: S::Resources,
+    /// Resources used to pay for system operations that are precharged by the
+    /// intrinsic computational native formula (e.g. ecrecover, keccak of the
+    /// signed/tx hash, account read, nonce write, fee prepayment, refund and
+    /// coinbase payment). In production this is `FORMAL_INFINITE` so call sites
+    /// behave exactly as before. With the `verify_intrinsic_native` feature it
```
0xVolosnikov: > behave exactly as before

This part is not future-proof, may become confusing later

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
```diff
@@ -141,6 +154,7 @@ impl<S: EthereumLikeTypes> core::fmt::Debug for TxContextForPreAndPostProcessing
             .field("validation_pubdata", &self.validation_pubdata)
             .field("total_pubdata", &self.total_pubdata)
             .field("native_used", &self.native_used)
+            .field("intrinsic_tracker", &self.intrinsic_tracker)
```
0xVolosnikov: Note: `authorization_list_num` is missed here

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
AntonD3: Can you elaborate a bit, please? Do you mean that you are not happy with the implementation approach, or are you just about the comment?

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
0xVolosnikov: If we use it only for validation, maybe we should rename it to better reflect the meaning? Like native_resources_for_validation or smth

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
```diff
@@ -562,6 +579,45 @@ where
...
 
+        // Under `verify_intrinsic_native`, compare the native that was actually
+        // consumed for operations covered by the intrinsic computational native
+        // formula (everything charged through `intrinsic_tracker`, plus the
+        // per-site deltas accumulated during validation for the access list
+        // and authorization list) against the formula value. The formula must
+        // be an upper bound; otherwise a transaction could consume more native
+        // than was precharged.
+        #[cfg(feature = "verify_intrinsic_native")]
+        {
```
0xVolosnikov: Do you think it makes sense to move it into a separate function?

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
AntonD3: There is already compile time flag. It's used to init this variable(see gas_helpers), but I decided not to introduce many conditions across codebase, and this variable is actually used instead of infinite resources after validation. Though I don't have a strong opinion, can rework it if you think it will be better.

AntonD3: Not only during validation, but for all the computation which is considered intrinsic(post execution logic also)

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -140,19 +142,60 @@ where
         .ok_or(TxError::Validation(InvalidTransaction::PubdataPriceTooHigh))?;
     let native_prepaid_from_gas = native_per_gas.saturating_mul(tx_gas_limit);
 
-    // Now we will materialize resources, from which we will try to charge intrinsic cost on top
-    let mut tx_resources = create_resources_for_tx::<S, L2ResourcesPolicy>(
+    let mut access_list_accounts = 0;
+    let mut access_list_storage_keys = 0;
+    let mut authorization_list_num = 0;
```
0xVolosnikov: Pls fix the compilation warnings

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
0xVolosnikov: Maybe `intrinsic_resources` then?

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
0xVolosnikov: I'm talking about "as before" part of the comment

basic_bootloader/src/bootloader/constants.rs
0xVolosnikov: nit: you have some repeating blocks like `WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST + WARM_STORAGE_READ_NATIVE_COST + COLD_NEW_STORAGE_READ_NATIVE_COST // worst case account`

maybe adding a new constant would be useful here. Like: 

`const WORST_CASE_ACCOUNT_READ_COST: u64 = ...`

0xVolosnikov: Also we can introduce `const CACHE_CHANGE_NATIVE_COST: u64 = WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST + WARM_ACCOUNT_CACHE_WRITE_EXTRA_NATIVE_COST`. The name is not ideal though

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -28,9 +38,98 @@ pub const TX_CALLDATA_OFFSET: usize = 0x60;
...
+    WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST + WARM_ACCOUNT_CACHE_WRITE_EXTRA_NATIVE_COST + // balance change for fee prepayment
+    (WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST + WARM_ACCOUNT_CACHE_WRITE_EXTRA_NATIVE_COST) * 2 + keccak256_native_cost_for_rounds_u64(1); // post execution logic: transferring fee to coinbase, transferring the gas refund, hashing of tx hash into rolling hash
+
+/// Service tx intrinsic computational native cost.
+pub const SERVICE_TX_INTRINSIC_COMPUTATIONAL_NATIVE_COST: u64 =
+    WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST + WARM_STORAGE_READ_NATIVE_COST + COLD_NEW_STORAGE_READ_NATIVE_COST + // worst case account read
+    keccak256_native_cost_for_rounds_u64(2) + // keccak for full hash, 1 rounds worst case tx size + 1 round precharge for dynamic parts
+    WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST + WARM_ACCOUNT_CACHE_WRITE_EXTRA_NATIVE_COST + // balance change for fee prepayment
+    (WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST + WARM_ACCOUNT_CACHE_WRITE_EXTRA_NATIVE_COST) * 2 + keccak256_native_cost_for_rounds_u64(1); // post execution logic: transferring fee to coinbase, transferring the gas refund, hashing of tx hash into rolling hash
```
0xVolosnikov: But do we have any fee transfers here? Gas price is 0 for service txs

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -28,9 +38,98 @@ pub const TX_CALLDATA_OFFSET: usize = 0x60;
...
+
+/// L2 tx calldata byte intrinsic computational native cost.
+pub const SERVICE_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_CALLDATA_BYTE: u64 =
+    DYNAMIC_PART_KECCAK_COMPUTATIONAL_NATIVE_PER_BYTE; // to cover full hash
+
+/// Native computational cost to cover keccak256 hashing overhead for dynamic fields of the transaction per byte.
+/// NOTE: we are precharging 1 keccak round in the constant part since dynamic part, can consume 136*n + 1 bytes in encoding, so it will pay for ~n rounds, but consume (n + 1) rounds of keccak
+const DYNAMIC_PART_KECCAK_COMPUTATIONAL_NATIVE_PER_BYTE: u64 =
```
0xVolosnikov, line 99: The comment is a bit confusing because this constant is actually just a KECCAK_NATIVE_PER_BYTE

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -28,9 +38,98 @@ pub const TX_CALLDATA_OFFSET: usize = 0x60;
...
+    33 * DYNAMIC_PART_KECCAK_COMPUTATIONAL_NATIVE_PER_BYTE * 2; // keccak for signing + full hash, 33 contribution to rlp encoding length
+
+/// L2 tx authorization computational native cost.
+pub const L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_AUTHORIZATION: u64 =
+    PER_AUTH_NATIVE_COMPUTATIONAL_OVERHEAD + // computational overhead
+    keccak256_native_cost_for_rounds_u64(1) + // auth message keccak cost (1 round)
+    ECRECOVER_NATIVE_COST + // signature verification
+    WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST + WARM_STORAGE_READ_NATIVE_COST + COLD_NEW_STORAGE_READ_NATIVE_COST + // worst case account read
+    WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST + WARM_ACCOUNT_CACHE_WRITE_EXTRA_NATIVE_COST + // nonce update
+    WARM_ACCOUNT_CACHE_ACCESS_NATIVE_COST + WARM_ACCOUNT_CACHE_WRITE_EXTRA_NATIVE_COST + PREIMAGE_CACHE_SET_NATIVE_COST + keccak256_native_cost_for_rounds_u64(1) /*bytecode hashing */ + 1140 /* 1 round blake2s padded bytecode */ + // delegation write
```
0xVolosnikov: 1140 can be made a constant iiuc

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -81,6 +180,29 @@ pub const L1_TX_INTRINSIC_L2_GAS: u64 = 21_000;
 // first mint / call to L2AssetTracker can fail due to out-of-native
 pub const L1_TX_INTRINSIC_NATIVE_COST: u64 = 2_875_420;
 
+/// L1 tx calldata byte intrinsic computational native cost.
+// caldata per byte copy: COPY_BYTE_NATIVE_COST = 1
```
0xVolosnikov: Can we just use `COPY_BYTE_NATIVE_COST` then?

0xVolosnikov: Instead of "1" here I mean

basic_system/src/system_functions/keccak256.rs
```diff
@@ -33,6 +33,10 @@ pub fn keccak256_native_cost_u64(len: usize) -> u64 {
     (rounds as u64) * KECCAK256_ROUND_NATIVE_COST + KECCAK256_BASE_NATIVE_COST
```
0xVolosnikov: Can we reuse `keccak256_native_cost_for_rounds_u64` here?

basic_bootloader/src/bootloader/constants.rs
0xVolosnikov: Imo now it is much better, but still can be improved a bit

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
0xVolosnikov: So basically we will overcharge authorization lists and don't care too much?

vv-dev-ai:
[AI-review] Reviewed commit `f44bea288b65de26fddfed64421959037dc00903` — **APPROVE** (0 inline comments).

vv-dev-ai (commented):
[AI-review] PR #616 — **APPROVE**

Reply-review: evaluated 0xVolosnikov's feedback threads. No prior open issues.

Re: service tx fee transfers (constants.rs:82-86) — `refund_and_commit_fee` calls `update_account_nominal_token_balance` unconditionally for all tx types including service txs with gas_price=0. The balance diff is 0, but the account cache access/write native cost is still incurred. The constant is a valid upper bound, not over-budgeted for operations that don't execute — they do execute, just with zero amounts. That said, 0xVolosnikov's point is fair as a tightening opportunity: a future optimization could skip the balance write when the diff is zero, which would make the current constant unnecessarily large.

basic_bootloader/src/bootloader/constants.rs
AntonD3: Yes, service txs following the same flow as l2 txs, so amounts are 0, but there are still calls to the io and computational part is charged. I decided to not refactor service txs flow in this PR

basic_bootloader/src/bootloader/constants.rs
AntonD3: Yes, but the important thing here is that this constant should be used only for dynamic length tx fields(given that we are precharging one round in const intrinic part). Basically, I explained in the comment why, perhaps I can try to rephrase this comment if you think it makes sense. Or move this explanation to some other place.

basic_bootloader/src/bootloader/constants.rs
AntonD3: Perhaps then it makes sense to introduce blake2s native cost const function

basic_bootloader/src/bootloader/constants.rs
0xVolosnikov: Can you pls create a task in the task tracker for v0.4.0?

api/src/helpers.rs
```diff
@@ -372,3 +377,83 @@ pub fn encode_set_settlement_layer_chain_id_calldata(new_sl_chain_id: U256) -> V
...
+    // native_per_gas = ceil(gas_price / native_price)
+    if native_price.is_zero() {
+        return Err(());
+    }
+    let native_per_gas = u256_try_to_u64(gas_price.div_ceil(native_price)).ok_or(())?;
+
+    // native_per_pubdata = pubdata_price / native_price
+    let native_per_pubdata = u256_try_to_u64(pubdata_price.wrapping_div(native_price)).ok_or(())?;
+
+    let native_prepaid = native_per_gas.saturating_mul(gas_limit);
```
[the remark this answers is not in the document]
0xVolosnikov: @AntonD3 do you think we should fix it?

basic_bootloader/src/bootloader/constants.rs
AntonD3: I am planning to do it in the scope of my task about tx flow refactoring

0xVolosnikov: Ok, just wanted to make sure that we won't forget about it

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
AntonD3: I thought about it, but didn't want to make impression that this is intrinsic resources we actually charge

AntonD3: But maybe it's not so important as long as we have good comment about it, so I can rename it you believe it's better

0xVolosnikov: Yeah, imo it's ok if it looks like we charge resources inside the functions that actually use it - because we kinda do it in tests for example. And we can mention separately that in production we use infinite resources there. 

I'd prefer it this way because the "tracker" concept is a little bit confusing for me here - I'd prefer to provide a neutral API (resources) and allow the caller to decide how to use this component (infinite/non-infinite resources)

basic_bootloader/src/bootloader/constants.rs
0xVolosnikov: But can't it be used for any data? I mean it's always `ceil(len / KECCAK256_CHUNK_SIZE) * KECCAK256_ROUND_NATIVE_COST`  for any data, isn't it?

0xVolosnikov: Maybe I'm missing something

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
0xVolosnikov: "variable does not need to be mutable"

vv-dev-ai:
[AI-review] Reviewed commit `f44bea288b65de26fddfed64421959037dc00903` — **APPROVE** (0 inline comments).

vv-dev-ai (approved):
[AI-review] PR #616 — **APPROVE**

Re `wrapping_div` vs `div_ceil` for `native_per_pubdata` (api/src/helpers.rs:435): the API helper mirrors the bootloader's own computation at `validation_impl.rs:139`, which also uses `wrapping_div`. Changing this would be a design decision affecting both layers, not an API-only fix.

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
AntonD3: IMHO it looks ok, logic is relatively short, no duplication. But no strong opinion, LMK if you want me to create a function

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
AntonD3: should be resolved

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
0xVolosnikov: It just looks like a self-contained separate piece of logic, so imo it would be natural to move it to a separate function. But it is not very important

vv-dev-ai:
[AI-review] Reviewed commit `060d5b74c1c537fe3472e025745ba16434efd161` — **APPROVE** (0 inline comments).

vv-dev-ai (approved):
[AI-review] PR #616 — **APPROVE**

New commit `060d5b7` ("Minor review fixes") is cosmetic cleanup: symbolic constant `COPY_BYTE_NATIVE_COST` replacing hardcoded `1`, `authorization_list_num` binding cleaned up with proper `#[cfg(not(feature = "eip-7702"))]` fallback, keccak cost function deduplication. All changes are numerically equivalent to the prior code. No new concerns.

basic_bootloader/src/bootloader/constants.rs
AntonD3: It's not a round-based formula, so not `ceil(len / KECCAK256_CHUNK_SIZE) * KECCAK256_ROUND_NATIVE_COST`, but `len*ceil(KECCAK256_ROUND_NATIVE_COST/KECCAK256_ROUND_NATIVE_COST)`. But given that Keccak pricing depends on rounds, it's not the correct way to charge for a keccak256 call. This constant is used only for dynamic fields of tx, because for tx hash we call keccak once, and using this field, we'll kinda count the number of rounds for the total concatenation of the fields.

0xVolosnikov: But `len*ceil(KECCAK256_ROUND_NATIVE_COST/KECCAK256_CHUNK_SIZE)` essentially is an approximation of `ceil(len / KECCAK256_CHUNK_SIZE) * KECCAK256_ROUND_NATIVE_COST`, no? Let's discuss it tomorrow

api/src/helpers.rs
AntonD3: But do you think it will be worth it in the future?

AntonD3: IMHO it doesn't really matter, unless we are trying to fully reuse bootloader native validation logic(ideally as a single function)

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
AntonD3: Outdated

api/src/helpers.rs
AntonD3: Discussed in the call

basic_bootloader/src/bootloader/constants.rs
AntonD3: I tried to explain better in comments, IMHO name is good enough, I didn't came up with better one, PTAL

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
AntonD3: I am always using infinite now(and updated the comment), PTAL

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
AntonD3: we'll overcharge only if authorization failed - it's acceptable IMHO, perhaps we can fix by changing the authorization processing code, will consider in follow ups

api/src/helpers.rs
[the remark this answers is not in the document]
AntonD3: Yes, good catch, fixed

crypto/src/blake2s/mod.rs
```diff
@@ -22,6 +22,20 @@ mod delegated_extended;
 ))]
 pub use delegated_extended::{initialize_blake2s_delegation_context, Blake2s256};
 
+/// Native costs for blake2s hashing
```
0xVolosnikov: I'm not sure that it makes sense to have it here. Crypto crate will be moved to airbender-platform later + corresponding calculations for keccak live in basic_system/src/system_functions/keccak256.rs

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -437,40 +491,36 @@ where
...
-        let intrinsic_gas = intrinsic_gas.saturating_add(floor_tokens_gas_cost);
+        let intrinsic_gas = TX_INTRINSIC_GAS.saturating_add(floor_tokens_gas_cost);
 
         (num_tokens, intrinsic_gas)
     }
 
     #[cfg(not(feature = "eip_7623"))]
     {
-        (num_tokens, intrinsic_gas)
+        (num_tokens, 0)
```
0xVolosnikov, line 524: But why is it 0? Iiuc we can over-refund gas because of that

0xVolosnikov: Ok, seems like this is correct, just confusing

0xVolosnikov: @AntonD3 this part should be refactored in the next version

crypto/src/blake2s/mod.rs
0xVolosnikov: Addressed in https://github.com/matter-labs/zksync-os/pull/616/commits/97443a4ff0d99c1d955cab7a80369fec75d14c9f

0xVolosnikov (approved):

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
AntonD3: Why? It is following pre eip_7623, min_gas == 0 - no check, if I am not wrong it's implemented same way in revm

AntonD3: IMHO it was confusing before, over-refund(more than execution body) is not possible because of other guaranties, so practically floor gas pre eip_7623 can be any value in range [0, intrinsic]. But this function was using intrinsic_gas, and even more calculated incorrectly(no access/authorization list). So in my opinion it's better to have 0, which is basically natural "no floor gas"

PR #626, opened 2026-04-20
feat: fri precompile

Raid5594:
## What ❔

Adds FRI precompile to zksync-os.
FRI proofs submitted in the sidecar with FRI transactions are checked during transaction validation and the call to precompile within the same transaction reveals whether the proof is valid (under current assumptions, transactions that include invalid proofs are dropped from the block).

## Why ❔

The goal is to add a new tx type, that would allow GW users to submit FRI proofs along the transaction in the sidecar. This would in turn, enable faster interop.

## Is this a breaking change?
- [x] Yes
- [ ] No

## Checklist


basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
```diff
@@ -165,10 +173,108 @@ impl<S: EthereumLikeTypes> core::fmt::Debug for CachedPubdataInfo<S> {
...
+
+    let mut oracle_stream = alloc::vec::Vec::with_capacity(oracle_stream_len);
+    for raw_word in response {
+        let word = u32::try_from(raw_word)
+            .map_err(|_| TxError::Validation(InvalidTransaction::FriProofVerificationFailed))?;
+        oracle_stream.push(word);
+    }
+
+    nd_source_std::set_iterator(oracle_stream.into_iter());
+    let verification_result = catch_unwind(AssertUnwindSafe(|| {
```
antoniolocascio: Why `catch_unwind`?

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
```diff
@@ -165,10 +173,108 @@ impl<S: EthereumLikeTypes> core::fmt::Debug for CachedPubdataInfo<S> {
...
+
+#[cfg(not(target_arch = "riscv32"))]
+fn verify_fri_statement_host<S: EthereumLikeTypes>(
+    system: &mut System<S>,
+    statement_versioned_hash: Bytes32,
+) -> Result<(), TxError>
+where
+    S::IO: IOSubsystemExt,
+{
+    let mut response = system
```
antoniolocascio: Can't we use a higher level helper? We discussed before about separating data read from actual verifier work, is it doable?

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
```diff
@@ -165,10 +173,108 @@ impl<S: EthereumLikeTypes> core::fmt::Debug for CachedPubdataInfo<S> {
...
+        assert_ne!(baseline, changed_recursion_hash);
+
+        let mut hasher = Keccak256::new();
+        for word in output.iter() {
+            hasher.update(word.to_le_bytes());
+        }
+        let without_version = Bytes32::from_array(hasher.finalize());
+        assert_ne!(changed_recursion_hash, without_version);
+    }
+}
```
antoniolocascio: Let's move this and the fri verification logic to a new module

basic_bootloader/src/bootloader/fri_verifier.rs
```diff
@@ -0,0 +1,27 @@
+/// FRI verifier function pointer, registered at startup by `zksync_os/src/main.rs`.
```
antoniolocascio: I don't understand this indirection

tests/binary_checker/src/lib.rs
Raid5594, line 1: I believe we still need to use prover here and not transpiler

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -52,12 +54,29 @@ pub(crate) fn validate_and_compute_fee_for_transaction<
...
+    // `FriProofTx` transactions. A failure here drops the tx from the
+    // block entirely (no state change, no fee charge, no receipt in the
+    // block header). Running this first means that even the nonce /
+    // balance / fee checks below cannot commit work for a tx whose
+    // proofs we cannot accept.
+    //
+    // The verified hashes are installed on the tx-level metadata below
+    // when `set_tx_context` runs, so the FRI precompile can answer
+    // membership queries during the tx body execution.
+    let verified_fri_statements = if transaction.is_fri_proof() {
```
Raid5594: `is_fri_proof` => maybe worth replacing with has/includes_fri_proof?

evm_interpreter/src/precompile_addresses.rs
```diff
@@ -15,6 +15,17 @@ pub const BLAKE2F_HOOK_ADDRESS_LOW: u16 = 0x0009;
...
+/// Looks up a `statement_versioned_hash` in tx-scoped state and returns
+/// a boolean indicating whether it was verified earlier in the current
+/// tx. Always compiled in; registration at `add_precompiles` time is
+/// conditional on the runtime `is_gateway` flag.
+///
+/// Intentionally NOT in `PRECOMPILE_ADDRESSES_LOWS`: that list drives
+/// EVM semantics (cold-access discount via `is_precompile`), and FRI is
+/// a ZKsync-only hook with no EVM spec. Dispatch is via `HooksStorage`,
+/// independent of the precompile list.
+pub const FRI_PRECOMPILE_ADDRESS_LOW: u16 = 0x0101;
```
Raid5594: Remove LOW as it's not in LOWS?

tests/fixtures/fri/airbender_test_proof.bin
antoniolocascio, line 1: Do we need this?

basic_bootloader/src/bootloader/block_flow/ethereum/post_init_op.rs
```diff
@@ -10,7 +10,7 @@ where
         _system: &mut System<S>,
         system_functions: &mut HooksStorage<S, <S as SystemTypes>::Allocator>,
     ) -> Result<(), InternalError> {
-        add_precompiles(system_functions)?;
+        add_precompiles(system_functions, false)?;
```
antoniolocascio: I think a more generic config here would be nice. Initially it can be just `PrecompilesConfig {fri: bool}`. But we might want to add more to it soon (moving from compile-time to runtime configs)

zk_ee/src/system/metadata/basic_metadata.rs
```diff
@@ -54,6 +54,19 @@ pub trait BasicTransactionMetadata<IOTypes: SystemIOTypesConfig> {
...
+    /// more than one statement; this predicate lets callers check
+    /// membership without iterating the full list.
+    fn is_fri_statement_verified(&self, statement_versioned_hash: &Bytes32) -> bool;
+
+    /// Record that a FRI statement hash was verified for the current
+    /// transaction. Appends to the per-tx list; duplicates are
+    /// deliberately not deduplicated — a `FriProofTx` carrying a
+    /// duplicate hash in its `statement_versioned_hashes` list verifies
+    /// that statement twice, and the metadata reflects that intent.
+    fn push_verified_fri_statement(&mut self, statement_versioned_hash: Bytes32);
```
antoniolocascio: I'm not sure these are metadata ops, imo the metadata op would be akin to "give me proof object at index i", then we verify and storage somewhere in the system

zksync_os_runner/src/lib.rs
antoniolocascio, line 1: Not familiar with this part, but is this copied from aribender/platform repo? If so, can we deduplicate by making it more configurable?

Cargo.toml
```diff
@@ -63,12 +63,20 @@ keywords = ["blockchain", "zksync", "zk", "risc-v"]
...
 
 [workspace.dependencies]
-zksync_os_evm_errors = { version = "0.1.1", default-features =  false }
-zksync_os_interface = {  version = "0.1.1" }
+# TODO: switch to a tagged crates.io release once `is_gateway` lands
+# upstream. Until then we pin to the commit on `ra/add-gw-to-blockcontext`
+# that adds the field `BlockContext.is_gateway` used by FRI precompile
+# work.
+zksync_os_evm_errors = { git = "https://github.com/matter-labs/zksync-os-interface.git", rev = "a8785751b44a356bb6a70e147bec62b9ec2c8e42", default-features = false }
+zksync_os_interface = { git = "https://github.com/matter-labs/zksync-os-interface.git", rev = "a8785751b44a356bb6a70e147bec62b9ec2c8e42" }
```
antoniolocascio: Comment to cut a release before merging

system_hooks/src/lib.rs
```diff
@@ -202,6 +204,12 @@ where
             P256VerifyErrors,
         >(hooks, P256_VERIFY_PREHASH_HOOK_ADDRESS_LOW)?;
     }
+    if is_gateway {
+        hooks.add_call_hook(
+            FRI_PRECOMPILE_ADDRESS_LOW,
+            SystemCallHook::new(fri_precompile_hook),
+        )?;
+    }
```
antoniolocascio: Actually, I wouldn't add this here. add_precompiles is for EVM precompiles, this fri verification is just a generic system hook imo

basic_bootloader/src/bootloader/transaction_flow/zk/fri.rs
```diff
@@ -0,0 +1,259 @@
...
+///
+/// Returns `Ok(())` when the FRI verifier accepts the proof and the
+/// derived statement hash matches `statement_versioned_hash`. Maps any
+/// other outcome to the appropriate `InvalidTransaction` variant.
+///
+/// In the host path we read the proof stream from the sidecar and run
+/// the verifier directly here. In the RISC-V path the verifier runs
+/// in-circuit via the non-determinism CSR, so we only drain the oracle
+/// response here to keep the host-side query machinery in sync and
+/// then call the runtime entry point.
```
antoniolocascio: Why not unify this 2?

basic_bootloader/src/bootloader/transaction_flow/zk/fri.rs
```diff
@@ -0,0 +1,259 @@
...
+/// Host path: consume any oracle words the verifier left behind so the
+/// next tx starts with an empty non-determinism source.
+#[cfg(not(target_arch = "riscv32"))]
+fn drain_fri_verifier_iterator() -> usize {
+    let mut remaining = 0usize;
+    while nd_source_std::try_read_word().is_some() {
+        remaining += 1;
+    }
+    remaining
+}
```
antoniolocascio: This looks suspicious, why isn't it fully consumed?

basic_bootloader/src/bootloader/transaction_flow/zk/fri.rs
```diff
@@ -0,0 +1,259 @@
...
+        // abort the binary.
+        let mut response = system
+            .io
+            .oracle()
+            .raw_query(FRI_PROOF_QUERY_ID, &statement_versioned_hash)?;
+        response.next().ok_or(TxError::Validation(
+            InvalidTransaction::FriProofSidecarMissing,
+        ))?;
+        drop(response);
+        let output = crate::bootloader::fri_verifier::run_fri_verifier();
```
antoniolocascio: I think we can get rid of that module and just inline here

basic_bootloader/Cargo.toml
```diff
@@ -25,6 +25,12 @@ paste = { workspace = true }
 cfg-if = { workspace = true }
 serde-big-array = "0.5"
 
+[target.'cfg(not(target_arch = "riscv32"))'.dependencies]
+prover = { workspace = true, features = ["prover"] }
+full_statement_verifier = { workspace = true, default-features = false, features = ["unified_verifier_only_security_80", "blake2_with_compression", "replace_csr"] }
+
+[target.'cfg(target_arch = "riscv32")'.dependencies]
+full_statement_verifier = { workspace = true, default-features = false, features = ["unified_verifier_only_security_80", "modular_ops", "blake2_with_compression"] }
```
antoniolocascio: This seems like a full_statement_verifier concern

evm_interpreter/src/precompile_addresses.rs
antoniolocascio: I don't think this should even be here. It's not an evm precompile

basic_bootloader/src/bootloader/transaction/rlp_encoded/transaction_types/fri_proof_tx.rs
antoniolocascio, line 1: Can you add some tests parsing these, like we have for the other tx types?

oracle_provider/src/lib.rs
```diff
@@ -332,9 +332,30 @@ impl IOOracle for ReadWitnessSource {
...
+                } else {
+                    record_usize_as_u32_words(&mut read_items, *v);
+                    remaining_fri_payload_words = Some(*v);
+                }
+            }))
+        } else {
+            Box::new(inner.inspect(move |v| {
+                record_usize_as_u32_words(&mut read_items.borrow_mut(), *v);
+            }))
+        };
```
antoniolocascio: Why this change?

zk_ee/src/system/metadata/basic_metadata.rs
antoniolocascio: Actually the first one is ok, the second one should be just updated when we set the tx-level metadata, no? Like we do for blobs

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -52,12 +54,29 @@ pub(crate) fn validate_and_compute_fee_for_transaction<
...
+        super::fri::verify_all_fri_statements(system, transaction)?
```
antoniolocascio: I don't think we want to do this work at sequencing level, so this needs to be disabled.

tests/fixtures/fri/airbender_test_proof.bin
Raid5594: For e2e test

oracle_provider/src/lib.rs
Raid5594: Because FRI oracle responses use a packing where each payload usize carries two independent u32 verifier words (low | high << 32), and on odd-length payloads the last usize's high half is zero padding that the prover-side CSR reader does NOT consume. The generic recorder treats every usize as a (low, high) pair to record both halves; for FRI payload words we have to do the same EXCEPT skip the high half on the last word when the payload count is odd. The length prefix at the front of the response (recorded as a regular two-half usize) tells us how many verifier words to expect, which is how we know when we've hit the odd-truncation case.

basic_bootloader/src/bootloader/config.rs
```diff
@@ -4,6 +4,28 @@ pub trait BasicBootloaderExecutionConfig: 'static + Clone + Copy + core::fmt::De
...
+    /// and ETH-replay (`ForwardETHLikeConfig`) set this `false`.
+    /// Those paths trust the admission layer's FRI check and populate
+    /// `TxLevelMetadata.verified_fri_statements` directly from the
+    /// tx body's claimed list. The in-circuit verifier is the final
+    /// authority — a mismatch there fails the block proof, same as
+    /// any other bad witness.
+    ///
+    /// Mirrors the pattern `VALIDATE_EOA_SIGNATURE` uses for signature
+    /// checks: admission-layer work that the bootloader deliberately
+    /// skips during sequencing.
```
antoniolocascio: This is too LLM-verbose for me. Could be simplified to something like "Flag to disable FRI proof verification. Disabled for sequencing."

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
Raid5594: It follows the naming pattern for other tx types, so we can keep it as is

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
Raid5594: Resolved, as the code for host fri_verification was removed

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
Raid5594: resolved, as we now have different behaviour for host vs risc-v (host simply drains)

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
Raid5594: moved to `fri.rs`

basic_bootloader/src/bootloader/fri_verifier.rs
Raid5594: resolved, now FRI verifier is called directly in `fri.rs`:
`let output = full_statement_verifier::unified_circuit_statement::verify_unrolled_or_unified_circuit_recursion_layer();`

evm_interpreter/src/precompile_addresses.rs
Raid5594: resolved, moved to `address_constants.rs`

zk_ee/src/system/metadata/basic_metadata.rs
Raid5594: Resolved, FRI statements are set in `set_tx_context`.

zksync_os_runner/src/lib.rs
Raid5594: Right now we're taking `FullUnsignedMachineDecoderConfig` and setting up MOP support manually, however as discussed in slack we could export the shared config

Cargo.toml
Raid5594: Added comment to Cargo.toml

system_hooks/src/lib.rs
Raid5594: FRI precompile registration happens separately in `add_fri_proof_verification_hook`. (P.S. will probably rename and stick to precompile for FRI everywhere in naming)

basic_bootloader/src/bootloader/block_flow/ethereum/post_init_op.rs
Raid5594: Resolved as FRI precompile registered separately

basic_bootloader/src/bootloader/transaction_flow/zk/fri.rs
Raid5594: Resolved, as logic is different for host vs risc-v

basic_bootloader/src/bootloader/transaction_flow/zk/fri.rs
Raid5594: resolved, it was deleted

basic_bootloader/src/bootloader/transaction_flow/zk/fri.rs
Raid5594: Resolved, ran inline

basic_bootloader/Cargo.toml
Raid5594: Resolved, full statement verifier is only used for risc-v

basic_bootloader/src/bootloader/transaction/rlp_encoded/transaction_types/fri_proof_tx.rs
Raid5594: resolved, tests added

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
Raid5594: Resolved, sequencer skips verification now

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -52,12 +54,29 @@ pub(crate) fn validate_and_compute_fee_for_transaction<
...
+    } else {
+        arrayvec::ArrayVec::new()
+    };
```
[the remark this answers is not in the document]
Raid5594: resolved, reordered

basic_bootloader/src/bootloader/transaction_flow/zk/fri.rs
```diff
@@ -0,0 +1,259 @@
...
+            ));
+        }
+        for statement_versioned_hash in statement_versioned_hashes.iter() {
+            let statement_versioned_hash =
+                Bytes32::from_array(*statement_versioned_hash.map_err(TxError::Validation)?);
+            verify_fri_statement(system, statement_versioned_hash)?;
+            verified
+                .try_push(statement_versioned_hash)
+                .expect("cap already checked against MAX_FRI_STATEMENTS_PER_TX");
+        }
```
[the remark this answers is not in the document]
Raid5594: resolved, hash is verified only once

basic_bootloader/src/bootloader/transaction_flow/zk/fri.rs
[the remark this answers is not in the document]
Raid5594: Documented, not resolved, as malformed proof data in the guest still aborts the binary because there is no fallible verifier entry point.

forward_system/src/run/query_processors/fri_proof.rs
```diff
@@ -0,0 +1,175 @@
...
+        // Decode failure maps to an empty response, same as a missing
+        // sidecar entry. The bootloader rejects the transaction with
+        // `FriProofSidecarMissing`.
+        let bincode_config = bincode_v2::config::standard();
+        let Ok((proof, _)) = bincode_v2::serde::decode_from_slice::<UnrolledProgramProof, _>(
+            &proof_bytes,
+            bincode_config,
+        ) else {
+            return DynUsizeIterator::from_constructor(Vec::new(), |r| r.iter().copied());
+        };
```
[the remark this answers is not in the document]
Raid5594: Three different errors are now lgoged

zk_ee/src/system/metadata/basic_metadata.rs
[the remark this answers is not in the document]
Raid5594: Resolved, as it was removed

Raid5594:
> ## Code Review Summary
> **Reviewed by:** Claude + Codex (3 rounds each, 6 total review passes) **Findings:** 0 critical, 8 major, 6 minor, 4 nits
> 
> Findings were validated against the actual diff; several low-confidence automated findings were removed.
> 
> ### Critical
> > None.
> 
> ### Major
> 1. **Expensive FRI verification runs before cheap tx checks** — `basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs:74-78`. `verify_all_fri_statements` runs as the very first step, before nonce / balance / gas checks. A submitter can force up to 8 full FRI verifier runs per tx (and 80M native units of in-circuit work) on a tx that would trivially fail nonce/balance checks. The sequencer pays for the verifier work; the user is never charged because the tx is dropped. The in-code comment explains the rationale (drop tx without committing any state), but the DoS-amplification concern is not addressed. Consider moving FRI verification after the cheap structural/affordability checks that do not depend on the FRI precompile.
> 2. **Duplicate statement hashes are not deduplicated — 8× verifier amplification** — `basic_bootloader/src/bootloader/transaction_flow/zk/fri.rs:82-89`, `zk_ee/src/system/metadata/zk_metadata.rs:372-376` (unit test locks in duplicate preservation). Running the verifier N times for the same hash produces no additional information; the precompile's membership check treats duplicates identically. Dedup is O(cap) = O(8) and trivially cheap. Recommend rejecting at decode time (`DuplicateFriStatement`), or silently dedup. At minimum, add an integration test that pins whichever policy you choose.
> 3. **Cross-mode divergence on malformed proofs** — `basic_bootloader/src/bootloader/transaction_flow/zk/fri.rs:119-139` vs `196-212`. Host path catches verifier panics via `catch_unwind` and maps them to `FriProofVerificationFailed`. RISC-V path calls `run_fri_verifier()` with no unwind boundary. Today this is protected by the sequencer recording forward-mode output before handing to RISC-V, so if forward accepts, the recorded stream feeds RISC-V the same bytes and RISC-V succeeds. But this is an undocumented coupling: any future change that lets the RISC-V path see a stream the forward path didn't (e.g., a different sidecar source, a recorder bug) turns a graceful rejection into an unrecoverable STF panic. Worth either adding a fallible entry point in the verifier (the existing TODO at line 193-194) or documenting the invariant explicitly.
> 4. **`BasicTransactionMetadata::push_verified_fri_statement` is dead API surface on the production path** — trait declared at `zk_ee/src/system/metadata/basic_metadata.rs:69`, implemented at `zk_ee/src/system/metadata/zk_metadata.rs:49-56` and `basic_bootloader/src/bootloader/transaction_flow/ethereum/tx_level_metadata.rs:40` (no-op), but the real flow in `verify_all_fri_statements` builds an `ArrayVec` locally and installs it via `set_tx_context`. Only unit tests call the trait method. Either (a) route validation through `push_verified_fri_statement` so the Ethereum no-op stub actually enforces a "FRI disabled" invariant, or (b) drop the trait method and leave `is_fri_statement_verified` as the only trait addition.
> 5. **No RLP parser unit tests for `FriProofTx`** — `basic_bootloader/src/bootloader/transaction/rlp_encoded/transaction_types/fri_proof_tx.rs`. The precedent, `eip_4844_tx.rs` lines 78–249, has comprehensive tests: empty list, multiple entries, invalid element length, bad `to` length, full-tx round-trip parse. `fri_proof_tx.rs` has zero tests. This is new consensus-relevant parsing code — replicate the EIP-4844 patterns.
> 6. **No negative test for `FriProofStatementHashMismatch`** — `tests/instances/system_hooks/src/lib.rs`. Of the five FRI-related error variants (`FriProofTxNotSupported`, `FriProofSidecarMissing`, `FriProofVerificationFailed`, `FriProofStatementHashMismatch`, `TooManyFriStatements`), only the first two have coverage. `FriProofStatementHashMismatch` is the core security property of the feature (a submitter claiming a different hash than what was proven). Add: (a) valid proof + perturbed `statement_versioned_hash` → `FriProofStatementHashMismatch`, (b) corrupted proof bytes → `FriProofVerificationFailed`, (c) 9-hash list → `TooManyFriStatements`.
> 7. **Empty oracle response conflates three distinct failure modes** — `forward_system/src/run/query_processors/fri_proof.rs:77-102`. (a) sidecar has no entry for this hash, (b) `artifacts: None` (sequencer misconfig — gateway chain forgot to wire verifier artifacts), (c) bincode decode failure (corrupted sidecar) all produce `Vec::new()` → `FriProofSidecarMissing`. Cases (b) and (c) are sequencer bugs that will silently reject all FRI txs for the duration of the misconfig. At minimum add a `tracing::error!` on each of the three sites so operators can grep for the cause; ideally distinguish "sidecar missing" (user retries) from "sequencer bug" (operational alert).
> 8. **Centralize the FRI low/high packing format** — producer at `forward_system/src/run/query_processors/fri_proof.rs:110-116`, host consumer at `basic_bootloader/src/bootloader/transaction_flow/zk/fri.rs:172-178`, record/replay at `oracle_provider/src/lib.rs:335-358`, CSR path at `proof_running_system/src/io_oracle/mod.rs:91-137`. Three open-coded implementations of the same u32-pair packing across three crates. A shared helper (`pack_u32_pairs` / `unpack_u32_pairs`) with one invariant doc block would make this a single source of truth and give the cross-implementation round-trip test (currently missing) a natural home.
> 
> ### Minor
> 1. **Simulation path does not bypass FRI verification** — `validation_impl.rs:74-78`. Other validation branches (nonce, balance, fee) are gated on `Config::SIMULATION`, but the FRI check is unconditional. Any `eth_call` / `estimate_gas`-style simulation of a `FriProofTx` will hit the oracle and fail with `FriProofSidecarMissing`. Either skip FRI verification and short-circuit `is_fri_statement_verified` to `true` under simulation, or document explicitly that simulation of this tx type is unsupported.
> 2. **`FriProofSidecarMissing → InvalidStructure` is semantically incorrect** — `forward_system/src/run/convert.rs:151`. "Missing sidecar" is not an invariant of the signed RLP body; the body itself is well-formed. RPC callers seeing `InvalidStructure` assume the tx is permanently bad and will not retry; a user whose sidecar upload raced the tx gossip could retry successfully. Consider `FilteredByValidator` (matching `FriProofTxNotSupported`) or a dedicated `SidecarUnavailable` variant.
> 3. **`BlockMetadataFromOracle` serialization change has no version guard** — `zk_ee/src/system/metadata/zk_metadata.rs:245-313`. Appending `is_gateway` to `UsizeSerializable` silently reinterprets old oracle dumps. The PR context calls this out as intentional (all gateway chains ship together), but a format version byte / length guard would turn silent corruption of a stale dump into a clean error on replay. Cheap insurance.
> 4. **Airbender dependencies pinned to `branch = "dev"`** — `Cargo.toml`, `basic_bootloader/Cargo.toml`. Pin exact commit hashes before merge. This is consensus/proving code; moving targets make builds non-reproducible and let verifier behavior change without a repo diff.
> 5. **Workspace-wide `profile.dev.debug = false` / `profile.test.debug = false`** — root `Cargo.toml`. Workaround for an LLVM/DWARF SIGBUS from the Airbender dep graph. Disabling debug info workspace-wide degrades every contributor's debugging experience; scope the override to the crate that triggers the bug.
> 6. **Test fixture scaffolding duplicated** — `tests/instances/system_hooks/src/lib.rs`. `mod fri_precompile` and `mod fri_precompile_e2e` each redefine `FRI_PRECOMPILE`, `FRI_STATEMENT_HASH_VERSION`, the tx-builder boilerplate, and the output-matching arms (~150 lines). Extract a helper (`build_fri_proof_tx`, `assert_precompile_returned`) so the negative tests recommended under Major [feat: pubdata compression #6](https://github.com/matter-labs/zksync-os/pull/6) are a handful of lines each.
> 
> ### Nits
> 1. **Large binary committed to repo root** — `matter-labs_b18507c4-50f3-4638-854a-ed625c7e685a_11024243.bin` (~1.1 MB). Move under `tests/fixtures/fri/` alongside the existing setup/layout fixtures, or use git-lfs / fetch-on-test.
> 2. **`FRI_PRECOMPILE_ADDRESS_LOW` naming** — `evm_interpreter/src/precompile_addresses.rs:28`. The `_LOW` suffix suggests membership in `PRECOMPILE_ADDRESSES_LOWS`; this constant is intentionally excluded (hook, not precompile). The doc comment explains it, but a name like `FRI_PRECOMPILE_HOOK_ADDRESS` avoids the friction.
> 3. **`fri_verifier.rs` shim has no documentation** — `basic_bootloader/src/bootloader/fri_verifier.rs`. A one-line doc comment explaining the indirection (so the RISC-V path can link against the verifier symbol without pulling `full_statement_verifier` into the proving binary's graph) would answer Raid5594's inline review comment permanently.
> 4. **`is_fri_proof` naming** — `basic_bootloader/src/bootloader/transaction/mod.rs:120`. Raid5594 suggested `has_fri_proof`. The current name is consistent with sibling predicates (`is_service`, `is_l1_l2`, `is_upgrade`), all of which answer "is this tx of type X?". Keeping is-is defensible; worth replying on the PR to close the thread.
> 
> ### Non-issues investigated and dropped
> * **`is_gateway` as a consensus attack vector** (Codex flagged as critical): `BlockMetadataFromOracle` is the established delivery path for all consensus-critical block metadata including `chain_id`, `block_number`, `coinbase`, etc. Binding to public inputs happens at the block commitment boundary outside this PR's scope. `is_gateway` is no more/less trusted than the surrounding fields.
> * **Unbounded allocation from sidecar-reported `oracle_stream_len`** (Codex/Claude R2 flagged as major): The sidecar is sequencer-controlled, not user-controlled. A malicious oracle is a sequencer bug, not a user-facing DoS vector.
> * **`nd_source_std::set_iterator` could panic on non-empty previous iterator**: the drain happens after `catch_unwind`, always consuming leftover words before the next call. Not currently reachable.
> * **Intrinsic-native overcharge bypass**: confirmed safe — `statement_versioned_hashes_num` is only populated on the success path.
> * **Host path length check vs div_ceil(2)**: verified correct for all reachable `oracle_stream_len` values.
> 
> _Automated review by Claude Code (Opus 4.7) + Codex CLI_

Major
1. Expensive FRI verification before cheap tx checks. ✅ Resolved. Sequencer/simulation/replay configs skip the verifier entirely.

2. Duplicate statement hashes — 8× verifier amplification. ✅ Resolved. Silent dedup in build_verified_fri_statements_list; billing follows raw count (submitter pays as submitted).

3. Cross-mode divergence on malformed proofs. ✅ Resolved by architectural change. Host no longer verifies. RISC-V still aborts on bad input — that's a property of the airbender verifier (no fallible entry point). 

4. push_verified_fri_statement dead API. ✅ Resolved by removal. Trait now only exposes is_fri_statement_verified.

5. No RLP parser unit tests for FriProofTx. ✅ Resolved. 5 tests added mirroring the EIP-4844 patterns at `fri_proof_tx.rs`.

6. No negative test for FriProofStatementHashMismatch / corrupted proof / 9-hash list. ⚠️ Partially resolved. Hash mismatch and corrupted proof tests were removed when forward-mode FRI verification was deleted. Adding them requires proving-mode rig infrastructure — open follow-up, not merge-blocking, or add a commit in this PR.

7. Empty oracle response conflates failure modes. ✅ Resolved. Three distinct log::debug! / log::error! sites in FriProofResponder so operators can check the cause.

8. Centralize FRI low/high packing format. ✅ Resolved. Shared helpers + spec doc + round-trip test.

Minor
1. Simulation path doesn't bypass FRI verification. ✅ Resolved. VERIFY_FRI_PROOFS = false on BasicBootloaderCallSimulationConfig — simulation skips oracle/verifier work.

2. FriProofSidecarMissing → InvalidStructure semantically wrong. ✅ Resolved. Mapped to FilteredByValidator along with the other transient FRI variants; TooManyFriStatements correctly stays InvalidStructure.

3. BlockMetadataFromOracle no version guard. ❌ Will oppen follow-up PR, as it applies to every metadata field addition, not FRI-specific.

4. Airbender deps on branch = "dev". ⚠️ Partially resolved. Pinned to `591efe90...` But airbender-platform stll uses airbender branch `dev`.

5. Workspace-wide profile.dev.debug = false. ❌ Will open follow-up PR.

6. Test fixture scaffolding duplicated. ❌ Open follow-up. The remaining duplication is much smaller now that the forward-mode negative tests were removed. Worth a small cleanup PR.

Nits
1. Large binary committed at repo root. ⚠️ Will move in next commit and rename

2. FRI_PRECOMPILE_ADDRESS_LOW naming. ✅ Resolved. Constant moved out of evm_interpreter::precompile_addresses (where the _LOW suffix was misleading) and into system_hooks::addresses_constants (where _LOW is the consistent local convention — L1_MESSENGER_ADDRESS_HOOK_LOW, MINT_HOOK_ADDRESS_LOW, etc.).

3. fri_verifier.rs shim doc. ✅ Resolved by deletion.

4. is_fri_proof naming. ❌ No-action. Matches sibling convention (is_service, is_l1_l2, is_upgrade).

Cargo.toml
antoniolocascio: Sorry, it was just a reminder to not resolve this thread until we do cut the release. No need to add a comment in the code

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
antoniolocascio: As discussed, it should be disabled for sequencing, but not removed from forward run

basic_bootloader/Cargo.toml
antoniolocascio: No, what I mean is that having features selected based on arch for a dependency is a bit strange. But we can ignore

basic_bootloader/src/bootloader/block_flow/zk/post_init_op.rs
```diff
@@ -1,16 +1,19 @@
...
+    S::Metadata: GatewayModeMetadata,
 {
     fn post_init_op<Config: BasicBootloaderExecutionConfig>(
-        _system: &mut System<S>,
+        system: &mut System<S>,
         system_functions: &mut HooksStorage<S, <S as SystemTypes>::Allocator>,
     ) -> Result<(), InternalError> {
         system_hooks::add_precompiles(system_functions)?;
 
+        // TODO: maybe rename
```
antoniolocascio, line 18: Rename what? `is_gateway`? I think it's ok

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -140,6 +143,12 @@ where
     let native_per_pubdata = u256_try_to_u64(&pubdata_price.wrapping_div(native_price))
         .ok_or(TxError::Validation(InvalidTransaction::PubdataPriceTooHigh))?;
     let native_prepaid_from_gas = native_per_gas.saturating_mul(tx_gas_limit);
+    let statement_versioned_hashes_num = transaction
+        .statement_versioned_hashes()
+        .map(|hashes| hashes.count as u64)
+        .unwrap_or(0);
+    let fri_proof_intrinsic_native_cost =
+        FRI_PROOF_INTRINSIC_NATIVE_COST_PER_PROOF.saturating_mul(statement_versioned_hashes_num);
```
antoniolocascio: This should be included in the helper used to compute intrinsic_computational_native and also in the api. CC: @AntonD3

cycle_marker/src/lib.rs
```diff
@@ -296,9 +296,9 @@ pub fn revert(snap: Snapshot) {
 #[cfg(target_arch = "riscv32")]
 pub fn revert(_: Snapshot) {}
 
-/// Re-export `airbender_host` cycle marker types for use by the runner.
+/// Re-export `CycleMarker` from `riscv_transpiler::cycle`.
 #[cfg(feature = "use_riscv_transpiler")]
-pub use airbender_host::{CycleMarker, Mark};
+pub use riscv_transpiler::cycle::{CycleMarker, Mark};
```
antoniolocascio: This looks like a bad merge, no?

cycle_marker/Cargo.toml
```diff
@@ -10,8 +10,8 @@ keywords.workspace = true
 categories.workspace = true
 
 [dependencies]
-airbender-host = { workspace = true, optional = true }
+riscv_transpiler = { workspace = true, optional = true }
```
antoniolocascio: Why?

forward_system/src/run/interface_impl.rs
```diff
@@ -49,6 +49,8 @@ impl RunBlock for RunBlockForward {
             storage,
             preimage_source,
             tx_source,
+            NoFriProofSidecar,
+            None,
```
antoniolocascio: Shouldn't we add these as arg?

forward_system/src/run/mod.rs
```diff
@@ -291,12 +316,45 @@ pub fn make_oracle_for_proofs_and_dumps<T: ReadStorageTree, PS: PreimageSource,
...
     add_uart: bool,
     use_native_callable_oracles: bool,
+) -> ZkEENonDeterminismSource {
+    make_oracle_for_proofs_and_dumps_with_fri_sidecar(
+        block_context,
+        tree,
+        preimage_source,
+        tx_source,
+        NoFriProofSidecar,
+        None,
```
antoniolocascio: Why introduce a new function rather than adding the new args to this one?

tests/common/Cargo.toml
```diff
@@ -11,7 +11,14 @@ categories.workspace = true
...
 alloy = { workspace = true, features = [
-    "serde",
+    "consensus",
+    "dyn-abi",
+    "eips",
+    "network",
+    "rlp",
+    "rpc",
+    "rpc-types",
+    "signers",
```
antoniolocascio: Surprised we need to add them now

tests/rig/src/fri.rs
```diff
@@ -0,0 +1,173 @@
...
+/// sidecar resolution.
+///
+/// It mirrors the server/runtime boundary at the point where the
+/// bootloader resolves a `statement_versioned_hash` by issuing a
+/// `FRI_PROOF_QUERY_ID` oracle query.
+#[derive(Debug, Clone, Default)]
+pub struct FriProofOracleFactory<const RANDOMIZED_TREE: bool> {
+    sidecar_source: InMemoryFriProofSidecarSource,
+    verifier_artifacts: Option<FriVerifierArtifacts>,
+}
```
antoniolocascio: I think this can be simplified if we don't create a new method for making an oracle

tests/rig/src/fri.rs
```diff
@@ -0,0 +1,173 @@
...
+/// `FriProofResponder`.
+///
+/// The source keeps a shared `lookup_count` that tests can read to
+/// assert how many times the bootloader issued a `FRI_PROOF_QUERY_ID`
+/// query against it — useful for pinning validator-level dedup
+/// behavior. The counter is `Arc<AtomicUsize>` so it survives the
+/// `Clone` that `FriProofOracleFactory` performs when constructing the
+/// forward-mode oracle.
+#[derive(Debug, Clone, Default)]
+pub struct InMemoryFriProofSidecarSource {
```
antoniolocascio, line 24: Normally we define such structs in the forward system, but here might be ok too

zksync_os_runner/src/lib.rs
```diff
@@ -1,6 +1,36 @@
-use airbender_host::{FlamegraphConfig, Program, Runner as _};
+//! ZKsync OS RISC-V runner.
+//!
+//! TODO(airbender): once `TranspilerRunner` accepts a generic decoder
```
antoniolocascio: Let's do it now, no?

zk_ee/src/oracle/fri_proof_packing.rs
```diff
@@ -0,0 +1,119 @@
+//! On-the-wire packing format for `FRI_PROOF_QUERY_ID` responses.
```
antoniolocascio: I think this can be simplified a lot. Look at the latest commit here: `alocascio-fri-precompile-draft`

basic_bootloader/src/bootloader/block_flow/zk/post_init_op.rs
Raid5594: It suggests that `disable_system_contracts` might be a bit too broad

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
Raid5594: Will be resolved in follow-up PR

zk_ee/src/oracle/fri_proof_packing.rs
Raid5594: resolved

forward_system/src/run/interface_impl.rs
Raid5594: resolved

forward_system/src/run/mod.rs
Raid5594: resolved

tests/common/Cargo.toml
Raid5594: removed unused ones

tests/rig/src/fri.rs
Raid5594: resolved

tests/rig/src/fri.rs
Raid5594: kept it in rig

zksync_os_runner/src/lib.rs
Raid5594: https://github.com/matter-labs/airbender-platform/pull/63/changes

Awaiting this PR and can be resolved

tests/common/Cargo.toml
antoniolocascio: Can you explain why this PR adds all those?

tests/rig/src/chain.rs
```diff
@@ -88,8 +90,12 @@ pub trait TestingOracleFactory<const RANDOMIZED_TREE: bool> {
...
 }
 
-/// Default oracle factory that uses the existing make_oracle_for_proofs_and_dumps function
-pub struct DefaultOracleFactory<const RANDOMIZED_TREE: bool>;
+/// Default oracle factory used by normal rig runs.
+#[derive(Clone, Default)]
+pub struct DefaultOracleFactory<const RANDOMIZED_TREE: bool> {
+    pub fri_sidecar: Option<crate::fri::InMemoryFriProofSidecarSource>,
+    pub fri_artifacts: Option<Arc<FriVerifierArtifacts>>,
+}
```
antoniolocascio: Why add them as fields here instead of args in the `create_` functions?

Raid5594: If we put them in interface, then we'll have to import `zskync-airbender` in the interface, so I wanted to avoid this dep

cycle_marker/src/lib.rs
Raid5594: resolved

zksync_os_runner/src/lib.rs
Raid5594: resolved

cycle_marker/Cargo.toml
Raid5594: resolved

tests/common/Cargo.toml
Raid5594: resolved

zk_ee/src/system/metadata/basic_metadata.rs
```diff
@@ -68,6 +73,11 @@ pub trait ZkSpecificPricingMetadata {
     fn get_pubdata_price(&self) -> U256;
 }
 
+/// Runtime block-scoped flag for Gateway-specific features.
+pub trait GatewayModeMetadata {
```
AntonD3: Why separate trait, but not just block metadata?

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -140,6 +143,12 @@ where
     let native_per_pubdata = u256_try_to_u64(&pubdata_price.wrapping_div(native_price))
         .ok_or(TxError::Validation(InvalidTransaction::PubdataPriceTooHigh))?;
     let native_prepaid_from_gas = native_per_gas.saturating_mul(tx_gas_limit);
+    let statement_versioned_hashes_num = transaction
+        .statement_versioned_hashes()
+        .map(|hashes| hashes.count as u64)
+        .unwrap_or(0);
+    let fri_proof_intrinsic_native_cost =
```
AntonD3: IMHO we need to charge gas as well

basic_bootloader/src/bootloader/block_flow/zk/post_init_op.rs
antoniolocascio: Let's remove the TODO or make explicit. I would remove

tests/rig/src/chain.rs
antoniolocascio: In what interface? `TestingOracleFactory` is part of this same module, so it has the dep already, no?

forward_system/src/run/fri_proof_decode.rs
```diff
@@ -0,0 +1,36 @@
+//! Shared bincode-decode + flatten helper for FRI proof bytes.
+
+use crate::run::query_processors::FriVerifierArtifacts;
+use execution_utils::unified_circuit::flatten_proof_into_responses_for_unified_recursion;
+use execution_utils::unrolled::UnrolledProgramProof;
+
+/// Errors returned by [`decode_and_flatten_proof`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DecodeAndFlattenError {
```
antoniolocascio: This seems a bit over engineered, let's make decode_and_flatten_proof return an option

tests/rig/src/lib.rs
```diff
@@ -330,6 +348,83 @@ impl<const RANDOMIZED_TREE: bool> TestingFramework<RANDOMIZED_TREE> {
...
+        }));
+        self
+    }
+
+    /// Same as [`Self::with_mock_fri_sidecars_and_artifacts`] but also
+    /// returns a shared counter that reports how many times the
+    /// bootloader queried the sidecar source for proof bytes during
+    /// this block's execution. Intended for tests that need to pin
+    /// validator-level dedup behavior.
+    pub fn with_mock_fri_sidecars_and_artifacts_with_counter<I>(
```
antoniolocascio: Let's unify these three helpers to one that always returns the counter and takes the artifacts as an option.

antoniolocascio:
This PR needs to be documented:
- system_hooks.md should mention it (saying it's GW gated)
- transaction_format.md should document the new tx type (again, mentioning that it's GW gated)
- oracles.md should document the new oracle query
- We need a new doc page for the fri precompile imo. Essentially the design doc. It should help reviewers understand the entire picture

basic_bootloader/src/bootloader/transaction_flow/zk/fri.rs
```diff
@@ -0,0 +1,320 @@
+//! FRI proof handling for `FriProofTx` transactions.
```
antoniolocascio, line 1: Feels a bit too complicated to me still. Too many target cfg. I also don't like adding a host module for verification to bootloader codebase which isn't actually used in bootloader. I think something like this would work better: https://github.com/matter-labs/zksync-os/pull/655
Basically:
- Unifies some duplicated logic (hash computation for example)
- Cleaner separation of guest/host
- Run verification in prover input generation run

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
Raid5594: resolved

forward_system/src/run/fri_proof_decode.rs
Raid5594: resolved

tests/rig/src/lib.rs
Raid5594: resolved

tests/rig/src/chain.rs
Raid5594: resolved

basic_bootloader/src/bootloader/transaction_flow/zk/fri.rs
Raid5594: resolved

docs/fri_precompile.md
```diff
@@ -0,0 +1,217 @@
+# FRI precompile (Gateway-only)
+
+The FRI precompile lets contracts on a Gateway-mode chain check, from
+inside a transaction, whether a specific `statement_versioned_hash`
+is in the current tx-scoped verified-statements list. The list is
+populated by `FriProofTx` validation; actual proof verification
+happens upstream in server, host and proving paths. The proof bytes
```
antoniolocascio: Makes it sound that the proof verification isn't part of the STF, even if it says "proving paths". Let's say that the verification is part of the tx validation

Raid5594: resolved

zk_ee/src/system/metadata/basic_metadata.rs
Raid5594: resolved

zksync_os_runner/src/lib.rs
```diff
@@ -1,8 +1,26 @@
...
+//! in our MOP-aware decoder.
+
+use airbender_host::{ExecutionResult, FlamegraphConfig, Program, Runner as _};
+use riscv_transpiler::ir::{DecodingOptions, FullUnsignedMachineDecoderConfig};
 use std::path::PathBuf;
 
+/// Decoder config used by the FRI-aware RISC-V runner.
+struct FullUnsignedMachineWithMopDecoderConfig;
+
+impl DecodingOptions for FullUnsignedMachineWithMopDecoderConfig {
```
AntonD3, line 13: Struggling to understand this part - does it mean that we have to run zksync os on some non-default risc-v simulator? Is it supported by default by our prover?

zksync_os/.cargo/config.toml
```diff
@@ -1,7 +1,7 @@
 [build]
 target = "riscv32im-risc0-zkvm-elf"
 rustflags = [
-  "-C", "target-feature=+m,-unaligned-scalar-mem,+relax",
+  "-C", "target-feature=+m,-unaligned-scalar-mem,+relax,+zimop",
```
AntonD3, line 4: Do we need it for delegations? I thought we use `csrw` for delegations

zk_ee/src/system/metadata/basic_metadata.rs
```diff
@@ -38,6 +38,9 @@ pub trait BasicBlockMetadata<IOTypes: SystemIOTypesConfig> {
 
     /// Base fee per blob gas*(EIP-4844), if supported.
     fn blob_base_fee_per_gas(&self) -> U256;
+
+    /// Runtime block-scoped flag for Gateway-specific features.
+    fn is_gateway(&self) -> bool;
```
AntonD3: Don't want to spend much more time on this, if you are going to change it in a follow up - feel free to ignore the comment.
But I thought that during latest discussion we agreed to add this flag to `ZkSpecificPricingMetadata`(and rename to something like `ZkSpecificMetadata`

system_hooks/src/addresses_constants.rs
```diff
@@ -30,6 +30,13 @@ pub const L2_BASE_TOKEN_ADDRESS: B160 = B160::from_limbs([L2_BASE_TOKEN_ADDRESS_
 pub const MINT_HOOK_ADDRESS_LOW: u16 = 0x7100;
 pub const MINT_HOOK_ADDRESS: B160 = B160::from_limbs([MINT_HOOK_ADDRESS_LOW as u64, 0, 0]);
 
+// FRI proof verification hook (Gateway only, runtime-gated by `metadata.is_gateway`).
+// Looks up a `statement_versioned_hash` in tx-scoped state and returns a boolean
+// indicating whether it was verified earlier in the current tx.
+pub const FRI_PRECOMPILE_ADDRESS_LOW: u16 = 0x0101;
```
AntonD3: How did you come up with this address? Perhaps not really important, but there is an RIP with a pecompile at this address [RIP-7728](https://ethereum-magicians.org/t/rip-7728-l1sload-precompile/20388)

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -140,6 +140,12 @@ where
...
```
AntonD3: better to move it inside calculate_l2_tx_intrinsic_computational_native_resources

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -51,6 +51,14 @@ pub const TX_INTRINSIC_GAS: u64 = 21_000;
 /// Extra cost for deployment transactions.
 pub const DEPLOYMENT_TX_EXTRA_INTRINSIC_GAS: u64 = 32_000;
 
+/// FRI proof verification cost.
+pub const FRI_PROOF_INTRINSIC_NATIVE_COST_PER_PROOF: u64 = 10_000_000;
+/// Flat intrinsic gas surcharge for a `FriProofTx` carrying at least one
```
AntonD3: not true, according to the code, it's amount charged per statement

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
```diff
@@ -940,7 +947,12 @@ where
         // present: failed auths (bad sig, wrong chain id, nonce overflow)
         // consume only PER_AUTH_NATIVE_COMPUTATIONAL_OVERHEAD while the
         // formula budgets worst-case success cost per entry.
-        if context.authorization_list_num == 0 {
+        //
+        // Also skip when FRI statement hashes are present: the budget
+        // covers the airbender unified verifier on RISC-V, which is
+        // much more expensive than the host-mode verifier this check
+        // runs under.
+        if context.authorization_list_num == 0 && context.statement_versioned_hashes_num == 0 {
```
AntonD3: I don't understand why we should special case it on `statement_versioned_hashes_num`, I think you may had failures because you didn't properly account proofs native cost into intrinsic(mentioned in another comment)

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
AntonD3: And btw you need to charge separately from `let mut intrinsic_resources = S::Resources::FORMAL_INFINITE;`, it's used to validate formula, you can check signature as an example

basic_bootloader/src/bootloader/fri_verifier.rs
```diff
@@ -0,0 +1,386 @@
...
+        let op_type = DefaultNonDeterminismSource::read_word();
+        if op_type != OP_VERIFY_UNIFIED_RECURSION_LAYER_IN_UNIFIED_CIRCUIT {
+            return Err(TxError::Validation(
+                InvalidTransaction::FriProofVerificationFailed,
+            ));
+        }
+
+        Ok(
+            full_statement_verifier::unified_circuit_statement::verify_unified_circuit_recursion_layer(
+                full_statement_verifier::verifier_common::SecurityModel::Security80,
```
AntonD3: Do I understand correctly that it will read proof from non-determenism source?

forward_system/src/run/convert.rs
```diff
@@ -146,6 +147,11 @@ impl IntoInterface<InvalidTransaction>
             basic_bootloader::bootloader::errors::InvalidTransaction::NativeResourcesAreTooExpensive => {InvalidTransaction::NativeResourcesAreTooExpensive}
             basic_bootloader::bootloader::errors::InvalidTransaction::EIP7623IntrinsicGasIsTooLow => {InvalidTransaction::EIP7623IntrinsicGasIsTooLow}
             basic_bootloader::bootloader::errors::InvalidTransaction::CallerGasLimitMoreThanTxLimit => {InvalidTransaction::CallerGasLimitMoreThanTxLimit}
+            basic_bootloader::bootloader::errors::InvalidTransaction::FriProofTxNotSupported => {InvalidTransaction::FilteredByValidator}
```
AntonD3: Why FilteredByValidator?

basic_bootloader/src/bootloader/fri_verifier.rs
Raid5594: Yes, consumed via `DefaultNonDeterminismSource::read_word()`.

zksync_os/.cargo/config.toml
Raid5594: we need it for MOP instructions

zksync_os_runner/src/lib.rs
Raid5594: No, it's just that we don't export every possible config from zksync-airbender, so it's up to consumer to decide. here we simply set MOP to true

forward_system/src/run/convert.rs
Raid5594: resolved

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
Raid5594: removed

basic_bootloader/src/bootloader/constants.rs
Raid5594: resolved

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
Raid5594: resolved

system_hooks/src/addresses_constants.rs
Raid5594: changed to `7003`

zk_ee/src/system/metadata/basic_metadata.rs
Raid5594: resolved

zksync_os/.cargo/config.toml
AntonD3: But which MOP instructions?

zksync_os_runner/src/lib.rs
AntonD3: but is these MOP instructions we use supported by the prover?

basic_bootloader/Cargo.toml
```diff
@@ -25,6 +25,14 @@ paste = { workspace = true }
 cfg-if = { workspace = true }
 serde-big-array = "0.5"
 
+[target.'cfg(target_arch = "riscv32")'.dependencies]
+full_statement_verifier = { workspace = true, default-features = false, features = ["unified_verifier_only_security_100", "modular_ops", "blake2_with_compression"] }
```
AntonD3, line 29: So, right now it's 100 bit? A decision on this is a blocker, right?

AntonD3 (approved):

PR #627, opened 2026-04-21
fix: l2 tx intrinsic native cost fixes

AntonD3:
## What ❔

Fix few issues related to l2 txs intrinsic native cost and its estimation.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/api/src/helpers.rs
+++ b/api/src/helpers.rs
@@ -408,13 +408,11 @@ pub fn validate_l2_tx_intrinsic_native_resources(
         return Err(());
     }
 
-    // following bootloader: native is unlimited on 0 base fee chains
-    if base_fee == 0 {
-        return Ok(());
-    }
-
     // Compute effective gas price
-    let gas_price = {
+    let gas_price = if base_fee == 0 {
+        // Following bootloader: if base fee is zero, then we ignore priority fee
+        U256::ZERO
+    } else {
         let priority_fee = min(max_priority_fee_per_gas, max_fee_per_gas - base_fee);
         base_fee + priority_fee
     };
@@ -428,13 +426,18 @@ pub fn validate_l2_tx_intrinsic_native_resources(
     // native_per_pubdata = pubdata_price / native_price
     let native_per_pubdata = u256_try_to_u64(&pubdata_price.wrapping_div(native_price)).ok_or(())?;
 
-    let native_prepaid = native_per_gas.saturating_mul(gas_limit);
+    // following bootloader behavior
+    let native_limit = if native_per_gas == 0 {
+        u64::MAX - 1
+    } else {
+        native_per_gas.saturating_mul(gas_limit)
+    };
 
     // Intrinsic pubdata
     let intrinsic_pubdata = calculate_l2_tx_intrinsic_pubdata(authorization_list_num, false);
     let intrinsic_pubdata_overhead = native_per_pubdata.saturating_mul(intrinsic_pubdata);
 
-    let native_limit = native_prepaid
+    let native_limit = native_limit
         .checked_sub(intrinsic_pubdata_overhead)
         .ok_or(())?;
 
--- a/basic_bootloader/src/bootloader/constants.rs
+++ b/basic_bootloader/src/bootloader/constants.rs
@@ -107,7 +107,7 @@ pub const L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_CALLDATA_BYTE: u64 =
 pub const L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_ACCESS_LIST_PER_ADDRESS: u64 =
     PER_ADDRESS_ACCESS_LIST_NATIVE_COMPUTATIONAL_OVERHEAD + // computational overhead
     NEW_COLD_ACCOUNT_READ_COST + // worst case account read
-    31 * DYNAMIC_PART_KECCAK_COMPUTATIONAL_NATIVE_PER_BYTE * 2; // keccak for signing + full hash, 30 - worst case contribution to rlp encoding (5 length of payload,  21 address, 5 keys list length encoding)
+    31 * DYNAMIC_PART_KECCAK_COMPUTATIONAL_NATIVE_PER_BYTE * 2; // keccak for signing + full hash, 31 - worst case contribution to rlp encoding (5 length of payload,  21 address, 5 keys list length encoding)
 
 /// L2 tx access list storage slot computational native cost.
 pub const L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_ACCESS_LIST_PER_STORAGE_KEY: u64 =
@@ -123,7 +123,7 @@ pub const L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_AUTHORIZATION: u64 =
     NEW_COLD_ACCOUNT_READ_COST + // worst case account read
     ACCOUNT_UPDATE_COST + // nonce update
     ACCOUNT_UPDATE_COST + PREIMAGE_CACHE_SET_NATIVE_COST + keccak256_native_cost_for_rounds_u64(1) /*bytecode hashing */ + blake2s_native_cost(24) /* blake2s padded bytecode */ + // delegation write
-    132 * DYNAMIC_PART_KECCAK_COMPUTATIONAL_NATIVE_PER_BYTE * 2; // keccak for tx signing + full hash, 133 - worst case contribution to rlp encoding (33 chain_id, 21 address, 9 nonce, 1 y_parity, 33 r, 33 s, 2 list overhead)
+    132 * DYNAMIC_PART_KECCAK_COMPUTATIONAL_NATIVE_PER_BYTE * 2; // keccak for tx signing + full hash, 132 - worst case contribution to rlp encoding (33 chain_id, 21 address, 9 nonce, 1 y_parity, 33 r, 33 s, 2 list overhead)
 
 /// Native computational overhead of 7702 auth.
 pub const PER_AUTH_NATIVE_COMPUTATIONAL_OVERHEAD: u64 = 2000;
--- a/basic_bootloader/src/bootloader/constants.rs
+++ b/basic_bootloader/src/bootloader/constants.rs
@@ -88,7 +88,7 @@ pub const SERVICE_TX_INTRINSIC_COMPUTATIONAL_NATIVE_COST: u64 = NEW_COLD_ACCOUNT
 
 /// Service tx calldata byte intrinsic computational native cost.
 pub const SERVICE_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_CALLDATA_BYTE: u64 =
-    DYNAMIC_PART_KECCAK_COMPUTATIONAL_NATIVE_PER_BYTE; // to cover full hash
+    COPY_BYTE_NATIVE_COST + DYNAMIC_PART_KECCAK_COMPUTATIONAL_NATIVE_PER_BYTE; // to cover copying + full hash
 
 /// Native computational cost to cover keccak256 hashing overhead for dynamic fields of the transaction per byte.
 /// NOTE: this is approximate cost for hashing of 1 byte, but it shouldn't be used to estimate cost of one keccak call,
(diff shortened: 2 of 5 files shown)


antoniolocascio (approved):

0xVolosnikov:
@vv-dev-ai ptal

vv-dev-ai:
[AI-review] Reviewed commit `6752ef8a8cf2dd9a097d1c18134c80a0d9552791` — **COMMENT** (6 inline comments).

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -107,12 +107,12 @@ pub const L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_CALLDATA_BYTE: u64 =
 pub const L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_ACCESS_LIST_PER_ADDRESS: u64 =
     PER_ADDRESS_ACCESS_LIST_NATIVE_COMPUTATIONAL_OVERHEAD + // computational overhead
     NEW_COLD_ACCOUNT_READ_COST + // worst case account read
-    30 * DYNAMIC_PART_KECCAK_COMPUTATIONAL_NATIVE_PER_BYTE * 2; // keccak for signing + full hash, 30 - worst case contribution to rlp encoding (21 address, 9 keys list length encoding)
+    31 * DYNAMIC_PART_KECCAK_COMPUTATIONAL_NATIVE_PER_BYTE * 2; // keccak for signing + full hash, 30 - worst case contribution to rlp encoding (5 length of payload,  21 address, 5 keys list length encoding)
```
vv-dev-ai: [AI-review] **Nit:** Stale leading number in the trailing comment. The multiplier was changed to `31` and the parenthetical now enumerates `5 + 21 + 5 = 31`, but the comment still opens with `30 - worst case contribution`. Also `keys list length encoding` is misleading — per the description this is the address-entry length prefix, not the keys list.
Fix: `31 - worst case contribution to rlp encoding (5 length of payload, 21 address, 5 keys list length encoding)`.

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -107,12 +107,12 @@ pub const L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_CALLDATA_BYTE: u64 =
...
 
 /// L2 tx access list storage slot computational native cost.
 pub const L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_ACCESS_LIST_PER_STORAGE_KEY: u64 =
     PER_SLOT_ACCESS_LIST_NATIVE_COMPUTATIONAL_OVERHEAD + // computational overhead
-    COLD_NEW_STORAGE_READ_NATIVE_COST + // worst case storage slot read
+    WARM_STORAGE_READ_NATIVE_COST + COLD_NEW_STORAGE_READ_NATIVE_COST + // worst case storage slot read
     33 * DYNAMIC_PART_KECCAK_COMPUTATIONAL_NATIVE_PER_BYTE * 2; // keccak for signing + full hash, 33 contribution to rlp encoding length
```
vv-dev-ai, line 116: [AI-review] **Low:** The change `COLD_NEW_STORAGE_READ_NATIVE_COST → WARM_STORAGE_READ_NATIVE_COST + COLD_NEW_STORAGE_READ_NATIVE_COST` correctly adds the unconditional 4000-native warm charge that `apply_read_impl` in `basic_system/.../generic_pubdata_aware_plain_storage.rs:152` issues on every materialization — the old constant was underestimating and could break `verify_intrinsic_native`'s `actual_used <= formula` invariant under a long access-list input.
Suggestion: add a short comment referencing `apply_read_impl` (or `storage_touch`) so the next reader does not wonder why warm+cold are summed here; it is easy to read as a double-charge. E.g. `// touch_storage_slot charges warm unconditionally (apply_read_impl), plus cold-new extra on first access`.

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -123,7 +123,7 @@ pub const L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_AUTHORIZATION: u64 =
     NEW_COLD_ACCOUNT_READ_COST + // worst case account read
     ACCOUNT_UPDATE_COST + // nonce update
     ACCOUNT_UPDATE_COST + PREIMAGE_CACHE_SET_NATIVE_COST + keccak256_native_cost_for_rounds_u64(1) /*bytecode hashing */ + blake2s_native_cost(24) /* blake2s padded bytecode */ + // delegation write
-    133 * DYNAMIC_PART_KECCAK_COMPUTATIONAL_NATIVE_PER_BYTE * 2; // keccak for tx signing + full hash, 133 - worst case contribution to rlp encoding (33 chain_id, 21 address, 9 nonce, 1 y_parity, 33 r, 33 s, 3 list overhead)
+    132 * DYNAMIC_PART_KECCAK_COMPUTATIONAL_NATIVE_PER_BYTE * 2; // keccak for tx signing + full hash, 133 - worst case contribution to rlp encoding (33 chain_id, 21 address, 9 nonce, 1 y_parity, 33 r, 33 s, 2 list overhead)
```
vv-dev-ai: [AI-review] **Nit:** Stale leading number in the trailing comment. The multiplier was changed to `132` and the parenthetical sums to `33 + 21 + 9 + 1 + 33 + 33 + 2 = 132`, but the comment still opens with `133 - worst case contribution`.
Fix: change to `132 - worst case contribution to rlp encoding (33 chain_id, 21 address, 9 nonce, 1 y_parity, 33 r, 33 s, 2 list overhead)`.

api/src/helpers.rs
```diff
@@ -428,13 +426,18 @@ pub fn validate_l2_tx_intrinsic_native_resources(
     // native_per_pubdata = pubdata_price / native_price
     let native_per_pubdata = u256_try_to_u64(&pubdata_price.wrapping_div(native_price)).ok_or(())?;
 
-    let native_prepaid = native_per_gas.saturating_mul(gas_limit);
+    // following bootloader behavior
+    let native_limit = if native_per_gas == 0 {
+        u64::MAX - 1
+    } else {
+        native_per_gas.saturating_mul(gas_limit)
+    };
```
vv-dev-ai, line 434: [AI-review] **Low:** This helper now mirrors the production bootloader `create_resources_for_tx` (free-native when `native_per_gas == 0`, including the `u64::MAX - 1` sentinel), but it does not mirror the simulation branch in `validation_impl.rs:125-135` (`Config::SIMULATION && gas_price.is_zero()` → derive `native_per_gas` from `base_fee.div_ceil(native_price)`). Any caller using this helper to estimate simulation/eth_call behavior will now diverge from the bootloader on free-fee chains.
Suggestion: either document that this helper reflects non-simulation behavior only, or add a `simulation: bool` parameter that selects the `base_fee`-derived path.

vv-dev-ai (commented):
[AI-review] PR #627 — **COMMENT**

The changes look like correctness fixes to the L2 intrinsic native/pubdata accounting, and I verified the intended bootloader behavior in each case:

- `validate_l2_tx_intrinsic_native_resources` now mirrors `create_resources_for_tx` for the `base_fee == 0` / `native_per_gas == 0` (free-native) case.
- `L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_ACCESS_LIST_PER_STORAGE_KEY` now includes `WARM_STORAGE_READ_NATIVE_COST`. This matches `apply_read_impl` in `generic_pubdata_aware_plain_storage.rs`, which charges warm (4000) unconditionally plus cold-extra (up to 200000) when materializing a slot via `storage_touch` — the old intrinsic formula was under-estimating by 4000 native per access-list slot, which could fail the `verify_intrinsic_native` upper-bound assertion.
- `SERVICE_TX_INTRINSIC_COMPUTATIONAL_NATIVE_PER_CALLDATA_BYTE` now includes `COPY_BYTE_NATIVE_COST`, matching the service-tx per-calldata-byte work.
- The pubdata-reporting refactor in `zk/mod.rs` removes a double-count: the old reported `pubdata_used = net_pubdata + intrinsic_pubdata` summed validation-phase pubdata twice (once inside `net_pubdata`, once inside `intrinsic_pubdata` which already budgets the sender and coinbase balance diffs). The new reported value `(net_pubdata - validation_pubdata) + intrinsic_pubdata` is consistent with the native charges.

Non-blocking observations below.

### Behavior changes to confirm with downstream consumers

- `TxOutput.pubdata_used` decreases by `validation_pubdata` per L2 tx (≈ sender nonce/balance diff bytes, typically tens of bytes). This feeds `block_data.block_pubdata_used` and the `check_for_block_limits` pubdata check (`block_flow/zk/mod.rs:77`). The new value is only safe under the documented assumption that `intrinsic_pubdata ≥ validation_pubdata` for every path (including revert, where validation-phase writes persist). If the server or prover uses `pubdata_used` for billing, DA accounting, or gas estimation, please confirm they agree with the new semantics.
- `validate_l2_tx_intrinsic_native_resources` no longer short-circuits on `base_fee == 0`. Any external caller (sequencer pre-filter, RPC helper) will now reject transactions on free-fee chains that pass `checked_sub(intrinsic_pubdata_overhead)` or `checked_sub(intrinsic_computational_native)` below zero; previously all such txs were accepted. This matches the production bootloader path, but note it does not mirror the `Config::SIMULATION && gas_price == 0` branch in `validation_impl.rs:125-135` that derives `native_per_gas` from `base_fee` — callers using this helper for simulation/estimation may want a parallel helper or a `simulation` flag.

### Suggested Tests

- `validate_l2_tx_intrinsic_native_resources`: a case with `base_fee = 0, native_price > 0, pubdata_price > 0, large intrinsic_pubdata` to verify the pubdata-overhead check now fires (old early-return path would have returned Ok). Assert `Err(())`.
- `validate_l2_tx_intrinsic_native_resources`: a case with `base_fee = 0` and tiny `intrinsic_computational_native` / zero `pubdata_price` to confirm the free-native path still accepts otherwise-valid txs. Assert `Ok(())`.
- L2 tx pubdata reporting: a tx that succeeds with non-zero `validation_pubdata` and non-zero execution pubdata — assert `TxOutput.pubdata_used == (net_pubdata - validation_pubdata) + intrinsic_pubdata` so a future regression that re-introduces the double-count is caught.
- L2 tx reverting during execution — assert `TxOutput.pubdata_used == intrinsic_pubdata` (execution pubdata saturates to 0 after rollback).
- `verify_intrinsic_native` coverage: a single-entry access-list tx (cold new storage key) to exercise the `WARM + COLD_NEW` cost path and confirm `formula >= actual_used` still holds.

basic_bootloader/src/bootloader/transaction_flow/zk/mod.rs
```diff
@@ -398,24 +398,27 @@ where
...
+        );
 
         // Pubdata for validation has been charged already,
         // we charge for the rest now.
         let (total_pubdata_used, to_charge_for_pubdata) = match pubdata_info {
             Some(CachedPubdataInfo {
                 pubdata_used,
                 to_charge_for_pubdata,
-            }) => (pubdata_used + validation_pubdata, to_charge_for_pubdata),
+            }) => (pubdata_used + intrinsic_pubdata, to_charge_for_pubdata),
```
vv-dev-ai, line 412: [AI-review] **Low:** With this refactor, on the revert path the reported `pubdata_used` becomes exactly `intrinsic_pubdata` (since `execution_pubdata_spent` saturates to 0 after the execution frame is rolled back), dropping `validation_pubdata` from the report. That is correct only if `intrinsic_pubdata` is always an upper bound on the permanent validation-phase pubdata — i.e. sender nonce+balance diff ≤ `SENDER_BALANCE_INTRINSIC_PUBDATA`, plus authorization-list diffs ≤ `authorization_list_num * L2_TX_INTRINSIC_PUBDATA_PER_AUTHORIZATION`. The relation holds under the current constants, but it is now load-bearing in three places (`before_refund`, `after_execution`, block-limit accounting). Worth a short invariant comment here so a future tweak to `SENDER_BALANCE_INTRINSIC_PUBDATA` or `L2_TX_INTRINSIC_PUBDATA_PER_AUTHORIZATION` is flagged as affecting pubdata reporting.

basic_bootloader/src/bootloader/transaction_flow/zk/validation_impl.rs
```diff
@@ -214,6 +216,20 @@ where
...
+        <<S as SystemTypes>::Resources as Resources>::Native::from_computational(
+            calldata.len() as u64 * COPY_BYTE_NATIVE_COST,
+        ),
+    ))?;
+    // 2. Hashing tx hash into the rolling hash after the execution
+    intrinsic_resources.charge(&Resources::from_native(
+        <<S as SystemTypes>::Resources as Resources>::Native::from_computational(
+            keccak256_native_cost_for_rounds_u64(1),
+        ),
+    ))?;
```
vv-dev-ai, line 231: [AI-review] **Nit:** These two charges are routed to `intrinsic_resources` (the FORMAL_INFINITE tracker), not to `context.resources.main_resources`, so they only matter under `verify_intrinsic_native`. The comment `"we still need to charge the user for them"` reads as if these affect actual user resources — but the user is charged via the upfront `intrinsic_computational_native` deduction inside `create_resources_for_tx`, and these lines only bring the verification tracker in sync with that formula. Consider rewording, e.g. `"Record these against the verification tracker so the formula upper-bound check covers them — actual user charge is already done via intrinsic_computational_native."`

basic_bootloader/src/bootloader/constants.rs
AntonD3: fixed

basic_bootloader/src/bootloader/constants.rs
AntonD3: fixed

antoniolocascio (approved):

PR #619, opened 2026-04-13
feat: integrate airbender-platform build, runtime, and host APIs

vibelyova:
## What ❔

Integrate airbender-platform into zksync-os (Phase 1), replacing direct dependencies on `riscv_transpiler`, `execution_utils`, and `prover_examples` with the platform's standardized build, execution, and proving pipeline.

### Build
- `cargo airbender build` replaces `cargo build --target riscv32i` + `cargo objcopy` in `dump_bin.sh`
- Target changed from `riscv32i-unknown-none-elf` to `riscv32im-risc0-zkvm-elf`
- Build artifacts use `dist/<app>/` layout with `manifest.toml` (from prep PR #614)

### Guest runtime (`zksync_os/`)
- `#[airbender::main(allocator_init = init_allocator)]` replaces manual boot assembly, `.data`/`.rodata` section loading, custom `memcpy.s`/`memset.s`, allocator init, and panic handler
- CSR read/write uses `airbender::rt::sys::{read_word, write_word}` instead of `riscv_common`
- `heapless` dependency removed; `quasi_uart` simplified accordingly

### Host runner (`zksync_os_runner/`)
- `TranspilerRunner` (via `Program::load` + `runner.run(&[u32])`) replaces direct `VM` usage
- Runner now accepts pre-recorded `&[u32]` input instead of interactive oracle sources
- `CpuProver` replaces manual `execution_utils` wiring for e2e proving
- `FlamegraphOptions` preserved as public API; flamegraph support now unconditionally available via `airbender-host`

### Test rig (`tests/rig/`)
- RISC-V run now consumes `prover_input_forward` (pre-recorded words from native run) directly
- Interactive oracle wrapping (`ReadWitnessSource`) removed from the RISC-V run path
- `simulate_witness_gen` feature removed - it was already obsolete, now became incompatible with new API
- eth_stf path: uses `EthereumStorageSystemTypesWithPostOps` with `USE_ADVICE=true` and native callable oracles for prover-input recording

### Dependencies
- Added: `airbender-guest`, `airbender-host`, `airbender-sdk` (from `matter-labs/airbender-platform` rev `7e559f95`)
- Removed: `common_constants`, `prover_examples`, `execution_utils`, `heapless`, `riscv_common`
- Kept: `riscv_transpiler` at rev `73f46c2b` (transitive via `airbender-host`)

### CI
- All workflows install `cargo-airbender` and use it for RISC-V builds
- Blake test binaries built in CI instead of committed to git
[...]

PR #628, opened 2026-04-21
chore: airbender-platform migration follow-ups

vibelyova:
## What ❔

Follow-up cleanup after #619 (airbender-platform migration). Builds on #630 (composite toolchain action + `bench-base` migration), which was split out and merged separately into `draft-0.4.0`.

### CI
- **Drop the 10× crypto test loop** — run once, without `--quiet` so failures are readable.
- **Remove `-j 3` throttling** across `ci.yml` and `bench.yml`. The flag caps `cargo`'s *build* parallelism to 3 and has no effect on test determinism (that would be `--test-threads`). The CI runners have plenty of cores.
- **Speed up fuzzer regression**:
  - `cargo fuzz build -s none` drops AddressSanitizer for regression replay; this also skips the implicit `-Zbuild-std` (stdlib rebuild).
  - Drop the redundant `dump_bin.sh --type for-tests` step — `fuzz.sh regression` already does it via `prepare()`.
  - Replace `fuzz.sh install` (reinstalls cargo-binutils, cargo-airbender, rustfilt, lcov-util, and npm @lcov-viewer/cli) with a targeted `cargo install cargo-fuzz --locked`.
- **Simplify `bench-base`** — now that the composite `setup-toolchain` action exists in the merge-base (#630), use it instead of inlining the rust + cargo-airbender setup.

### `reproduce.sh` rewrite
Replace the Dockerfile-based reproduce pipeline with `cargo airbender build --reproducible` (which spawns its own pinned container per build and shares a cargo registry volume across builds).
- Drop `zksync_os/reproduce/Dockerfile`.
- `reproduce.sh` regenerates both `Cargo.lock` files (workspace + guest), then runs `dump_bin.sh --type <t> --reproducible` sequentially for every app type.
- Sequential, not parallel: concurrent `--reproducible` builds race on the shared `airbender-cargo-registry` Docker volume (`.cargo-ok` / `.cache` mkdir conflicts). Proper fix belongs in `cargo-airbender` (multi-binary-per-container support).
- Copy `dist/<app>/app.bin → zksync_os/<app>.bin` at the end for backwards compatibility with downstream consumers (release workflow, zksync-era).

### `zksync_os_runner` API cleanup
Collapse `run` / `run_and_get_effective_cycles` / `run_with_flamegraph` into a single builder:
```rust
zksync_os_runner::Runner::new(dist_dir)
    .with_cycles(1 << 25)                    // optional; default 1 << 36
    .with_flamegraph(opts)                   // optional
    .run(&input_words);                      // -> RunResult { output, block_effective }
```
The default cycle limit is `1 << 36`, which is what every non-test caller was passing. `block_effective` is only populated when the `cycle_marker` feature is enabled.

### Script / binary cleanup in `zksync_os/`
- Delete stale helper scripts (`dump_bin_manual.sh`, `objdump.sh`, `objdump_manual.sh`, `sections_layout.sh`, `show_asm.sh`, `build_release_std.sh`, `dump_bin_no_delegation.sh`) — all referenced the obsolete `riscv32i-unknown-none-elf` target, the `_manual` variants hardcoded macOS toolchain paths, and they don't work with the new `cargo airbender build` flow.
- Fix `upd_local_era.sh` — source paths (`singleblock_batch.bin`) were broken since the move to `dist/<app>/app.bin` layout.

### Misc
- Replace the local `QuasiUART` with airbender-platform's implementation.
- Bump `airbender-platform` to `097f1489`.

## Why ❔

Cleaner post-migration state, and CI savings on jobs where every minute counts (fuzzer, bench).

## Is this a breaking change?
- [ ] Yes
- [x] No

(`zksync_os_runner` API change affects only in-repo callers, all updated here.)

## Checklist


zksync_os/reproduce/reproduce.sh
```diff
@@ -1,35 +1,35 @@
...
 
-# create a fresh docker
-docker build \
-  --build-arg SOURCE_DATE_EPOCH="$SDE" \
-  --platform linux/amd64 \
-  -t zksync-os-bin \
-  -f zksync_os/reproduce/Dockerfile .
+# `cargo airbender build --reproducible` passes `--locked` to cargo, so
+# both the workspace root and the guest crate need a Cargo.lock regenerated
+cargo +nightly-2026-02-10 generate-lockfile --manifest-path "$REPO_ROOT/Cargo.toml"
```
0xVolosnikov, line 15: We should not forget to change it when we update the toolchain version

ly0va: yes, hence the [comment](https://github.com/matter-labs/zksync-os/pull/628/changes#diff-5387e036acddfa43ec271b6c47e20b87eaaaa0779e6808d8ea4722ab93f9f712L1)

antoniolocascio (approved):

PR #631, opened 2026-04-23
refactor(bootloader): seed block-intrinsic pubdata/native at pre_tx_loop

antoniolocascio:
## What ❔
Block-level overheads (pubdata encoding envelope, future pre-tx-loop system work) must be visible to per-tx block-limit enforcement, not only to the final reported value. Seed block_data.block_pubdata_used and block_data.block_computational_native_used with BLOCK_INTRINSIC_* constants at pre-tx-loop start so check_for_block_limits enforces the true budget.

Plumb both counters out through the result keeper via new record_block_native_used / record_block_pubdata_used hooks, so the forward path's BlockOutput.{computational_native_used, pubdata_used} are the bootloader's post-accumulation values rather than an independent per-tx sum (native) or a zero placeholder (pubdata).

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/post_tx_op_proving_multiblock_batch.rs
+++ b/basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/post_tx_op_proving_multiblock_batch.rs
@@ -62,6 +62,8 @@ where
         )?;
         let block_hash = Bytes32::from(block_header.hash());
         result_keeper.block_sealed(block_header);
+        result_keeper.record_block_native_used(block_data.block_computational_native_used);
+        result_keeper.record_block_pubdata_used(block_data.block_pubdata_used);
 
         let mut logger = system.get_logger();
         logger_log!(logger, "Basic header information was created\n");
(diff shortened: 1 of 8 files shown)


forward_system/src/run/output.rs
```diff
@@ -88,7 +88,6 @@ impl<TR: TxResultCallback>
                         } else {
                             ExecutionResult::Revert(output.output)
                         };
-                        block_computational_native_used += output.computational_native_used;
```
AntonD3, line 91: Looks like now you are not counting txs native computational and pubdata in the results keeper, I assume you expect that record_block_native_used/record_block_pubdata_used will be called with total block pubdata(including txs), but now in fact block intrinsic passed there, perhaps you need to change tx loop

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/post_tx_op_proving_multiblock_batch.rs
```diff
@@ -62,6 +62,8 @@ where
         )?;
         let block_hash = Bytes32::from(block_header.hash());
         result_keeper.block_sealed(block_header);
+        result_keeper.record_block_native_used(block_data.block_computational_native_used);
```
AntonD3, line 65: nit: computational_native? I remember in some places I was using word "native" for both pubdata and computational

AntonD3 (approved):

AntonD3 (approved):

PR #598, opened 2026-03-25
feat: Additional optimizations

0xVolosnikov:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.
- [ ] Code has been formatted.

PR #638, opened 2026-04-29
feat: custom u256 and other optimizations

ly0va:
## What ❔

Migrates the EVM interpreter from `ruint::aliases::U256` to the custom non-`Copy` `u256::U256` and adds oracle-backed advice for MULMOD and DIV opcodes.

### Custom U256 migration (from PR #588, #598)
- Custom U256 utilities in `zk_ee` (`custom_u256_utils`)
- EVM interpreter stack, opcode, and frame state migration to non-Copy U256
- Forward system tracer boundary adaptation (lazy conversion to alloy U256)
- Delegation feature wiring through the proving path
- `push_u64`/`push_b160` for direct small-value stack writes
- Specialized PUSH1 to skip bytereverse+shift overhead
- Environment opcodes (chainid, timestamp, etc.) use `push_u64`/`push_b160`

### Oracle-backed MULMOD and DIV advice
- New `U256_MULMOD_ADVICE_QUERY_ID` and `U256_DIV_REM_ADVICE_QUERY_ID` oracle queries
- IOOracle-based advice functions in `zk_ee::utils::u256_arithmetic_advice` that work on both RISC-V and native (no raw CSR — uses `IOOracle::raw_query`)
- Value-based query format (limbs serialized directly via `UsizeSerializable`) — same handler works for both `ArithmeticQuery` and `NativeArithmeticQuery`
- `NativeArithmeticQuery` extended to serve both query IDs, keeping witness streams synchronized between RISC-V simulation and native prover-input runs

### Performance optimizations
- `widening_mul_assign_into` with both halves from limbs instead of `widening_mul_assign`
- Direct sub+borrow for `r < d` checks instead of `Ord::cmp` (saves 1 EQ delegation)
- `write_be_bytes_into` for LOG topics and storage opcode boundary conversions (avoids clone+bytereverse delegation)
- Limb-by-limb `i256_cmp` avoids clone+sub delegation in SLT/SGT same-sign comparison
- `UsizeDeserializable` impl for `[T; N]` arrays

## Why ❔

The proving-oriented custom U256 reduces RISC-V cycle count by routing arithmetic through the prover's bigint delegation circuit instead of software implementations. Oracle advice replaces expensive software division (especially 512÷256 for MULMOD) with a hint+verify pattern that costs fewer delegation calls.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


forward_system/src/run/tracing_impl.rs
```diff
@@ -26,23 +26,30 @@ struct ExecutionEnvironmentLaunchParamsWrapped<'a, 'b, S: EthereumLikeTypes>(
...
-    inner: &'a dyn zk_ee::system::evm::EvmStackInterface,
+    cached_values: OnceCell<Vec<U256>>,
+    inner: &'a dyn EvmStackInterface,
 }
 
 impl<'a, S: EthereumLikeTypes + 'a, T: EvmFrameInterface<S>> From<&'a T>
     for EvmFrameInterfaceWrapped<'a, S, T>
 {
     fn from(value: &'a T) -> Self {
+        let call_value_converted: ruint::aliases::U256 = value.call_value().clone().into();
```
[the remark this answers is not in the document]
ly0va: they're the same type

callable_oracles/src/arithmetic/mod.rs
```diff
@@ -3,13 +3,29 @@ use basic_system::system_functions::modexp::{
...
+    U256MulmodAdviceParams64,
+};
 
 use crate::utils::{
     evaluate::{read_memory_as_u64, read_struct},
     usize_slice_iterator::UsizeSliceIteratorOwned,
 };
 use crate::{read_host_struct, read_u64_words};
 
+const _: () = assert!(U256_DIV_REM_ADVICE_QUERY_ID == 0x4005_0030);
```
0xVolosnikov: Is this just a sanity check?

zk_ee/src/oracle/usize_serialization/mod.rs
```diff
@@ -388,3 +387,16 @@ impl<T: UsizeSerializable, const N: usize> UsizeSerializable for [T; N] {
         )
     }
 }
+
+impl<T: UsizeDeserializable + Copy, const N: usize> UsizeDeserializable for [T; N] {
+    const USIZE_LEN: usize = <T as UsizeDeserializable>::USIZE_LEN * N;
+
+    fn from_iter(src: &mut impl ExactSizeIterator<Item = usize>) -> Result<Self, InternalError> {
+        let mut result: [MaybeUninit<T>; N] = unsafe { MaybeUninit::uninit().assume_init() };
```
0xVolosnikov: Let's at least add a comment that this is safe in RISC-V environment. Otherwise `MaybeUninit::uninit().assume_init()` will be reported as UB

ly0va: we can actually replace it with 

```rust
let mut result: [MaybeUninit<T>; N] = [const { MaybeUninit::uninit() }; N];
```

which does not need unsafe and does the same thing

zk_ee/src/oracle/query_ids.rs
```diff
@@ -39,6 +39,20 @@ pub const ACCOUNT_AND_STORAGE_SUBSPACE_MASK: u32 = BASIC_SUBSPACE_MASK | 0x00_03
 pub const STATE_AND_MERKLE_PATHS_SUBSPACE_MASK: u32 = BASIC_SUBSPACE_MASK | 0x00_04_00_00; // 0x40040000
 /// Computational advice queries (e.g. division/modexp advice)
 pub const ADVICE_SUBSPACE_MASK: u32 = BASIC_SUBSPACE_MASK | 0x00_05_00_00; // 0x40050000
+
+// ========== Advice Subspace Queries ==========
+
+/// Query to get division hint (quotient, remainder) for U256 values.
+/// Guest sends pointers to dividend and divisor; host returns q and r.
+/// Guest verifies q * d + r == n and r < d using delegated arithmetic.
+pub const U256_DIV_REM_ADVICE_QUERY_ID: u32 = ADVICE_SUBSPACE_MASK | 0x30; // 0x40050030
```
AntonD3, line 48: Why 0x30? Not saying it's wrong, just curious how you came up with this number

zk_ee/src/oracle/query_ids.rs
```diff
@@ -39,6 +39,20 @@ pub const ACCOUNT_AND_STORAGE_SUBSPACE_MASK: u32 = BASIC_SUBSPACE_MASK | 0x00_03
...
+
+/// Query to get mulmod hint (quotient, remainder) for U256 values.
+/// Guest sends pointers to a, b, and modulus; host returns q and r
+/// such that a*b == q*m + r and r < m.
+/// Guest verifies the relation using delegated widening arithmetic.
```
AntonD3: Maybe a dumb question - but why not reuse division advice with `a*b`?

AntonD3: The only issue that I see is that a * b is 512 bit, so maybe it's good to have 2 advices, one for 256, and one for 512. But anyway this one seems a bit ad-hoc, while we can have more generic query

AntonD3: Btw maybe I missed it, but did you consider having advice for addmod? Maybe we can reuse query, if it will be more generic

zk_ee/src/utils/integer_utils.rs
```diff
@@ -157,6 +157,75 @@ pub fn u256_try_to_b160(src: U256) -> Option<B160> {
     Some(result)
 }
 
+// Functions that work with u256::U256 (custom non-Copy U256)
+pub mod custom_u256_utils {
+    use u256::U256 as CustomU256;
+
+    #[inline(always)]
+    pub fn custom_u256_to_b160(src: &CustomU256) -> ruint::aliases::B160 {
```
AntonD3: nit: maybe mention that it's BE(if I understood correctly)

evm_interpreter/src/u256_helpers.rs
```diff
@@ -0,0 +1,7 @@
+use u256::U256;
+
+pub(crate) fn log2floor(value: &U256) -> u64 {
+    assert!(!value.is_zero());
```
AntonD3, line 6: nit: maybe mention in comment that panics if value is zero

zk_ee/src/oracle/query_ids.rs
ly0va: I think this is a good point, will try to implement and compare the benchmarks. Note on the addmod though - it won't catch the difference because our benchmarks don't have diverse enough operands to trigger the case where advice is needed.

zk_ee/src/oracle/query_ids.rs
ly0va: [It wasn't me :)](https://github.com/matter-labs/zksync-os/pull/588/changes#diff-7c57153e5a4fd02fa472a5c2a5c8f31bf71edc41e56f63a6480eacb4feb73d30R48)

AntonD3: Maybe @0xVolosnikov has an answer

zk_ee/src/utils/integer_utils.rs
ly0va: I'm pretty sure it's LE, same as everywhere else

zk_ee/src/oracle/query_ids.rs
0xVolosnikov: No particular reason, maybe I had more queries initially and forgot to update

zk_ee/src/utils/integer_utils.rs
AntonD3: Internal representation of U256 and B160 uses LE, but in a fact most significant byte of U256 value goes into first byte of B160, if I understand correctly, so this B160 will hold BE representation of a given number

zk_ee/src/utils/integer_utils.rs
```diff
@@ -157,6 +157,75 @@ pub fn u256_try_to_b160(src: U256) -> Option<B160> {
     Some(result)
 }
 
+// Functions that work with u256::U256 (custom non-Copy U256)
```
AntonD3: Do you think it makes sense to move them to U256 implementation?

basic_system/src/system_functions/u256_advice.rs
```diff
@@ -0,0 +1,231 @@
...
+        divisor_or_remainder,
+        q_limbs
+    ));
+
+    // dividend_or_quotient is now the remainder; swap then write quotient
+    core::mem::swap(dividend_or_quotient, divisor_or_remainder);
+    *dividend_or_quotient = U256::from_limbs(q_limbs);
+}
+
+fn u256_wide_div_rem_software(dividend_lo: &mut U256, dividend_hi: &mut U256, divisor: &mut U256) {
```
AntonD3: Not sure `_software` is a great name for this function

basic_system/src/system_functions/u256_advice.rs
```diff
@@ -0,0 +1,231 @@
...
+    assert!(!divisor.is_zero());
+    let mut product = [0u64; 8];
```
AntonD3: It's not product

evm_interpreter/src/instructions/bitwise.rs
```diff
@@ -1,76 +1,105 @@
 use super::*;
 use native_resource_constants::*;
 
+#[inline(always)]
+fn arithmetic_shr_in_place(value: &mut U256, shift: usize) {
```
AntonD3: Can't we reuse ruint implementation?

zk_ee/src/utils/u256_arithmetic_advice.rs
```diff
@@ -0,0 +1,22 @@
+/// Params for U256 div_rem oracle query (pointer-based, like modexp).
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+pub struct U256DivRemAdviceParamsGeneric<W> {
```
AntonD3: All other advices defined in the system implementation, do you think we should keep it here? IMHO in `zk_ee` we should keep things needed to define interface and helpers needed for "users" of the system, but not things related to internal implementation

supporting_crates/u256/src/risc_v/mod.rs
```diff
@@ -166,12 +193,13 @@ impl U256 {
     }
 
     #[inline(always)]
-    /// Panics if divisor is 0
+    /// Panics if divisor is 0.
+    /// Note: EVM opcodes use the IOOracle-based advice path in zk_ee instead.
+    /// This software fallback is used by div_ceil, add_mod, and tests.
     pub fn div_rem(dividend_or_quotient: &mut Self, divisor_or_remainder: &mut Self) {
```
AntonD3, line 199: Do you think it makes sense to try to integrate advice into custom U256 here directly? So advice flow will be used by default, without separate system function. Or it complicates the design significantly? (I understand that we'll need to pass the oracle)

AntonD3: cc @0xVolosnikov

ly0va: unfortunately not possible because we need `IOOracle` for that, which is defined in a crate that depends on u256

ly0va: We can do a complicated version of accepting a trait object or a generic instead of a concrete oracle, and make IOOracle implement this trait, but idk if the complexity is worth it

evm_interpreter/src/instructions/bitwise.rs
ly0va: @0xVolosnikov

basic_system/src/system_functions/u256_advice.rs
ly0va: `_naive`?

evm_interpreter/src/instructions/bitwise.rs
ly0va: I think ruint does not have an arithmetic shr, only logical.

supporting_crates/u256/src/risc_v/mod.rs
AntonD3: Complicated version not worth it IMHO, but I think main reason why zk_ee depends on u256 is that we actually have system functions on u256, if we'll move this logic to u256 - no dependency

ly0va: Sure, but do we really want to pull in IOOracle into a primitive crate like u256?

evm_interpreter/src/instructions/bitwise.rs
AntonD3: I think it was used before this PR:
```
let shift = usize::try_from(op1).unwrap_or(256);
        *op2 = op2.arithmetic_shr(shift);
```

AntonD3: https://docs.rs/ruint/latest/ruint/struct.Uint.html#method.arithmetic_shr

0xVolosnikov: The current version may be a bit more effective - makes sense to benchmark the difference

0xVolosnikov: Generally I'd say it's preferable to reuse ruint, if the difference is not very significant

ly0va: SAR cycles reduced by 26% so I'd say it's worth it

0xVolosnikov: The only downside is that there is a bigger probability of issues in the custom implementation

supporting_crates/u256/src/risc_v/mod.rs
0xVolosnikov: I'd prefer not to pull IOOracle there

evm_interpreter/src/instructions/bitwise.rs
0xVolosnikov: Maybe we should add more comments to explain the design choice here and what exactly is optimized

AntonD3: > SAR cycles reduced by 26% so I'd say it's worth it

Can it be affected by general savings on stack operations?

AntonD3 (approved):

PR #645, opened 2026-05-12
feat(bench): benchmark pipeline (CI, scripts, tracers, docs)

0xVolosnikov:
## What ❔

End-to-end benchmark pipeline that consumes the STF cycle markers added in #656.

- `cycle_marker`: per-execution `<label>.cycles` (raw) and `<label>.effective.cycles` (raw + delegation weights) dumps; weighting uses the existing `process_block` formula (`raw + 16×Blake + 4×BigInt + 4×Keccak`).
- `forward_system`: `Pair` tracer combinator + `PrecompileStatsTracer` (per-precompile gas/native, dump samples parallel to `EvmOpcodeStatsTracer`).
- `bench_scripts/`: shared `benchlib.py`; `compare_*` / `join_*` scripts; `cycles_per_native_report.py` for local cycles/native ratio reports.
- Test scaffolding: `tests/rig/src/chain.rs` `SeqLabelsGuard` (needed under `BlobsZKsyncOS` DA where `blob_versioned_hash` fires only in proving mode); `eth_runner` tracer composition + `BENCH_DA_SCHEME`; precompile sweep tests.
- `.github/workflows/bench.yml`: full bench job — both DA schemes, `for-tests-benchmarking-pectra` proving binary, headline + collapsible PR comment, merge-base fallbacks for graceful degradation against older base commits.
- `bench-fast` profile (root + `eth_runner` mirror) to speed up host-side bench builds.
- `docs/benchmarking.md`: rewrite as agent-targeted reference.

## Why ❔

Split from the original #645 so reviewers can scrutinize the STF instrumentation in #656 (the small / security-relevant half) separately from this pipeline scaffolding (CI/scripts/tracers/docs/test rigging — review-skim). Once #656 lands, this PR's base auto-retargets to `draft-0.4.0`.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -46,7 +46,7 @@ jobs:
   bench-base:
     name: Bench base
     needs: merge-base
-    runs-on: [matterlabs-ci-runner-highmem]
+    runs-on: [matterlabs-ci-runner-high-performance]
     permissions:
       contents: read
     steps:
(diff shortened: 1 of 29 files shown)


bench_scripts/compare_bench.py
```diff
@@ -1,10 +1,28 @@
+import os
 import sys
 import re
 import ast
 
-U256BIGINTOPS_RATIO = 4
-BLAKE2ROUNDEXTENDED_RATIO = 16
-KECCAK_RATIO = 4  # TODO(EVM-1242): calibrate with actual proving benchmarks
```
ly0va, line 7: the TODO got lost

bench_scripts/compare_opcode_stats.py
```diff
@@ -11,18 +11,11 @@
 
 import os
 import re
+import os
```
ly0va: duplicate `import os`

cycle_marker/src/lib.rs
```diff
@@ -478,19 +532,75 @@ pub fn print_cycle_markers(cm: CycleMarker) -> CycleMarkerResults {
...
+    // it. Effective values let cycles/gas analysis reflect prover cost
+    // including delegation work (which raw RISC-V cycles do not capture).
+    // Caller contract: clean the dir before the first invocation so that stale
+    // files from a previous run are not mixed in. This block APPENDS so that
+    // multi-block test runs accumulate all samples rather than keeping only the
+    // last block's data.
+    if let Ok(dir) = std::env::var("LABEL_CYCLE_SAMPLES_DIR") {
+        let dir = std::path::Path::new(&dir);
+        std::fs::create_dir_all(dir).expect("Failed to create label cycle samples dir");
+        let append = |path: std::path::PathBuf, samples: &[u64]| {
```
ly0va: almost identical closure defined a few lines above, maybe worth reusing

forward_system/src/system/tracers/precompile_stats.rs
```diff
@@ -0,0 +1,542 @@
...
+impl<S: EthereumLikeTypes> Tracer<S> for PrecompileStatsTracer<S> {
+    fn on_new_execution_frame(&mut self, request: &ExecutionEnvironmentLaunchParams<S>) {
+        // Single-slot pending state: precompile frames are leaves in the
+        // current EVM dispatch (a precompile body cannot itself enter
+        // another EVM frame through this hook). If that invariant ever
+        // changes — e.g. a precompile bounces back through the EVM call
+        // frame mechanism — the assertion below catches the silent
+        // overwrite that would otherwise misattribute stats. Promote to
+        // a stack of `Option<PendingFrame>` if that becomes a real path.
+        debug_assert!(
```
ly0va: should this be promoted to `assert!`?

bench_scripts/compare_precompile_stats.py
```diff
@@ -0,0 +1,243 @@
...
+    names = sorted(set(base) | set(head))
+    rows = []
+    for name in names:
+        b = base.get(name, {})
+        h = head.get(name, {})
+        b_max_gas = b.get("max_gas", 0)
+        h_max_gas = h.get("max_gas", 0)
+        b_max_native = b.get("max_native", 0)
+        h_max_native = h.get("max_native", 0)
+        if b_max_gas == h_max_gas and b_max_native == h_max_native:
```
ly0va: what if median changed but not max?

bench_scripts/compare_bench.py
0xVolosnikov: I had an impression that it is basically done, isn't it?

.github/workflows/bench.yml
```diff
@@ -158,33 +256,103 @@ jobs:
...
+          cp -r head-results/precompile_samples/* precompile_samples/ 2>/dev/null || true
+          cp -r head-results/precompile_cycles/* precompile_cycles/ 2>/dev/null || true
 
       - name: Generate comparison
         shell: bash
         id: comparison
         run: |
           mkdir -p bench_results
-          pairs=""
+          # Build three separate pair lists so the resulting PR comment has a
```
ly0va: This step is currently extremely dense. Might be out of scope of this PR, but maybe worth extracting it into a separate script (either bash or maybe python). Same goes for "run benchmarks" step

bench_scripts/bench.sh
```diff
@@ -67,16 +75,69 @@ run_block() {
 run_precompiles() {
     local output_dir="$1"
 
+    local samples_dir="$output_dir/precompile_samples"
+    local cycles_dir="$output_dir/precompile_cycles"
+
+    # Ensure a clean slate for this run's samples/cycles to avoid stale files
```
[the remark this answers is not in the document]
0xVolosnikov: [AI] Resolved on commit 27b16571 (already in the current tree, not in this push). `run_precompiles` now writes samples/cycles to `precompile_samples/test_precompiles` and `precompile_cycles/test_precompiles` subdirs and only `rm -rf`s its own subdir, so the per-block artifacts `run_all_blocks` already wrote under `precompile_{samples,cycles}/block_*` survive.

bench_scripts/bench.sh
```diff
@@ -207,6 +270,28 @@ do_compare() {
             "${cycle_args[@]}" --gas-stats "${gas_args[@]}" --sample-dirs "${cycle_sample_args[@]}" \
             2>/dev/null || true
     fi
+    # Aggregate per-precompile stats across test-crate + all block benchmarks.
+    local precompile_stats_args=()
+    if [ -f "$BASELINE_DIR/precompile_stats.csv" ] && [ -f "$CURRENT_DIR/precompile_stats.csv" ]; then
+        precompile_stats_args+=(
+            "$BASELINE_DIR/precompile_stats.csv"
+            "$CURRENT_DIR/precompile_stats.csv"
```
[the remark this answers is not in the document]
0xVolosnikov: [AI] Resolved on 27b16571. Both the writer (`run_precompiles`, `PRECOMPILE_STATS_PATH="$output_dir/precompile_stats.csv"`) and the reader (compare step, `BASELINE_DIR/precompile_stats.csv`) now use the same `precompile_stats.csv` filename.

.github/workflows/bench.yml
```diff
@@ -69,18 +69,27 @@ jobs:
             OPCODE_SAMPLES_DIR=$(pwd)/opcode_samples/base_${blk} \
             OPCODE_CYCLE_SAMPLES_DIR=$(pwd)/opcode_cycles/base_${blk} \
             MARKER_PATH=$(pwd)/base_block_${blk}.bench \
+            PRECOMPILE_STATS_PATH=$(pwd)/base_block_${blk}_precompile_stats.csv \
+            PRECOMPILE_SAMPLES_DIR=$(pwd)/precompile_samples/base_${blk} \
+            LABEL_CYCLE_SAMPLES_DIR=$(pwd)/precompile_cycles/base_${blk} \
             cargo run --manifest-path tests/instances/eth_runner/Cargo.toml --release --features rig/no_print,rig/cycle_marker,rig/unlimited_native -- single-run --block-dir "$dir" --opcode-stats > base_block_${blk}.out
           done
-          MARKER_PATH=$(pwd)/base_precompiles.bench cargo test --release --features rig/no_print,precompiles/cycle_marker,rig/unlimited_native -p precompiles -- test_precompiles
+          MARKER_PATH=$(pwd)/base_precompiles.bench PRECOMPILE_STATS_PATH=$(pwd)/base_precompile_stats.csv PRECOMPILE_SAMPLES_DIR=$(pwd)/base_precompile_samples LABEL_CYCLE_SAMPLES_DIR=$(pwd)/base_precompile_cycles cargo test --release --features rig/no_print,precompiles/cycle_marker,rig/unlimited_native -p precompiles -- test_precompiles
```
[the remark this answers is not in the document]
0xVolosnikov: [AI] Documented as best-effort in commit eb902a07. The behavior is intentional — same compat pattern the workflow already uses for `for-tests-benchmarking-pectra`, `bench-fast`, and `BENCH_DA_SCHEME`: when a merge-base predates the new instrumentation, the env vars are ignored, no base CSV is produced, and `compare_precompile_stats.py` skips rows for which a base CSV is missing. The degradation resolves on the next PR cycle once a merge-base contains the tracer code. Added an explicit comment next to the env vars so this fallback has an in-tree explanation instead of implicit.

bench_scripts/compare_opcode_stats.py
0xVolosnikov: [AI] Fixed in commit 02e18e6b.

cycle_marker/src/lib.rs
0xVolosnikov: [AI] Done in commit 02e18e6b — hoisted the two byte-identical `append` closures into one shared `append_samples` closure declared once before both `if let Ok(dir) = std::env::var(...)` blocks.

forward_system/src/system/tracers/precompile_stats.rs
0xVolosnikov: [AI] Promoted to `assert!` in commit 45ab62b3. The single-slot pending state is load-bearing for stats attribution; silently overwriting it would misattribute gas/native to the wrong precompile. The tracer only runs host-side as part of benchmarking, so panicking in release is strictly better than producing wrong numbers — added a comment by the assert documenting the rationale.

bench_scripts/compare_precompile_stats.py
0xVolosnikov: [AI] Real bug — fixed in commit 22a8a66f. The row-suppress filter now also checks `med_gas` and `med_native`, so a median-only regression with a flat max no longer vanishes from the comment.

.github/workflows/bench.yml
0xVolosnikov: [AI] Done in commit b092c4b8 — extracted both into:

- `bench_scripts/run_bench_side.sh <base|head>` collapses the two near-identical "Run benchmarks" blocks (the bench-base and bench-head versions only differed in artifact prefix + a head-only opcode-stats CSV) into one parameterized bash script.
- `bench_scripts/render_pr_comment.sh <output.md>` holds all the pair-list construction and compare-script invocations that built the PR comment.

Workflow file dropped from 520 to 269 lines. The bench-base / bench-head / compare jobs each have one-line `run:` blocks now. Bench output is byte-equivalent — the same shell, just moved into checked-in scripts that can be linted/tested locally.

antoniolocascio (approved):

PR #656, opened 2026-05-15
feat(bench): STF cycle markers for benchmarking

0xVolosnikov:
## What ❔

Adds four cycle-marker labels in the STF that the benchmarking pipeline relies on:

- `keccak_execution_environment` — wraps the EVM SHA3 opcode dispatch (incl. the `len == 0` empty-slice shortcut)
- `ecrecover_execution_environment` — wraps the EE-precompile ecrecover dispatch via a new `EcRecoverEEInvocation` wrapper
- `da_commitment` — wraps the per-block `write_pubdata` call (keccak-DA absorbs there; blob-DA appends + KZG fires later under the pre-existing `blob_versioned_hash` marker)
- `state_commitment_update` — renamed from `verify_and_apply_batch` to reflect what the marker actually wraps (the state-tree merkle commit, distinct from DA commit and per-blob KZG)

Also extracts `install_precompile_hook` from `add_precompile_ext` so the `PRECOMPILE_ADDRESSES_LOWS` sanity check stays centralized for any future custom invocation type.

## Why ❔

Split out from #645 so the STF instrumentation can be reviewed without wading through the CI workflow, joiner scripts, tracers, and docs that consume the markers. Pure observability: no state-transition behavior changes. The new markers fire only on EE-initiated invocations, so they don't double-count the bootloader's intrinsic keccak/ecrecover calls. The benchmarking pipeline (CI + bench_scripts + forward-system tracers + docs) follows in a separate PR rebased onto this branch.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/basic_bootloader/src/bootloader/block_flow/ethereum/post_tx_op_proving.rs
+++ b/basic_bootloader/src/bootloader/block_flow/ethereum/post_tx_op_proving.rs
@@ -119,9 +119,9 @@ where
             &initial_state_commitment
         );
 
-        // // 3. Verify/apply reads and writes
+        // 3. Verify/apply reads and writes — state-tree merkle commit.
         let mut updated_state_commitment = initial_state_commitment;
-        cycle_marker::wrap!("verify_and_apply_batch", {
+        cycle_marker::wrap!("state_commitment_update", {
             io.update_commitment(
                 Some(&mut updated_state_commitment),
                 &mut logger,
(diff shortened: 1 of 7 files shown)


0xVolosnikov:
@antoniolocascio addressed the review feedback

antoniolocascio (approved):

PR #661, opened 2026-05-15
perf(point_evaluation): precompute tau*G2 prepared + fuse G1 scalar muls

0xVolosnikov:
## What ❔

Two follow-up optimizations on top of the rearranged KZG verifier (PR #660). Each commit is one optimization; benched A/B at the same parent commit using `test_kzg_regression` under `ZKSYNC_RISC_V_RUN=true` with the `for-tests-benchmarking` RISC-V binary.

### Commit 1 — precompute `PREPARED_G2_BY_TAU` as a const

τ·G2 is the fixed KZG trusted-setup point. Every call to `verify_kzg_proof` previously ran the 68-step BLS12-381 Miller-loop coefficient precomputation (63 G2 doublings + 5 G2 adds in Fq²) on it. Moved to compile time as a `const PREPARED_G2_BY_TAU`. Same pattern as the upstream `PREPARED_G2_GENERATOR` const in airbender-crypto, applied to τ·G2.

Two cfg-gated variants:
- host (`#[cfg(not(feature = "proving"))]`): 6-limb `ark_bls12_381::Fp`
- proving (`#[cfg(feature = "proving")]`): 8-limb `ark_ff_delegation::Fp` (delegation-aligned)

A `prepared_g2_by_tau_const_matches_runtime` test asserts byte-equality vs `G2_BY_TAU_POINT.into()` in each variant — catches stale literals if upstream τ·G2 or the Miller-loop precomputation shape ever changes.

### Commit 2 — fuse the two G1 scalar mults into a 2-base interleaved double-and-add

The rearranged verifier still did `y * G1_gen` and `z * proof` as two independent `sw_double_and_add_projective` calls — ~510 doublings total. Fused into a single interleaved loop over the 256 bits of the two scalars (~255 doublings + ~256 conditional adds).

Avoids arkworks' `VariableBaseMSM::msm_bigint`, which traps with an illegal instruction in the proving simulator because of internal allocation. Same pattern as the existing small-N path in `basic_system::system_functions::bls12_381::msm` — the local hand-rolled loop is allocation-free.

Negation of `z` (to fold the subtraction into an MSM-style positive-scalar form) is done once via `Fr::from_bigint(z).neg().into_bigint()`. `z` is canonical by upstream `parse_scalar` / `Fr::into_bigint`, so the `from_bigint` `Option` is statically safe to unwrap.

## Why ❔

Cumulative effect on `point_evaluation` cycles:

| Metric | Rearranged (#660) baseline | + τG2 const | + fused 2-base mul |
|---|---:|---:|---:|
| `point_evaluation` raw cycles | 31,791,155 | 29,615,041 (−6.84%) | 29,391,408 (−7.55%) |
| `point_evaluation` bigint delegations | 2,513,437 | 2,330,478 (−7.28%) | 2,308,496 (−8.15%) |
| effective (raw + 4·bigint) | 41,844,903 | 38,936,953 (−6.95%) | 38,625,392 (−7.69%) |
| `dist/for_tests/app.bin` | 1,337,196 B | 1,343,276 B (+6,080 B) | 1,340,044 B (+2,848 B) |

vs the **original pre-rearrangement verifier**: −24.77% effective.

The bigger of the two wins is the τG2 const — about 6 KB binary cost for ~7% cycle savings, which more than pays back compared to the residual binary headroom (still under the original baseline). The fused-mul is a smaller win (~0.8%) but is also a net binary-size **reduction** because it replaces two inlined `mul_bigint` instantiations with one short loop.

The diminishing returns reflect that the residual cost is now dominated by the Miller loop + final exponentiation — primitives that live inside airbender-crypto and aren't reachable from this codepath.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

- [x] PR title corresponds to the body of PR.

## Stacked on

Based on `vv/kzg-rearranged-verifier` (PR #660). Lands cleanly on top once #660 merges.

## Follow-ups not in this PR

- The τG2 const should eventually move into airbender-crypto next to `PREPARED_G2_GENERATOR`, so other consumers benefit and the const lives next to the source of truth.
- Investigate why `VariableBaseMSM::msm_bigint` traps in the proving simulator; if fixable, the small-N hand-rolled loop could be replaced by the upstream Pippenger path with its 3-bit windows for an additional ~2-3% in this codepath. (Probably an allocator quirk, since the precompiles crate hits the same issue.)


0xVolosnikov:
Superseded by [matter-labs/airbender-platform#80](https://github.com/matter-labs/airbender-platform/pull/80), which lands all three optimizations (rearranged G1-mul verifier, fused 2-base scalar mul, and the precomputed `PREPARED_G2_BY_TAU` const) directly in `airbender-crypto::bls12_381::verify_kzg_proof` — the source of truth that ZKsync OS already calls. Once that merges and the `airbender-crypto` rev is bumped here, point_evaluation picks up the speedup with no local code change needed. Closing in favor of the upstream port.

PR #675, opened 2026-05-26
fix: l1 txs pubdata charging overflow

AntonD3:
## What ❔

Use constant native_per_gas for all l1 -> l2 txs (including upgrade, service txs)

## Why ❔

- predictable charging for l1 -> l2 txs
- does not depend on base token and l1 gas price
- native limit u64 overflow is not possible, so whole native limit can be used on pubdata, and maximal amount of pubdata does not depend on gas price
- still big enough value to make computational native part negligible

## Is this a breaking change?
- [x] Yes
- [ ] No

## Checklist


tests/instances/transactions/src/l1_tx_resilience.rs
```diff
@@ -219,6 +218,90 @@ fn test_l1_tx_fee_independent_of_block_base_fee() {
...
+    // ABI calldata for `sendToL1(bytes)`: selector || offset(0x20) || length || data || pad
+    let mut calldata = Vec::with_capacity(4 + 64 + payload_len.next_multiple_of(32));
+    calldata.extend_from_slice(&hex::decode("62f84b24").unwrap()); // selector
+    calldata.extend_from_slice(&[0u8; 32 - 1]); // offset upper 31 bytes
+    calldata.push(0x20); // offset = 32
+    calldata.extend_from_slice(&[0u8; 32 - 8]); // length upper 24 bytes
+    calldata.extend_from_slice(&(payload_len as u64).to_be_bytes());
+    calldata.extend_from_slice(&payload);
+    let padding = (32 - (payload_len % 32)) % 32;
+    calldata.extend_from_slice(&vec![0u8; padding]);
```
antoniolocascio: Can you use alloy_sol_types and the macros to encode this?

antoniolocascio (approved):
Left only a nit, LGTM

tests/instances/transactions/src/native_charging.rs
```diff
@@ -54,24 +54,6 @@ fn run_tx(
...
-    let tx = L1TxBuilder::new()
-        .from(wallet.address())
-        .to(TO)
-        .gas_price(gas_price.into())
-        .gas_limit(70_000)
-        .input(hex::decode(ERC_20_TRANSFER_CALLDATA).unwrap())
-        .build()
-        .into();
-    run_tx(tx, gas_price, native_price, false, false)
-}
```
antoniolocascio, line 73: Seems like these tests are now outdated, but why remove this one?

antoniolocascio (approved):
LGTM

tests/instances/transactions/src/native_charging.rs
AntonD3: It's no longer possible to configure ratio for l1 txs(it's hardcoded now), this test used low ratio and expected l1 tx to fail, which is obviously no longer a case and there is no way to rewrite it

PR #683, opened 2026-06-02
feat: add runtime chain config

Raid5594:
## What ❔

Introduces a runtime `ChainConfig` that captures chain-level execution rules. It is read once per run from the oracle, frozen for the whole batch, and committed into the batch public input so proofs bind to the rules actually used.

`ChainConfig` fields (defaults preserve current mainnet behavior):
- `chain_id: u64` — moved here from per-block metadata.
- `fri_proof_verification_enabled: bool` — gates FRI proof verification. Default: off.
- `max_tx_gas_limit: u64` — EIP-7825 single-transaction gas cap; effective per-tx limit is `min(block_gas_limit, max_tx_gas_limit)`. Floored at and defaults to `1 << 24`.

The config is committed in `BatchPublicInput` as a keccak256 hash (`chain_config_hash`), replacing the previous per-block `chain_id`/`is_gateway` plumbing.

## Why ❔

FRI proof verification is only needed for Gateway chains, while the per-tx gas limit may be configured by customer chains. Making these runtime config — instead of per-block metadata or build-time flags — lets one binary serve different chains, with the public input pinning the exact rules each batch was executed under.

## Note: unrelated change bundled in

This PR also carries one change unrelated to the chain-config feature, in `forward_system/src/system/system_types/ethereum.rs`: the proving-mode Ethereum-storage post-op (`EthereumPostOp<_, true>`) was switched from `VecCtor` to `BiVecCtor`, aligning the forward (sequencer) side with the proving guest, which already uses `BiVecCtor` (see `proof_running_system`). This only changes the scratch-buffer allocation strategy used while building the Ethereum block's trie roots — it does not change the computed roots/output — and it affects only the experimental Ethereum-storage path (L1 re-execution / revm consistency checking), not the ZKsync rollup STF. It originally landed under a `fix: wip` commit; flagging it here since it is logically separate from this PR.

## Is this a breaking change?
- [x] Yes
- [ ] No

The batch public-input preimage layout changes: `chain_id` is removed from the `BatchOutput` commitment, and a `chain_config_hash` (keccak256 of `chain_id`, `fri_proof_verification_enabled`, `max_tx_gas_limit`) is committed in `BatchPublicInput`. The settlement-layer/verifier contracts must be updated to match the new layout.

## Checklist


basic_bootloader/src/bootloader/config.rs
```diff
@@ -8,6 +13,18 @@ pub trait BasicBootloaderExecutionConfig: 'static + Clone + Copy + core::fmt::De
     const VERIFY_FRI_PROOFS: bool;
 }
 
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct BootloaderStaticConfig {
+    pub chain_config: ChainConfig,
+}
+
+impl BootloaderStaticConfig {
```
Raid5594: Do you think the naming is fine?

zk_ee/src/system/metadata/chain_config.rs
```diff
@@ -0,0 +1,195 @@
...
+use crate::system::errors::internal::InternalError;
+use crate::utils::exact_size_chain::ExactSizeChain;
+
+/// Default EIP-7825 single-transaction gas limit (2^24).
+pub const DEFAULT_MAX_TX_GAS_LIMIT: u64 = 1 << 24;
+
+/// Static chain-level execution rules committed into the batch public input.
+///
+/// Changing this value is a protocol-upgrade boundary and batches must not span
+/// configurations.
```
AntonD3: nit: comment is not true - it's not static, and can change not only with protocol upgrade, e.g. fri_proof_verification_enabled with migration, max_tx_gas_limit with chain admin action

zk_ee/src/system/metadata/chain_config.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    pub const fn fri_proof_verification_enabled(&self) -> bool {
+        self.fri_proof_verification_enabled
+    }
+
+    pub const fn max_tx_gas_limit(&self) -> u64 {
+        self.max_tx_gas_limit
+    }
+
+    pub fn validate(&self) -> Result<(), InternalError> {
```
AntonD3, line 114: Is it really needed?

zk_ee/src/system/metadata/chain_config.rs
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+        let raw = <RawChainConfig as serde::Deserialize>::deserialize(deserializer)?;
+        let config = Self {
+            chain_id: raw.chain_id,
+            fri_proof_verification_enabled: raw.fri_proof_verification_enabled,
+            max_tx_gas_limit: raw.max_tx_gas_limit,
+        };
+        config
+            .validate()
```
AntonD3: Given that validate is no op, do we need this whole custom `deserialize` at all?

zk_ee/src/system/metadata/chain_config.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    pub const fn default_for_chain() -> Self {
+        Self {
+            chain_id: 0,
+            fri_proof_verification_enabled: false,
+            max_tx_gas_limit: DEFAULT_MAX_TX_GAS_LIMIT,
+        }
+    }
+
+    pub const fn with_chain_id(mut self, chain_id: u64) -> Self {
```
AntonD3: nit, but IMHO such "tests-only" setters is not nice in this struct. Maybe you can just init config in tests, or make fields public

zk_ee/src/system/metadata/chain_config.rs
```diff
@@ -0,0 +1,195 @@
...
+            chain_id,
+            fri_proof_verification_enabled,
+            max_tx_gas_limit,
+        };
+        config.validate()?;
+
+        Ok(config)
+    }
+
+    pub const fn default_for_chain() -> Self {
```
AntonD3, line 71: Would be nice to add some comments, is it for tests only?

zk_ee/src/system/metadata/zk_metadata.rs
```diff
@@ -13,11 +15,12 @@ use crate::{
 };
 use ruint::aliases::{B160, U256};
 
-pub type ZkMetadata = SystemMetadata<
-    EthereumIOTypesConfig,
-    BlockMetadataFromOracle,
-    TxLevelMetadata<EthereumIOTypesConfig>,
->;
+pub struct ZkMetadata {
```
AntonD3: Why not wrap SystemMetadata as before?

zk_ee/src/system/metadata/chain_config.rs
```diff
@@ -0,0 +1,195 @@
...
+            max_tx_gas_limit,
+        };
+        config.validate()?;
+
+        Ok(config)
+    }
+}
+
+/// Metadata types that expose static chain-level execution configuration.
+pub trait ChainConfigMetadata {
```
AntonD3: Probably better to move to the `basic_metadata.rs` where all similar traits are defined

zk_ee/src/system/metadata/system_metadata.rs
```diff
@@ -15,6 +16,9 @@ pub struct SystemMetadata<
     /// Provider of metadata for the current transaction.
     pub tx_level: TX,
 
+    /// Static chain-level configuration.
+    pub chain_config: super::chain_config::ChainConfig,
```
AntonD3: nit: for consistency with other fields maybe it's better to have a generic type which implements chain metadata trait here?

AntonD3: TBH I am not sure if we really need all these generics, but in my opinion, it's better to either keep it consistent with current approach for other fields, or change other fields as well.

zk_ee/src/system/metadata/zk_metadata.rs
```diff
@@ -172,15 +266,9 @@ impl BasicBlockMetadata<EthereumIOTypesConfig> for BlockMetadataFromOracle {
     }
 
     fn individual_tx_gas_limit(&self) -> u64 {
```
AntonD3: Maybe remove this method fully to avoid confusion, and use new method from chain metadata instead

zk_ee/src/system/metadata/zk_metadata.rs
```diff
@@ -46,6 +49,106 @@ impl BasicTransactionMetadata<EthereumIOTypesConfig> for TxLevelMetadata<Ethereu
     }
 }
 
+impl BasicBlockMetadata<EthereumIOTypesConfig> for ZkMetadata {
```
AntonD3: Not nice code duplication. I think if you define `ZkMetadata` as it was before, you'll not have to reimplement all these traits

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/public_input.rs
```diff
@@ -50,8 +51,9 @@ impl ChainStateCommitment {
 ///
 #[derive(Debug)]
 pub struct BatchOutput {
-    /// Chain id used during execution of the blocks.
-    pub chain_id: U256,
+    /// Static chain-level execution rules used during execution (includes the
+    /// chain id).
+    pub chain_config: ChainConfig,
```
AntonD3: I would move it to the upper level - `BatchPublicInput`

AntonD3 (commented):
Left some comments, also @Raid5594 maybe we can add `da_commitment_scheme` and `sl_chain_id` to this config as well, WDYT?

zk_ee/src/system/metadata/chain_config.rs
0xVolosnikov: I think that we should have some limitation, but it doesn't belong to deserialization

0xVolosnikov:
@AntonD3 I remember that server side prefer not to pass `da_commitment_scheme` during the sequencing step, so it will be easier for them to have it as a separate oracle query

forward_system/src/run/mod.rs
```diff
@@ -113,10 +117,43 @@ pub fn run_block<
     tx_result_callback: TR,
     tracer: &mut impl Tracer<ForwardRunningSystem>,
     validator: &mut impl TxValidator<ForwardRunningSystem>,
+) -> Result<BlockOutput, ForwardSubsystemError> {
+    run_block_with_chain_config(
+        ChainConfig::default(),
+        block_context,
```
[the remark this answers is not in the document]
0xVolosnikov: Fixed in 71098d1f. Removed the defaulting wrapper entirely and promoted `run_block_with_chain_config` to `run_block`, so `chain_config: ChainConfig` is now a required argument. Callers can no longer get a silent `chain_id = 0`; the canonical sequencer path (the `RunBlock` trait impl) already passed it explicitly. `api::run_block` keeps its name but now requires the config.

forward_system/src/run/mod.rs
```diff
@@ -164,10 +202,44 @@ pub fn generate_proof_input<
     fri_proof_sidecar: FS,
     fri_verifier_artifacts: Option<Arc<FriVerifierArtifacts>>,
     tx_result_callback: TR,
+) -> Result<(Vec<u32>, BlockOutput, Vec<u8>), ForwardSubsystemError> {
+    generate_proof_input_with_chain_config(
+        ChainConfig::default(),
+        block_context,
```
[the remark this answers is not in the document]
0xVolosnikov: Fixed in 71098d1f. `generate_proof_input_with_chain_config` is now `generate_proof_input` with `chain_config` as a required argument; the defaulting wrapper is gone. `api::generate_proof_input` keeps its name but now requires an explicit `ChainConfig`.

forward_system/src/run/mod.rs
```diff
@@ -712,6 +939,26 @@ pub fn simulate_tx<S: ReadStorage, PS: PreimageSource>(
     preimage_source: PS,
     tracer: &mut impl Tracer<CallSimulationSystem>,
     validator: &mut impl TxValidator<CallSimulationSystem>,
+) -> Result<TxResult, ForwardSubsystemError> {
+    simulate_tx_with_chain_config(
+        ChainConfig::default(),
+        transaction,
```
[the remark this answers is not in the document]
0xVolosnikov: Fixed in 71098d1f. Same treatment: `simulate_tx_with_chain_config` is now `simulate_tx` with a required `chain_config`, and the defaulting wrapper is removed, so `eth_call`/`eth_estimateGas` paths can no longer silently run with `chain_id = 0`.

tests/fuzzer/fuzz/fuzz_targets/precompiles/common/mod.rs
```diff
@@ -16,7 +16,10 @@ use rig::zksync_os_tests_common::zksync_tx::encoding::ZKsyncOsEncodable;
 ///
 /// The second call is just there to check consistency between forward and proof runs.
 pub fn run_precompile(id: &str, input: &[u8]) -> BlockOutput {
-    let gas = 1 << 27;
+    // Keep at or below the default ChainConfig per-tx gas cap (EIP-7825,
+    // `max_tx_gas_limit = 1 << 24`); a higher limit is rejected during
+    // validation. Ample for a single precompile call plus the forwarder.
+    let gas = 1 << 24;
```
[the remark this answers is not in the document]
0xVolosnikov: Fixed in 71098d1f. Now imports and uses `DEFAULT_MAX_TX_GAS_LIMIT` from `zk_ee::system::metadata::chain_config` instead of the hard-coded `1 << 24`, matching how the `tests/instances/precompiles` harness already references it.

0xVolosnikov (approved):

PR #684, opened 2026-06-02
feat: use simple merkle tree for transactions/receipts root in block header

AntonD3:
## What ❔

  Computes the block header's transactions_root / receipts_root as depth-32 Blake2s Merkle trees over the per-block tx/receipt hashes (was a keccak rolling hash), plus supporting cleanup:

  - Merkle: new merkle_tree::merkle_root_in_place (fixed-height, in-place); L2→L1 logs root reuses it. Empty-subtree hashes are hardcoded, test-locked tables.
  - Roots: ZKBasicBlockDataKeeper stores tx/receipt hashes in ListVec and builds the roots from the system allocator (no global allocator).
  - Bloom: LogsBloom moved to a shared module;
  - RLP: consolidated into one bootloader::rlp module (streaming encoders over WriteBytes); removed duplicated helpers.
  - Test: eth_runner::post_check validates the Merkle transactions_root via the STF helper.

## Why ❔

Expose openable Ethereum-style tx/receipts roots instead of an opaque rolling hash, with a single Merkle + RLP implementation shared across paths .

## Is this a breaking change?
- [x] Yes
- [ ] No

## Checklist


AntonD3:
Missed global allocator usage(with vector) - fixing now, that's while CI failed

basic_bootloader/src/bootloader/block_flow/zk/tx_loop.rs
```diff
@@ -243,3 +267,107 @@ where
...
+    let bloom = LogsBloom::default();
+
+    let mut receipt_encoder = ReceiptEncoder::new_from_fields(
+        tx_type,
+        status,
+        cumulative_gas_used,
+        bloom.as_bytes(),
+        events,
+    );
+    let mut receipt_hasher = Blake2s256::new();
```
0xVolosnikov: Nit: we could initialize it once outside and pass into this helper, may save some cycles. Not sure if it worth it though.

basic_bootloader/src/bootloader/block_flow/zk/tx_loop.rs
```diff
@@ -243,3 +267,107 @@ where
         Ok(())
     }
 }
+
+fn compute_receipt_hash<'events, I>(
```
0xVolosnikov: Tbh I'm not sure that this is the best place for this function, maybe at least create a separate module for that?

0xVolosnikov:
Would be nice to add property tests / compare with canonical rlp implementation

0xVolosnikov:
Some additional AI review comments:


### Correctness

1. `receipts_root` is not independently validated by `eth_runner`.

   The PR adds ZK receipt hashes and seals `receipts_root` into the block header, but `tests/instances/eth_runner/src/post_check.rs` only reconstructs and checks `transactions_root`. This means a bug in receipt hash encoding, receipt ordering, cumulative gas, log inclusion, or even a constant/empty `receipts_root` could pass the runner post-check as long as `transactions_root` is correct.

   Please add an independent `receipts_root` reconstruction in the post-check, using the same receipt layout as `compute_receipt_hash`: status, cumulative gas used, zero bloom, logs, then Blake2s leaf hash and the simple Merkle tree root.

2. `merkle_root_in_place` is public but relies on unchecked preconditions.

   `zk_ee/src/common_structs/merkle_tree.rs` exposes a safe public function that assumes:
   - `empty_subtree_hashes` is non-empty
   - `nodes.len() <= 2^height`

   If `empty_subtree_hashes` is empty, `len() - 1` underflows. If too many leaves are passed, the function silently returns a root after folding only `height` levels, which is not a valid fixed-height tree root.

   Current call sites appear bounded, but this is consensus-root infrastructure exported from `zk_ee`, so I would prefer explicit assertions or a fallible API.

3. Receipt tests should cover ZKsync-specific transaction types.

   `ReceiptEncoder` prefixes every nonzero `tx_type`, but tests only cover `0, 1, 2`. The ZK path can feed non-Ethereum/ZKsync-specific type bytes like `0x7c`, `0x7d`, `0x7e`, and `0x7f`.

   If typed-prefix behavior is the intended contract for these receipt roots, please add explicit tests and a short comment documenting that. Otherwise this is easy to misinterpret later as “Ethereum receipt encoding only”.

### Code quality / maintainability

4. `A::default()` for scratch allocation is a little suspicious in generic allocator code.

   `block_tx_tree_root` and `LogsStorage::tree_root` allocate scratch buffers with `A::default()` rather than preserving allocator provenance from the original structure. This is probably fine for the current allocators, but it weakens the generic allocator contract and is harder to audit in proving/no-std paths.

5. `LogsBloom` uses avoidable unsafe for a 256-byte field.

   I do not see a concrete UB issue in the current implementation, but the byte reinterpretation would be easier to audit if the storage were `[u8; 256]`, or at minimum if there were a known-vector test against Ethereum bloom bit positions.

basic_bootloader/src/bootloader/block_flow/zk/tx_loop.rs
antoniolocascio: Yeah, moved it out

antoniolocascio:
@0xVolosnikov resolved the comments. Added proptest for RLP.

0xVolosnikov (approved):

PR #686, opened 2026-06-09
feat(pricing): re-estimate native model costs for EVM opcodes and precompiles

vibelyova:
## What ❔

Re-estimates native resource costs for EVM opcodes and precompiles based on measured RISC-V effective cycle data from 10 mainnet block benchmarks.

Changes:
- **Fixed-cost EVM opcodes**: all constants in `native_resource_constants.rs` updated to match max observed effective cycles (rounded up to 10s). Both undercharged (e.g. SELFBALANCE +583) and overcharged (e.g. MUL -360, JUMPI -212) opcodes corrected.
- **Copy opcodes**: base costs bumped to 200 for CALLDATACOPY/RETURNDATACOPY/CODECOPY/EXTCODECOPY, new MCOPY constant added. Shared per-byte cost unchanged.
- **BLOBHASH/BLOBBASEFEE**: replaced hardcoded inline values with named constants at measured values.
- **SHA3/keccak**: fixed off-by-one in round counting (`ceil(len/136)` → `ceil((len+1)/136)`) that undercharged inputs at exact chunk boundaries.
- **Modexp**: replaced `gas * constant` model with `BASE + total_ops * PER_OP + total_ops * d² * PER_D²` where total_ops = squares + multiplies from scanning actual exponent bits (popcount). Eliminates 4 undercharge violations and reduces median overcharge from 6x to 1.6x.
- **bn254_pairing**: removed base cost (was equal to per-pair, fit showed negative base), lowered per-pair from 15M to 6.6M effective cycles.
- **ecrecover**: lowered from 522K to 368K effective cycles.
- **ripemd160**: lowered base from 1600 to 800, per-round from 4200 to 2500.
- **Dead code**: removed unused STOP_NATIVE_COST and INVALID_NATIVE_COST constants (these opcodes never charge via handler).
- **OOG filtering**: tracer now skips recording samples for OOG executions (partial native charges that distort benchmarks).

## Why ❔

Native resource costs must upper-bound actual RISC-V effective cycles to ensure the prover can always complete within budget. Several opcodes were undercharged (security risk) and many were significantly overcharged (unnecessarily expensive for users). This PR calibrates all costs against measured cycle data.

## Is this a breaking change?
- [x] Yes
- [ ] No

Native resource consumption changes for all transactions. Gas estimation results will differ.

## Checklist

- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.
[...]

PR #688, opened 2026-06-15
feat(pricing): re-estimate native model costs for EVM opcodes and precompiles

0xVolosnikov:
> Rebased version of #686 onto the latest `draft-0.4.0` (original branch `pricing-evm-opcodes` based at `662b83bc`, 3 commits behind). The merge commit was dropped in favor of a clean linear history; the only overlap with new draft commits (PUSH3..PUSH8 / MULMOD specialization in `interpreter.rs`) was in a disjoint function and merged cleanly. All original commits and authorship are preserved.

## What ❔

Re-estimates native resource costs for EVM opcodes and precompiles based on measured RISC-V effective cycle data from 10 mainnet block benchmarks.

Changes:
- **Fixed-cost EVM opcodes**: all constants in `native_resource_constants.rs` updated to match max observed effective cycles (rounded up to 10s). Both undercharged and overcharged opcodes corrected (e.g. MUL lowered). NOTE: some of these were further adjusted in the follow-up below (e.g. JUMPI and SELFBALANCE were raised after the 10-block sweep).
- **Copy opcodes**: base costs bumped to 200 for CALLDATACOPY/RETURNDATACOPY/CODECOPY/EXTCODECOPY, new MCOPY constant added. Shared per-byte cost unchanged.
- **BLOBHASH/BLOBBASEFEE**: replaced hardcoded inline values with named constants at measured values.
- **SHA3/keccak**: fixed off-by-one in round counting (`ceil(len/136)` → `ceil((len+1)/136)`) that undercharged inputs at exact chunk boundaries.
- **Modexp**: replaced `gas * constant` model with `BASE + total_ops * PER_OP + total_ops * d² * PER_D²` where total_ops = squares + multiplies from scanning actual exponent bits (popcount). Eliminates 4 undercharge violations and reduces median overcharge from 6x to 1.6x.
- **bn254_pairing**: re-fit against measured effective cycles as `base + per_pair·num_pairs` (base ~6.24M = one-time final exponentiation, per-pair ~6.9M, both lowered from the original 15M/pair). _An earlier revision of this PR dropped the base entirely; the 10-block sweep showed that under-charges low pair counts (1 pair = 1.89×), so it was re-introduced — see the follow-up section._
- **ecrecover**: lowered from 522K to 368K effective cycles.
- **ripemd160**: lowered base from 1600 to 800, per-round from 4200 to 2500.
- **Dead code**: removed unused STOP_NATIVE_COST and INVALID_NATIVE_COST constants (these opcodes never charge via handler).
- **OOG filtering**: tracer now skips recording samples for OOG executions (partial native charges that distort benchmarks).

## Why ❔

Native resource costs must upper-bound actual RISC-V effective cycles to ensure the prover can always complete within budget. Several opcodes were undercharged (security risk) and many were significantly overcharged (unnecessarily expensive for users). This PR calibrates all costs against measured cycle data.

## Is this a breaking change?
- [x] Yes
- [ ] No

Native resource consumption changes for all transactions. Gas estimation results will differ.

## Checklist

---

## Follow-up: cycles/native validation + recalibration (10-block sweep)

After rebasing, the costs were validated end-to-end with the repo's own tooling: a proving-mode (RISC-V) run over all 10 `eth_runner` block fixtures plus the `precompiles` test crate, fed to `bench_scripts/cycles_per_native_report.py`. The ratio is `effective_cycles / charged_native` per execution; **ratio > 1 means the native budget is below actual proving cost** (a soundness gap). This surfaced several residual undercharges and one over-charge, fixed in the follow-up commit:

**Undercharges fixed** (native now upper-bounds the max observed effective cycles, with margin):
- **JUMPI** 90→100 — taken branch was a flat 1.08×.
- **SHA3** — evm `KECCAK256_NATIVE_COST` 90→120 (empty-input path) and the shared `KECCAK256_BASE_NATIVE_COST` 400→1150. The undercharge was a **fixed ~720-cycle per-keccak-call overhead** (1-round SHA3 was 1.16×, amortizing toward ~1.0); the per-round slope was already correct. ⚠️ `KECCAK256_BASE_NATIVE_COST` is shared with **intrinsic keccak** (tx hashing), so intrinsic per-tx native rises too.
- **bn254_pairing** — **re-introduces the fixed base** (final exponentiation) that this PR's earlier revision had removed. The 10-block data shows a clear `base + per_pair·pairs` shape (≈5.95M + 6.58M·pairs); the per-pair-only model under-charged low pair counts (1 pair was **1.89×**). Now `base 6.244M + per-pair 6.908M`, charged as `base + per_pair·num_pairs`.
- **bn254_ecadd** raw 46k→51.4k, **bn254_ecmul** raw 600k→647k.
- **~1.01 cluster** (exact max + ~5% margin): `MOD SMOD SDIV SHL SAR SIGNEXTEND BLOBHASH SELFBALANCE PUSH9/10/11/12/18/23/28`.

**Over-charge recovered:**
- **PUSH3..8** 140–170 → 60–70 — the `push_small` specialization (#648, now in the base branch) roughly halved their effective cycles, leaving them ~2× over-charged.

**Investigated, intentionally left as-is** (documented below):
- **ecrecover** (1.008) — deferred: `ECRECOVER_NATIVE_COST` feeds the intrinsic per-tx signature-verification budget, so a change has tx-wide blast radius and warrants separate review.
- **STOP** (1.05) — a deterministic **1-cycle** gap; closing it needs a spend in the dispatch hot loop (binary-size / I-cache risk) or bumping `STEP` (affects every opcode).
- _(Copy opcodes were here; now fixed — see below.)_

### Copy-opcode undercharge — root cause (now fixed)

The copy opcodes copy via `zk_ee::utils::copy_and_zeropad_nonoverlapping` → `copy_from_slice` (i.e. `memcpy`). On the proving RISC-V target, `memcpy` cost depends on the **relative alignment** of `src` and `dst`:
- `(src_addr − dst_addr) % word == 0` → word-at-a-time copy, ~0.5–0.7 effective cyc/byte
- otherwise → byte/shift fallback (no cheap unaligned word access), ~1.4+ cyc/byte

The opcodes take operand-controlled offsets (`CALLDATACOPY` src = calldata at `source_offset`, dst = heap at `memory_offset`), so per-call cost is effectively **bimodal**. Pricing uses a single flat `COPY_BYTE_NATIVE_COST = 1` per byte (`evm_interpreter::gas::copy_cost`), which sits between the two regimes — over-charging the aligned common case, under-charging the misaligned worst case.

**Evidence** — for a *fixed* copy length (fixed native), effective cycles split into discrete clusters by alignment class:

| native (size) | count | fast cluster | slow cluster | spread |
|--:|--:|--:|--:|--:|
| 368 | 1907 | ~253 | ~340 | 43% |
| 432 | 974 | ~289 | ~350–451 | 56% |
| 624 | 451 | ~410 | ~727–744 | 87% |
| 300 (len 0) | 771 | 96 only | — | 0% |

The empty copy (no `memcpy`) has zero spread → the variance is in the copy itself. High-ratio executions are pure copies with **no memory growth** (gas matches a plain copy), gap scales linearly (~+0.44 cyc/byte). It is **not** the base / per-opcode / heap-growth terms — the *largest* copies are within budget (~0.92×).

**Fix options (follow-up):** (1) raise `COPY_BYTE_NATIVE_COST` to the misaligned worst case (~1.5/byte; ~2× over-charges the aligned path), (2) charge alignment-aware on `(src − dst) % word`, or (3) make the copy alignment-insensitive so the per-byte cost is uniform — fixing the cause rather than the price.

**Resolution (commit `71f12e63`):** chose to **over-charge** — `COPY_BYTE_NATIVE_COST 1 → 2` (the misaligned worst case rounds up to 2). This guarantees native upper-bounds effective cycles for all copies at the cost of ~2× over-charging the aligned common case on the per-byte term. Measured impact: copy native is ~0.46% of block proving cost; the bump adds **~+0.14% block proving cost** and **~+0.17% native for an average tx** (a simple transfer far less; **0 gas** on free-native chains). The constant also feeds the L1/L2 intrinsic per-calldata-byte cost (same `memcpy` exposure), raising L2 per-calldata-byte 59→60 and L1 1→2 — also overcharge / safe direction.

All recalibrations were validated against the sweep cycle data (constants-only → cycles unchanged), and the first batch was confirmed with a full real 10-block re-run that matched the analytical projection.


0xVolosnikov:
Addressed all three findings in `6db4e698`:

**1. modexp charge-before-allocation (Major).** The operand allocations are now gated up front with `has_enough(ergs + conservative_native)` — both computable from the header lengths + `exp_highp` without materializing operands — before `try_with_capacity_in`/`read_padded`. So an input declaring large lengths with a minimal payload is rejected before any large allocation/zero-fill, restoring the charge-before-work ordering even when `eip-7823-modexp-limit` is off. The exact native (scans the materialized exponent, ≤ the conservative bound just checked) is still charged afterward, so pricing accuracy from the rework is preserved.

**3. Dead conservative `native_cost` (Minor).** Resolved together with #1 — it is now the pre-allocation gate, and made private (no external callers).

**2. FREE/non-FREE intrinsic split (Major, maintainability).** Collapsed into a shared `L2_TX_INTRINSIC_COMPUTATIONAL_NATIVE_COST_COMMON` plus the sender delta (cold read + persist), so the two modes can't drift. Values are unchanged — verified byte-identical to the original explicit expressions via temporary compile-time `assert!`s (build passed, then removed).

Verification: `basic_system`/`basic_bootloader` build clean, modexp unit tests 11/11 pass.

antoniolocascio (approved):
LGTM

PR #691, opened 2026-06-17
chore: enable Pectra+Fusaka by default and remove their feature flags

0xVolosnikov:
## What ❔

ZKsync OS v0.4.0 targets the **Fusaka** hardfork (which includes Pectra), so the optional `pectra`/`fusaka` Cargo features — and all of their EIP/precompile constituents — are removed and the gated behavior is made unconditional. The pre-fusaka `#[cfg(not(...))]` code paths are deleted.

Removed feature flags (now always-on):
- **Aggregators:** `pectra`, `fusaka`, `evm_tester_pectra`, `evm_tester_fusaka`
- **EIPs:** `eip-2935`, `eip-7702`, `eip-7623`, `eip-4844`, `eip-7918`, `eip-7934`, `eip-7823-modexp-limit`, `modexp-repricing`
- **Precompiles/opcodes:** `bls12_381`, `blake2f`, `p256_precompile`, `point_eval_precompile`, `clz`
- **`mock-unsupported-precompiles`** (+ `mock_precompiles.rs`): only mocked BLAKE2F / point-eval, both now always registered → dead.

### Blob schedule kept behind `fusaka-bpo-2`

The blob-count schedule is **not** collapsed: it stays selectable via a dedicated `fusaka-bpo-2` flag, because the EF execution-spec-test fixtures for `fork_Osaka` assume the base-Osaka schedule while production targets the BPO2 schedule.
- **Default (off):** base-Osaka — `MAX_BLOBS_PER_BLOCK = 9`, target 6, update fraction 5007716.
- **`fusaka-bpo-2`:** BPO2 — 21 / 14 / 11684671.

It is attached to the `production` feature, so the singleblock/multiblock binaries and `for_tests` (forward run + its paired proof binary) run BPO2 consistently, while the forward-only EF EVM tester stays on base-Osaka and matches the EF Osaka fixtures.

### Other changes
- Build profiles (`production`/`for_tests`/`eth_stf`/`evm_tester`) drop their now-redundant per-EIP feature references.
- `dump_bin.sh` loses the `*-pectra`/`*-fusaka` variants except `evm-replay-benchmarking-fusaka`, which now enables `fusaka-bpo-2` for post-BPO bench replay. `ci.yml` drops the dedicated Pectra precompile step (`cargo test` covers those vectors by default).
- Tests updated: `precompiles` runs the BLAKE2F + BLS12-381 vectors unconditionally; `system_hooks` warm/cold probe now expects all Pectra precompiles (`0x09`, `0x0a`, `0x0b..0x11`, `0x100`) warm at tx start.

## Why ❔

This build *is* the Fusaka STF; the hardfork flags only added build configurations that are no longer used and complicated the code with dead pre-fusaka branches. The one genuinely variable parameter — the blob-count (BPO) schedule — is kept as `fusaka-bpo-2` so the EF Osaka conformance tests and the production schedule can coexist.

## Is this a breaking change?
- [x] Yes
- [ ] No

Removing the flag names is breaking for any external build that passes `--features pectra`/`fusaka` (Cargo hard-errors on unknown features). Downstream build invocations must drop those flags; production builds that want the BPO2 blob schedule pass `--features fusaka-bpo-2` (already implied by `production`).

## Checklist


tests/instances/precompiles/src/lib.rs
```diff
@@ -1275,34 +1275,27 @@ fn test_kzg_regression() {
...
-#[cfg(feature = "pectra")]
 const SCALAR_MEDIUM: &str = "73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001";
 
-#[cfg(feature = "pectra")]
 const SCALAR_WORST: &str = "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff";
 
 // Test vectors for precompiles gated behind the `pectra` feature (EIP-152 BLAKE2F and
 // EIP-2537 BLS12-381). Values come from the EIP specifications and well-known group identities
 // (G + 0 = G, G * 1 = G, e(0, 0) = 1). Exercised by `test_precompiles` when the `pectra`
 // feature is enabled.
```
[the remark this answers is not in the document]
0xVolosnikov: Good catch — fixed in 3cc2c1dd. The comment now states the PECTRA_TESTS vectors are always registered and always exercised by `test_precompiles`, matching the removal of the `#[cfg(feature = "pectra")]` gate.

tests/instances/eth_runner/Cargo.toml
```diff
@@ -35,11 +35,9 @@ airbender-host = { workspace = true }
...
 gpu = ["airbender-host/gpu-prover"]
 default = []
-# Forward the STF hardfork blob schedule so block.rs computes the blob base fee
-# with the matching update fraction. Pass these (not just rig/fusaka*) so the
-# eth_runner crate's own cfg sees the active schedule.
-fusaka = ["rig/fusaka"]
-fusaka-blobs = ["rig/fusaka-blobs"]
+# Selects the BPO2 blob-count schedule; default is base-Osaka. Forward this so
+# block.rs computes the blob base fee with the matching update fraction.
+fusaka-bpo-2 = ["rig/fusaka-bpo-2"]
```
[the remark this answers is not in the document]
0xVolosnikov: Fixed in 3cc2c1dd. The bench scripts (`bench_scripts/bench.sh`, `bench_scripts/run_bench_side.sh`) now detect `fusaka-bpo-2` to enable the post-BPO blob schedule on the host `eth_runner` run, with a fallback to the old `fusaka`/`fusaka-blobs` names so the A/B harness still builds correctly against a pre-rename merge-base.

basic_bootloader/src/bootloader/transaction/rlp_encoded/transaction_types/eip_4844_tx.rs
```diff
@@ -12,15 +12,7 @@ pub type BlobHashesList<'a> = FixedList<'a, &'a [u8; 32]>;
 
 /// EIP-4844 payload (type 0x03) layout: [chainId, nonce, maxPriorityFeePerGas, maxFeePerGas, gasLimit, to(20 bytes, not zero), value, data, accessList, maxFeePerBlobGas, blobVersionedHashes(32-byte items)]
 #[derive(Clone, Copy, Debug)]
-#[cfg_attr(not(feature = "eip-4844"), allow(dead_code))]
```
antoniolocascio, line 15: I don't think we should enable them in production, they make no sense for us

antoniolocascio: So, I think we should keep this feature flag

0xVolosnikov: Agreed — restored in ac2f0123. `eip-4844` is a feature flag again, gating blob-transaction (type 0x03) parsing, the blob-tx test, and the "blob base fee must be 1 when disabled" check. It's enabled only for the Ethereum-compatibility paths (`for_tests`/`eth_runner`/`evm_tester`, forward + proving) and **not** `production`, so production rejects blob transactions. The point-evaluation precompile (0x0a) is kept unconditional (general KZG verify usable by L2 contracts) rather than tied to `eip-4844`.

PR #614, opened 2026-04-06
chore: move build artifacts to dist/<app>/ layout

vibelyova:
## What ❔

Move RISC-V binary artifacts from flat files in `zksync_os/` (e.g. `for_tests.bin`) to a structured `dist/<app>/app.{bin,elf,text}` layout (e.g. `dist/for_tests/app.bin`).

## Why ❔

This prepares for the airbender-platform migration where `cargo airbender build` outputs to this layout natively. Splitting this out as a separate PR keeps the platform migration PR focused on the actual API changes.

## Changes

- `dump_bin.sh`: output to `dist/<app>/` instead of flat files
- `chain.rs`: add `get_zksync_os_dist_dir()`, update path resolution
- `binary_checker`: update paths to `dist/<app>/app.text`
- `reproduce.sh`: copy from `dist/<app>/app.bin`
- `.gitignore`: add `/dist`

## Is this a breaking change?
- [ ] Yes
- [x] No

The build script output changes but all consumers are updated in this PR.

## Checklist

- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [x] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.
- [x] Code has been formatted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #592, opened 2026-03-24
test(unit): add malicious oracle behavior tests

vv-dev-ai:
## What ❔

Add tests that verify the system correctly validates and rejects invalid/malicious oracle responses. Covers two oracle validation areas:

**Block metadata gas limit validation (4 tests):**
- `test_block_rejects_excessive_gas_limit` — gas limit just above `MAX_BLOCK_GAS_LIMIT` is rejected
- `test_block_accepts_max_gas_limit` — exact boundary value is accepted
- `test_block_rejects_u64_max_gas_limit` — extreme overflow value is rejected
- `test_empty_block_rejects_excessive_gas_limit` — validation fires before any transactions are processed

**DA commitment scheme validation (2 tests):**
- `test_da_commitment_scheme_accepts_all_valid_ids` — all valid IDs 0–4 are accepted
- `test_da_commitment_scheme_rejects_invalid_ids` — IDs 5, 128, 255 are rejected

## Why ❔

Oracle responses come from an untrusted host and must be validated. These tests confirm that existing validation logic correctly rejects malicious metadata. Gas limit validation prevents resource accounting overflow (ergs = gas × 256). DA commitment scheme validation prevents use of undefined commitment modes.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [x] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.
- [x] Code has been formatted.

PR #564, opened 2026-03-14
refactor(zk_ee): remove unused fields from WarmStorageValue

vv-dev-ai:
## What ❔

Removes four unused fields from `WarmStorageValue` in `zk_ee/src/common_structs/warm_storage_value.rs`:
- `value_at_the_start_of_tx`
- `changes_stack_depth`
- `last_accessed_at_tx_number`
- `pubdata_diff_bytes`

These fields were always initialized to their default values at all three construction sites and were never read anywhere in the codebase. The struct now contains only the four fields that are actually used by `StateRootView::verify_and_apply_batch`: `initial_value`, `current_value`, `initial_value_used`, and `is_new_storage_slot`.

The three construction sites (`storage_cache.rs`, `full_storage_cache.rs`, and `simple_growable_storage.rs`) are updated accordingly, removing the dead field assignments and the associated "temporarily using" comments that were present since the struct was initially over-designed.

## Why ❔

Removes dead code to make the storage cache abstraction boundary cleaner and easier to understand. The original struct was documented as a legacy type pending simplification.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.
- [x] Code has been formatted.

PR #543, opened 2026-02-27
feat(EVM-1167): cleanup storage types module

0xVolosnikov:
## What ❔
- Removed unused storage-type markers/bitflags.
- Split `storage_types` into focused modules.
- Moved `MAX_EVENT_TOPICS` from storage types into system constants and updated imports.

## Why ❔
- Reduces dead code and improves module structure.
- Places constants with the correct domain ownership for easier maintenance.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist
- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.
- [x] Code has been formatted.

PR #702, opened 2026-06-18
refactor(zk_ee): remove unused WarmStorageValue fields and rename to StorageSlotValue

0xVolosnikov:
## What ❔

Removes four unused fields from `WarmStorageValue` in `zk_ee/src/common_structs/warm_storage_value.rs`:
- `value_at_the_start_of_tx`
- `changes_stack_depth`
- `last_accessed_at_tx_number`
- `pubdata_diff_bytes`

These fields were always initialized to their default values at all three construction sites and were never read anywhere in the codebase. The struct now contains only the four fields that are actually used by `StateRootView::verify_and_apply_batch`: `initial_value`, `current_value`, `initial_value_used`, and `is_new_storage_slot`.

The three construction sites (`storage_cache.rs`, `full_storage_cache.rs`, and `simple_growable_storage.rs`) are updated accordingly, removing the dead field assignments and the associated "temporarily using" comments that were present since the struct was initially over-designed.

## Why ❔

Removes dead code to make the storage cache abstraction boundary cleaner and easier to understand. The original struct was documented as a legacy type pending simplification.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

---
Rebased from #564 onto the latest `draft-0.4.0`. The original branch was stale (its leading commits — #592, #543, #614 — are already merged), so I replayed only the 3 net commits (remove unused fields → rename `WarmStorageValue/Key`→`StorageSlotValue/Key` → drop dead test types) and dropped the no-op `repr(C)` add+revert pair. Resolved one conflict in `proof_running_system/src/tests.rs` (deletion of dead types). Builds, clippy, and zk_ee/proof_running_system tests pass.


0xVolosnikov:
"+0.44%" uhhh

antoniolocascio:
Let's skip this one for now

PR #584, opened 2026-03-21
feat(basic_system): optimize ecrecover by eliminating EncodedPoint overhead

vv-dev-ai:
## What

Optimize the ecrecover system function by eliminating unnecessary overhead in the
public-key-to-address conversion path:

1. **Skip `EncodedPoint` construction**: Add `Affine::to_xy_bytes()` that extracts raw
   x||y coordinate bytes (64 bytes) directly, bypassing the 65-byte `EncodedPoint`
   allocation, `ConditionallySelectable` infinity check, and the 0x04 prefix byte
   that is immediately discarded by the caller.
2. **Change `ecrecover_inner` return type** from `EncodedPoint` to `[u8; 64]` to
   avoid the intermediate encoded-point container.
3. **Replace iterator-chain output construction** (`repeat_n(0,12).chain(hash.skip(12))`)
   with direct array writes, avoiding iterator dispatch overhead.

## Why

The ecrecover path is one of the largest non-interpreter hotspots in the proving
flamegraph. The `ConditionallySelectable` check in `to_encoded_point` is pure
overhead since `recover_with_context` already checks for point-at-infinity and
returns an error. The iterator chain for output construction adds unnecessary
overhead per ecrecover call.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [x] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.
- [x] Code has been formatted.

PR #703, opened 2026-06-18
feat(basic_system): optimize ecrecover via Affine::to_xy_bytes()

0xVolosnikov:
## What

`ecrecover_inner` now returns the raw `x || y` coordinate bytes (`[u8; 64]`) via `crypto::secp256k1::Affine::to_xy_bytes()` instead of constructing a 65-byte `EncodedPoint`. The address-hash output is built with direct array writes instead of an iterator chain.

## Why

The ecrecover path is one of the largest non-interpreter proving hotspots. `to_encoded_point` runs a `ConditionallySelectable` infinity check and prepends a `0x04` prefix that the caller immediately discards — both pure overhead, since point-at-infinity is already checked inside `crypto::secp256k1::recover` / `recover_with_hooks`.

This is the **re-port** of the now-closed #584 (whose branch was on a stale base) onto current `draft-0.4.0`. The crypto-side `to_xy_bytes()` was moved to airbender-crypto: **matter-labs/airbender-platform#79**.

## Temporary pin

The four `airbender-platform` deps in `Cargo.toml` are temporarily pinned to the #79 branch commit (`bd6c4a2`) so this builds and tests today. **Before merge**, this must be repointed to a released `airbender-platform` rev once #79 lands.

- [ ] matter-labs/airbender-platform#79 merged
- [ ] `airbender-platform` rev repointed off the branch commit

## Verification

`cargo test -p basic_system --lib ecrecover` passes (5/5), including `test_geth_ecrecover` (known address derivation) and `test_point_of_infinity_in_result`. Behavior is unchanged.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

- [x] PR title corresponds to the body of PR.
- [x] Tests for the changes have been added / updated. (covered by existing ecrecover tests; behavior unchanged)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -76,8 +76,8 @@ dependencies = [
 
 [[package]]
 name = "airbender-codec"
-version = "0.2.2"
-source = "git+https://github.com/matter-labs/airbender-platform?rev=80c0541f6#80c0541f62414dc7f21c200218686c4bef9458d4"
+version = "0.2.3"
+source = "git+https://github.com/matter-labs/airbender-platform?rev=bd6c4a2992a0422b8f56a1dc74733e96a4a02b2b#bd6c4a2992a0422b8f56a1dc74733e96a4a02b2b"
 dependencies = [
  "bincode 2.0.1",
  "serde",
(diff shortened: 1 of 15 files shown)


shamatar:
Both crypto and this one look useless as conditionally selectable is replaced with assert of not-infinity, so net win is buffer size and 1 byte write at the hot path

PR #708, opened 2026-07-02
fix(bootloader): account for block pubdata framing

antoniolocascio:
## What ❔

This PR fixes block pubdata limit enforcement by charging the remaining scheme-independent per-block framing bytes: the serialized counters written into the pubdata stream.

It adds a regression test that exercises the block pubdata limit at the boundary with a single ERC20 transfer.

It also updates `deny.toml` to ignore `RUSTSEC-2026-0194` and `RUSTSEC-2026-0195`, which currently come transitively from `inferno` in tooling-only dependencies.

## Why ❔

`draft-0.4.0` already seeded the fixed block-level envelope bytes, but block-limit enforcement still missed the three serialized counters (12 bytes total). That allowed boundary cases to be admitted even though final pubdata serialization still needed additional bytes.

Charging those bytes upfront keeps admission aligned with the final serialized output and avoids late failure paths. This change is intentionally limited to the scheme-independent counters; it does not add DA-scheme-specific per-block overhead.

The cargo-deny ignore keeps CI green until the upstream tooling dependency updates `quick-xml`; it does not affect STF behavior.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


forward_system/src/run/mod.rs
```diff
@@ -170,6 +179,9 @@ pub fn run_block<
         block_metadata: block_context,
     };
     let chain_config_responder = ChainConfigResponder { chain_config };
+    let da_commitment_scheme_responder = DACommitmentSchemeResponder {
```
AntonD3: So you always use `BlobsAndPubdataKeccak256`(calldata) by default?

basic_bootloader/src/bootloader/block_flow/zk/pre_tx_loop.rs
```diff
@@ -3,8 +3,19 @@ use zk_ee::system::IOResultKeeper;
...
     block_flow::pre_tx_loop_op::PreTxLoopOp,
-    constants::{BLOCK_INTRINSIC_NATIVE, BLOCK_INTRINSIC_PUBDATA_BYTES},
+    constants::{
+        BLOBS_ZKSYNC_OS_LENGTH_PREFIX_PUBDATA_BYTES, BLOCK_INTRINSIC_NATIVE,
+        BLOCK_INTRINSIC_PUBDATA_BYTES,
+    },
 };
+use zk_ee::common_structs::DACommitmentScheme;
+
+fn block_da_commitment_intrinsic_pubdata(da_commitment_scheme: Option<DACommitmentScheme>) -> u64 {
```
AntonD3: Honestly not sure we should treat length encoding for blobs as pubdata. For me it feels more logical to just say that blobs capacity is 9 blobs without one word(reserved for length). Also there are some other issues:
- not sure it's convenient to pass commitment during sequencing on the server side, as I've seen you just pass `BlobsAndPubdataKeccak256` in the forward system
- and this overhead shouldn't be added for each block, it should be accounted only once per batch. Which actually is not trivial given that sequencer doesn't know about batches

AntonD3 (approved):

PR #719, opened 2026-07-15
feat(interop): carry the interop-root creation timestamp through import and rolling hash

StanislavBreadfulAI:
## What ❔

Adapts the atomic-interop OS branch to the timestamped interop roots of era-contracts `ad-atomic-interop` (matter-labs/era-contracts#2283):

- `InteropRoot` gains a `timestamp` field (`zk_ee` common struct + `api` sol helpers).
- The interop-root event hook parses the new `InteropRootAdded(uint256,uint256,uint256,bytes32[])` layout (timestamp in the data words) and the service-tx whitelist moves to the new `addInteropRootsInBatch((chainId,blockOrBatchNumber,timestamp,sides)[])` selector `0xc17a9fbd`.
- The bootloader's `interop_roots_rolling_hash` includes the timestamp, mirroring `ExecutorFacet`: `keccak(prev ‖ chainId ‖ blockOrBatchNumber ‖ timestamp ‖ sides)`.
- Interop test instance updated: timestamp slot assertions (value slot + 1) and the `L2InteropRootStorage` deployed bytecode re-vendored from the era-contracts build.

Companion PRs: zksync-os-server matter-labs/zksync-os-server#1459, era-contracts matter-labs/era-contracts#2290 (tooling fixes), plus the integration-tests branch update. Note: for the server pin (which must stay on nightly-2026-01-22 and pins zk_os by release tag), the same two commits are also published as fork tag `v0.3.2-interface-v0.1.3` = `v0.3.1-interface-v0.1.3` + this change.

**Follow-ups in this PR (same review round):**
- The legacy timestamp-less `addInteropRoot` entry point was removed from `L2InteropRootStorage` upstream (matter-labs/era-contracts#2291) — the vendored test bytecode is re-vendored accordingly.
- The bootloader now reads the interop commitment tree (IMT) root **directly from the engine's slots** instead of the removed slot-0 `_currentRoot` cache: `_height` from slot 0, then `_nodes[_height][0]` derived from the `_nodes` base slot 2 — the same scheme as `read_multichain_root` for the L2MessageRoot. Slot derivation pinned by unit tests, cross-checked against the Solidity layout lock test.

## Why ❔

The atomic-interop timeout protocol anchors timeout proofs on interop-root timestamps, so every imported root must carry (and commit to) the L1 timestamp it was created at; the executor double-checks the value via the dependency-roots rolling hash.

## Is this a breaking change?
- [x] Yes (interop-root import wire format + rolling-hash preimage change, in lockstep with the contracts branch)

## Checklist


AntonD3 (approved):

AntonD3:
https://github.com/matter-labs/zksync-os/pull/722

PR #720, opened 2026-07-15
perf(io): bound flat preimage-cache retention

antoniolocascio:
## Summary

- bound retained flat preimage-cache data to keep proving memory usage predictable
- apply a per-transaction retention budget and return out-of-native when it is exhausted
- reserve capacity for deferred account preimages before accepting a transaction
- preserve existing cache-hit and native-cost behavior

## Testing

- `cargo test -p basic_system preimage_cache`
- `cargo test -p basic_bootloader block_flow::zk`

### DIFF
--- a/basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
+++ b/basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
@@ -27,8 +27,8 @@ use zk_ee::system::errors::runtime::RuntimeError;
 use zk_ee::system::errors::subsystem::SubsystemError;
 use zk_ee::system::metadata::basic_metadata::{BasicMetadata, ZkSpecificMetadata};
 use zk_ee::system::metadata::zk_metadata::TxLevelMetadata;
-use zk_ee::system::tracer::Tracer;
-use zk_ee::system::validator::TxValidator;
+use zk_ee::system::tracer::{NopTracer, Tracer};
+use zk_ee::system::validator::{NopTxValidator, TxValidator};
 use zk_ee::system::Resource;
 use zk_ee::system::System;
 use zk_ee::system::{AccountDataRequest, CompletedExecution, Computational};
@@ -41,7 +41,7 @@ use zk_ee::{interface_error, internal_error, wrap_error};
 
 use system_hooks::addresses_constants::{
     BASE_TOKEN_HOLDER_ADDRESS, L2_ASSET_TRACKER_ADDRESS, L2_BASE_TOKEN_ADDRESS,
-    L2_CHAIN_ASSET_HANDLER_ADDRESS, L2_NATIVE_TOKEN_VAULT_ADDRESS, SYSTEM_CONTEXT_ADDRESS,
+    L2_CHAIN_ASSET_HANDLER_ADDRESS,
 };
 
 use super::validation_impl::compute_calldata_tokens;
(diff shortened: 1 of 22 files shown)


basic_system/src/system_implementation/flat_storage_model/preimage_cache.rs
```diff
@@ -1,11 +1,13 @@
 use alloc::{alloc::Global, collections::BTreeMap};
 use core::{alloc::Allocator, marker::PhantomData};
+use evm_interpreter::BitMapOwned;
```
AntonD3, line 3: nit: I know that we already had dependency on interpreter in basic system, but IMHO ideally we should try to minimize it

basic_system/src/system_implementation/flat_storage_model/preimage_cache.rs
```diff
@@ -22,6 +24,24 @@ use crate::cost_constants::blake2s_native_cost;
...
+/// use the same fixed domain; `begin_new_tx` rejects any attempt to exceed it.
+const MAX_PREIMAGE_CACHE_TX_IDS: usize = u16::MAX as usize + 1;
+/// The raw preimage cache may retain at most 256 MiB, including conservative
+/// per-entry map and allocator overhead.
+pub const MAX_PREIMAGE_CACHE_RETAINED_BYTES: usize = 256 * 1024 * 1024;
+/// A single transaction may add at most half of the block cache limit.
+pub const MAX_PREIMAGE_CACHE_BYTES_ADDED_PER_TX: usize = MAX_PREIMAGE_CACHE_RETAINED_BYTES / 2;
+
+const _: () =
+    assert!(MAX_PREIMAGE_CACHE_BYTES_ADDED_PER_TX * 2 == MAX_PREIMAGE_CACHE_RETAINED_BYTES);
```
AntonD3: nit: do we really need this check?

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
```diff
@@ -644,50 +782,48 @@ where
         .checked_sub(max_fee_commitment)
         .ok_or(internal_error!("td-mfc"))?;
 
+    let refund_recipient = u256_to_b160_checked(transaction.reserved[1].read());
+    prewarm_l1_postprocessing_balance_accounts(system, refund_recipient)?;
```
AntonD3, line 786: Why do you need to prewarm it here?

basic_system/src/system_implementation/flat_storage_model/preimage_cache.rs
```diff
@@ -30,24 +50,303 @@ pub struct PreimageRequest {
...
+    storage: BTreeMap<Bytes32, CachedPreimage<A>, A>,
     pub(crate) publication_storage: NewPreimagesPublicationStorage<A>,
     pub(crate) allocator: A,
+    accepted_transactions: BitMapOwned<A>,
+    /// ID of the candidate currently executing, or `None` outside candidate
+    /// execution and after the fixed `u16` ID space is exhausted.
+    current_tx_id: Option<u16>,
+    /// ID to assign to the next candidate. `None` means all `u16` IDs have
+    /// already been used.
+    next_tx_id: Option<u16>,
```
AntonD3, line 94: Do we really need it? Looks like it used only to increase current_tx_id, and in begin_new_tx we can just set current_tx_id to 0 if it's none

basic_system/src/system_implementation/flat_storage_model/preimage_cache.rs
```diff
@@ -30,24 +50,303 @@ pub struct PreimageRequest {
...
+    /// a preimage introduced only by the dropped candidate consumes the later
+    /// candidate's transaction-local budget, just as it does in proving.
+    pub(super) fn admit_cached_preimage_for_current_tx(
+        &mut self,
+        hash: &Bytes32,
+        expected_preimage_len: usize,
+    ) -> Result<(), SystemError> {
+        let admission = match self.storage.get(hash) {
+            Some(cached) if cached.as_slice().len() == expected_preimage_len => cached.admission,
+            Some(_) => return Err(internal_error!("Cached preimage length mismatch").into()),
```
AntonD3: Is it really reachable?

basic_system/src/system_implementation/flat_storage_model/preimage_cache.rs
AntonD3, line 634: any reason to have separate wrapper?

AntonD3: Oh, ok, I think it's because you don't want to expose apply_transaction_budget in this method, right?

basic_system/src/system_implementation/flat_storage_model/preimage_cache.rs
antoniolocascio: Fair

basic_system/src/system_implementation/flat_storage_model/preimage_cache.rs
antoniolocascio: Not really, let's get rid of it

basic_system/src/system_implementation/flat_storage_model/preimage_cache.rs
```diff
@@ -30,24 +50,303 @@ pub struct PreimageRequest {
...
+        }
+
+        Ok((next_tx_bytes, next_retained_bytes))
+    }
+
+    /// Returns the ID of the candidate that may own a `Pending` entry.
+    fn active_candidate_id(&self) -> Result<u16, SystemError> {
+        match self.current_tx_id {
+            Some(tx_id) => Ok(tx_id),
+            None if self.block_limit_hit_for_current_tx => Err(out_of_native_resources!().into()),
```
AntonD3, line 215: can it be called once we hit block limit?

basic_system/src/system_implementation/flat_storage_model/preimage_cache.rs
antoniolocascio: Yeah

basic_bootloader/src/bootloader/transaction_flow/zk/process_l1_transaction.rs
antoniolocascio: It has to be somewhere before execution, but can't be shared across the block like the asset tracker warming, as the refund recipient changes

basic_system/src/system_implementation/flat_storage_model/preimage_cache.rs
```diff
@@ -180,21 +503,104 @@ impl<R: Resources, A: Allocator + Clone> BytecodeAndAccountDataPreimagesStorage<
...
+            PreimageAdmission::Pending(self.active_candidate_id()?)
+        } else {
+            PreimageAdmission::Accepted
+        };
+        let (next_tx_bytes, next_retained_bytes) = if apply_transaction_budget {
+            self.next_estimated_byte_totals(estimated_entry_bytes, true)?
+        } else {
+            (
+                self.estimated_bytes_added_in_current_tx,
+                self.estimated_retained_bytes,
```
AntonD3, line 584: Shouldn't we increase `estimated_retained_bytes`?

basic_system/src/system_implementation/flat_storage_model/preimage_cache.rs
antoniolocascio: It's just used to prevent wrapping over u16. Don't think it's crucial, but I would keep it for now

basic_system/src/system_implementation/flat_storage_model/preimage_cache.rs
antoniolocascio: No, let's simplify it

basic_system/src/system_implementation/flat_storage_model/preimage_cache.rs
AntonD3: It's not cached in this case

basic_system/src/system_implementation/flat_storage_model/preimage_cache.rs
antoniolocascio: Yeah, same as the https://github.com/matter-labs/zksync-os/pull/720#discussion_r3814145828, it's just u16 wrapping protection. I don't think we need it, but honestly don't want to refactor it now

basic_system/src/system_implementation/flat_storage_model/preimage_cache.rs
antoniolocascio: No, that's done in `reserve_preimage_for_block_finalization` already

AntonD3 (approved):

PR #723, opened 2026-07-23
feat(da): always-public IMT leaves & L2→L1 logs; split DA into scheme (mechanism) + mode (scope)

AntonD3:
## What ❔

The STF half of the atomic-interop DA work (pairs with era-contracts `atomic-imt-interop-reduced-da`).

**1. IMT leaves are always committed to L1 DA**
- New **interop-commitment-leaf system hook** (`0x7004`, `system_hooks/src/call_hooks/interop_commitment_leaf.rs`) that the `L2InteropCommitmentTree` contract calls on every insert. It records the 32-byte leaf as an L2→L1 log via `IOSubsystem::emit_interop_commitment_leaf` (new `GenericLogContentData::InteropCommitmentLeaf` variant in `logs_storage.rs`), so the full set of inserted values — and thus the whole tree — is reconstructible from L1 DA regardless of the chain's state-diff DA choice. The hook charges only the log-record hashing native cost (the L1-message cost minus the message-payload hashing).

**2. Two orthogonal DA axes: scheme (mechanism) + mode (scope)**
- `DACommitmentScheme` is now **mechanism-only** (`None`, `EmptyNoDA`, `PubdataKeccak256`, `BlobsAndPubdataKeccak256`, `BlobsZKsyncOS`).
- New `DAMode { Rollup, Validium }` carries the committed **scope**, lives in `ChainConfig`, and is committed into the batch public input via `ChainConfig::hash` (appended as a 4th 32-byte word). It is read together with the rest of the chain config from the oracle.

**3. Per-mode pubdata layout — reported == committed** (`post_tx_op/mod.rs`)
- A version byte distinguishes layouts: **v2 (Rollup)** = full pubdata `[version, block_hash, timestamp, state diffs, logs, message payloads]` (unchanged from before); **v3 (Validium)** = `[version, logs_count, log records]` only (user-message logs, L1-tx result logs, IMT leaf logs). The **exact same bytes** are streamed to the DA commitment and to the sequencer/prover.

**4. Validium charges only committed pubdata**
- Pubdata cost (dynamic + intrinsic) counts only the committed log records in Validium; `System::net_pubdata_used` and the new `IOSubsystem::net_committed_log_records_pubdata_used` are mode-aware, with Validium intrinsic-pubdata constants and mode-threaded `gas_helpers`/`api` validation. The pubdata price is unchanged — only the byte counts differ.

**5. Golden vectors + docs**
- Pins `ChainConfig::hash`, `BatchPublicInput::hash`, and `BatchOutput::hash` golden vectors (shared with era-contracts `ZKsyncOSPublicInput.t.sol`), plus a `da_mode`-commits-to-PI test. Updated `docs/da_commitment_schemes.md` and `docs/system_hooks.md`.

## Why ❔

Atomic interop requires that IMT leaves and L2→L1 logs are **always** publicly reconstructible from L1 DA — even for a chain that does not publish full pubdata — so a Validium can still be a first-class interop participant. Separating the commitment **mechanism** (blobs vs calldata) from the committed **scope** (`Rollup` = full pubdata, `Validium` = mandatory logs only) lets a chain choose these independently, and committing `DAMode` into the public input lets the settlement layer enforce the chain's configured mode. Any chain operator can thus run Validium without giving up interop or log availability.

## Is this a breaking change?
- [x] Yes
- [ ] No

Requires coordinated deployment with the matching era-contracts changes: the batch public input's chain-config hash gains the `da_mode` word, and the pubdata stream now carries a version byte (with the new v3 Validium layout). The `Executor` chain-config hash and the `L2DAMode` diamond-storage field must land together.

## Checklist


basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/mod.rs
```diff
@@ -23,12 +24,32 @@ mod post_tx_op_proving_singleblock_batch;
...
+///
+/// The `Rollup` layout is unchanged from version 2 (full pubdata: block context,
+/// state diffs, logs and message payloads), so existing rollup DA consumers keep
+/// working.
+pub const ROLLUP_PUBDATA_ENCODING_VERSION: u8 = 2;
+
+/// Pubdata encoding version byte for `Validium` mode.
+/// Version 3: only the mandatory L2->L1 log section (`[version, logs_count, log
+/// records]`); state diffs and message payloads are not published.
+pub const VALIDIUM_PUBDATA_ENCODING_VERSION: u8 = 3;
```
antoniolocascio: Should we add another dimension instead? 1 byte for general version (we now bump from 2 to 3) and for version 3, we start with a byte for mode. I feel like this will make updating the version simpler

basic_bootloader/src/bootloader/constants.rs
```diff
@@ -298,6 +298,11 @@ pub const L1_TX_INTRINSIC_PUBDATA: u64 = 88
     + REFUND_RECIPIENT_BALANCE_INTRINSIC_PUBDATA
     + ASSET_TRACKER_INTRINSIC_PUBDATA;
 
+/// Validium (`DAMode::Validium`) L1 tx intrinsic pubdata: only the 88-byte L1->L2 tx log record is
+/// committed to DA. The coinbase/treasury/refund balance diffs and the asset-tracker diff folded into
+/// `L1_TX_INTRINSIC_PUBDATA` are state diffs that Validium does not commit (see `write_pubdata`).
+pub const VALIDIUM_L1_TX_INTRINSIC_PUBDATA: u64 = 88;
```
antoniolocascio: Nit: can we extract the 88 constant and share between this and L1_TX_INTRINSIC_PUBDATA?

tests/instances/eth_runner/src/single_run.rs
```diff
@@ -201,7 +201,6 @@ fn run_with_tracer<const RANDOMIZED: bool>(
                 DACommitmentScheme::BlobsAndPubdataKeccak256
             }
             "blobs_zksync_os" | "blobs" => DACommitmentScheme::BlobsZKsyncOS,
-            "empty_no_da" => DACommitmentScheme::EmptyNoDA,
```
antoniolocascio, line 204: Why?

AntonD3: Shouldn't be here, leftover from the previous design iteration

basic_bootloader/src/bootloader/block_flow/zk/post_tx_op/mod.rs
AntonD3: I thought about that - but why do you think updating will be simpler? To understand better what second byte specifically means: Imagine you need to update only one - validium or rollup, should we bump than general version? Or second byte for 1 of them?

antoniolocascio: I think they are different things. One is the mode: "rollup | prividium". The other is the version. Yes, maybe for some versions only one mode will change, but having to bump two whenever there's a common change seems messy.

antoniolocascio: In general: I think we should have a single version per protocol version. Conflating the mode here seems conceptually off.

antoniolocascio (approved):

PR #748, opened 2026-09-06
fix(divergence-validator): make EIP-7702, deep call stacks and block overrides testable

Valisthea:
## Summary

`tests/evm_divergence_validator` is offered in its README as a bug bounty triage tool and as an automated feedback loop. Four defects currently make parts of the EVM surface impossible to compare through it, and a fifth makes any divergence it does report hard to act on. Each is fixed here, with an example scenario that fails without the fix.

Nothing in the engine is touched. All changes are under `tests/`.

## What is currently unreachable

**Any scenario declaring a `block:` section.** `make_block_context` left `blob_fee` at zero, while the bootloader requires exactly 1 when EIP-4844 is disabled, which is the production configuration. The documented block override therefore aborted with an internal error before execution started. `tests/block_reexecutor` already builds its context with `U256::ONE`.

**Any account holding an EIP-7702 delegation.** Three separate layers:

1. `revm_state_provider::basic_ref` validated the still padded preimage through `code_by_hash_ref`. Preimages are stored padded with the true length in `unpadded_code_len`, so a well formed 23 byte designator was rejected as malformed and surfaced as a divergence.
2. Even once trimmed, the result was wrapped in `Bytecode::new_legacy`, so REVM would never see a designator as a delegation.
3. `storage_diff_comp::build_revm_accounts` returned an error for `BytecodeKind::Eip7702`, so the consistency checker could not compare such an account at all.

EIP-7702 is enabled by the `production` feature set, so these accounts are part of the surface this tool is meant to cover.

**Any scenario nesting more than a few hundred call frames.** Both engines recurse natively per EVM call frame. On the main thread's default stack the process aborted before producing a report, which put the 1024 call depth limit out of reach.

**Diagnosing a reported divergence.** The mismatch error carried only the number of differing storage slots and accounts, not which ones.

## Verification

Each example fails without its fix and passes with it:

| Fix reverted | Failing test |
| --- | --- |
| `runner.rs` blob base fee | `block_context` |
| `revm_state_provider.rs` preimage trimming | `eip7702_delegation` |
| `storage_diff_comp.rs` delegated account hashing | `eip7702_delegation` |
| `main.rs` worker stack | `deep_call_stack` |

With all four applied:

```
cd tests/evm_divergence_validator
cargo test --release
```

```
running 4 tests
test block_context ... ok
test predeployed_bytecode ... ok
test eip7702_delegation ... ok
test deep_call_stack ... ok
```

The `deep_call_stack` example is also useful on its own: raising the bound compiled into its bytecode past 1024 shows the depth limit binding rather than gas, with both engines agreeing on where it binds.

## Notes

`tests/evm_divergence_validator` is excluded from the workspace, so the new integration suite does not run in CI. Happy to wire it up in a follow up if that is wanted; it would need the crate's `Cargo.lock` committed as well, since a fresh resolve currently picks an `alloy` that `revm-inspectors 0.39` cannot build against.

That same resolution drift already affects the workspace itself, independently of this change. On an unmodified `origin/main`:

```
cargo clippy --all -- -D warnings

error[E0063]: missing fields `state_gas_cost` and `state_gas_reservoir` in initializer of `StructLog`
error[E0063]: missing fields `execution_gas_used`, `gas_refund` and `state_gas_used` in initializer of `DefaultFrame`
error: could not compile `revm-inspectors` (lib) due to 3 previous errors
```

This is why the jobs that build `revm-inspectors` are red while the others are green. It looked out of scope for this PR, and the `aba-add-cargo-lock` branch already addresses it, so nothing here touches dependency resolution.

### DIFF
--- a/tests/evm_divergence_validator/examples/block_context.yaml
+++ b/tests/evm_divergence_validator/examples/block_context.yaml
@@ -1,0 +1,26 @@
+# Example declaring an explicit `block:` section.
+#
+# The bootloader requires the blob base fee to be exactly 1 when EIP-4844 is
+# disabled, which is the production configuration. Any scenario that reaches
+# `make_block_context` therefore depends on that field being set correctly, so
+# this example exercises the block override path end to end.
+contracts:
+  Store:
+    bytecode: "0x60003560005500"
+    address: "0x0000000000000000000000000000000000000101"
+
+accounts:
+  alice:
+    balance: "1000000000000000000"
+
+block:
+  basefee: 1000
+  gas_limit: 30000000
+  timestamp: 1700000000
+
+steps:
+  - type: send_raw
+    to: Store
+    from: alice
+    data: "0x000000000000000000000000000000000000000000000000000000000000002a"
+    gas: 1000000
--- a/tests/evm_divergence_validator/examples/deep_call_stack.yaml
+++ b/tests/evm_divergence_validator/examples/deep_call_stack.yaml
@@ -1,0 +1,34 @@
+# Example driving the EVM call stack to its 1024 frame limit.
+#
+# The contract calls itself, carrying the current depth in 32 bytes of calldata
+# and stopping once that depth reaches the bound compiled into the bytecode. The
+# deepest frame that succeeds writes last, so slot 0 ends up holding the maximum
+# depth reached. A failing nested call does not revert its parent, so the write
+# survives and the depth limit itself becomes observable.
+#
+# Bytecode, with the bound 0x000400 = 1024 pushed at offset 6:
+#   PUSH0 CALLDATALOAD            d
+#   PUSH0 SSTORE                  slot 0 <- d
+#   PUSH0 CALLDATALOAD            d
+#   PUSH3 0x000400 GT             bound > d ?
+#   PUSH1 0x0f JUMPI              continue if so
+#   STOP
+#   JUMPDEST
+#   PUSH0 CALLDATALOAD PUSH1 1 ADD PUSH0 MSTORE
+#   CALL(gas=GAS, self, value=0, args=0..32, no return)
+#   POP STOP
+contracts:
+  Recur:
+    bytecode: "0x5f355f555f356200040011600f57005b5f356001015f525f5f60205f5f305af15000"
+    address: "0x0000000000000000000000000000000000000101"
+
+accounts:
+  alice:
+    balance: "1000000000000000000000"
+
+steps:
+  - type: send_raw
+    to: Recur
+    from: alice
+    data: "0x0000000000000000000000000000000000000000000000000000000000000000"
+    gas: 100000000000000
--- a/tests/evm_divergence_validator/examples/eip7702_delegation.yaml
+++ b/tests/evm_divergence_validator/examples/eip7702_delegation.yaml
@@ -1,0 +1,35 @@
+# Example covering an account that carries an EIP-7702 delegation designator.
+#
+# EIP-7702 is enabled by the production feature set, so a delegated account is
+# part of the surface this tool is meant to compare. Such an account holds the
+# 23 byte designator 0xef0100 || address as its bytecode.
+#
+# Delegate bytecode: 602a5f5500
+#   PUSH1 0x2a, PUSH0, SSTORE, STOP   (stores 42 in the caller's slot 0)
+#
+# Probe bytecode: 5f5f5f5f5f610202620186a0f15f556102023b6001556102023f60025500
+#   CALL(gas=100000, 0x0202, value=0, no args, no return)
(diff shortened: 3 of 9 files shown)


antoniolocascio:
Hey, thanks for the PR! A couple of notes:

* The stack issue should be fixed by running on release mode, no need to reserve so much memory.
* The 7702 issue is already fixed in our next version branch (draft-0.5.0). This will become main very soon, so we can just keep it like that.
* The blob base fee issue is real, will open a PR against the dev branch to fix it.

Thanks again for the contribution!

Valisthea:
Thanks for the quick look!

**On the stack:** that measurement was already release mode. The binary reserves 1 MB (`SizeOfStackReserve = 0x100000` in the PE header), which is the Windows default, against 8 MB on Linux. So the scenario that overflows around 350 frames here has roughly 8x more headroom on your CI, which is why you do not see it. It is a platform difference rather than an optimization one, and it only affects contributors on Windows. Happy to drop that commit, or shrink the reservation to something like 64 MB if you would rather keep deep-stack scenarios runnable there.

**On EIP-7702:** I may well be looking in the wrong place, but `draft-0.5.0` at `20fdb610` still seems to carry all three of the paths this PR touches:

```
git show 20fdb610:tests/revm_runner/src/revm_state_provider.rs | sed -n '178,185p'
git show 20fdb610:tests/revm_runner/src/storage_diff_comp.rs   | sed -n '278,287p'
```

The first still validates the still-padded preimage through `code_by_hash_ref` and then wraps the result in `Bytecode::new_legacy`, so a 23 byte designator is rejected as malformed and, even when it is not, never reaches REVM as a delegation. The second still returns `EIP-7702 bytecode is not supported on Consistency Checker`. Both are byte identical to `main`; the only change to `revm_state_provider.rs` on that branch is a `BlockHashes` import move.

The two EIP-7702 commits on `draft-0.5.0` (#709 and #710) both land in `tests/instances/eth_runner`, which looks like a different harness from `revm_runner` / `evm_divergence_validator`. If `draft-0.5.0` is the better target I am happy to rebase onto it.

**Blob base fee:** sounds good. Feel free to cherry-pick `49a06ac4` if that saves you time.

antoniolocascio:
@Valisthea must admit, I didn't check 0.5.0 and I had a false memory that this was fixed. You're right, we need it. So, sure, let's include it. Would you rather rebase it yourself (either dropping the stack commit or changing it to 64MB) on draft-0.5.0 or me to do it? Either is fine with me

Valisthea:
No problem at all, easy thing to misremember.

I will take the rebase. Moving it onto `draft-0.5.0` and keeping the stack commit, changed to 64 MB rather than dropped: it is reserved address space rather than committed memory, so it costs nothing on your runners, and it keeps the deep-stack scenarios runnable for contributors on Windows. If you would still rather see it gone, say the word and I will drop it.

Will push shortly.

Valisthea:
Done: rebased onto `draft-0.5.0`, stack commit changed to 64 MB, and I retargeted the PR base. Six commits, 9 files, +207/-13.

Red/green on that branch, using `examples/eip7702_delegation.yaml` from this PR.

Without the two `revm_runner` commits:

```
{
  "status": "divergence",
  "steps": [ { "description": "send_raw to Probe", "success": true, "gas_used": 114835 } ],
  "error": "Database error: MalformedCodePreimage { code_hash: 0x14b105ebf507bcb6cf2645d592bfd43d6564ca1b2de7aceb23c84938aa936a36, reason: \"Eip7702 is not 23 bytes long\" }"
}
```

With them, `cargo test --release` in `tests/evm_divergence_validator`:

```
test block_context ... ok
test deep_call_stack ... ok
test eip7702_delegation ... ok
test predeployed_bytecode ... ok
test result: ok. 4 passed; 0 failed
```

One practical note if you reproduce this: cargo did not reliably rebuild `zksync_os_revm_runner` for me after reverting those two files, and I got a false `"status": "match"` twice from a stale binary before noticing. Running `cargo clean -p zksync_os_revm_runner -p evm_divergence_validator` between arms made it deterministic.

Unrelated to this PR, but worth flagging while I was on that branch: building the validator on `draft-0.5.0` now requires the CUDA Toolkit. `zksync_os_runner` declares `airbender-host = { workspace = true, features = ["gpu-prover"] }`, which pulls `era_cudart_sys`, whose build script panics with `Failed to determine the CUDA Toolkit version` on a machine without it. The chain is `evm_divergence_validator -> rig -> zksync_os_runner -> airbender-host -> gpu_prover -> era_cudart`. Happy to open a separate issue if that is useful.

antoniolocascio (approved):

antoniolocascio:
LGTM, thanks!

PR #749, opened 2026-09-07
fix(evm_interpreter): decide memory windows before charging heap growth

antoniolocascio:


shamatar:
4Gb limit is unreachable by TX gas limits already, so can you just go usize -> u32 instead in grow and offset calls? At least in `cast_offset_and_len` call

